PR #2, opened 2017-10-11
Fix declarations and definitions

kdheepak:


kdheepak:
There were some declarations and definitions that did not match, which this PR addresses. Let me know once you take a look, and I can merge.

PR #14, opened 2017-10-31
Mac Travis CI config

nightlark:
Mac Travis CI config that builds on OSX 10.10 and 10.11

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -74,7 +74,7 @@ before_install:
   - eval "${MATRIX_EVAL}"
 
 install:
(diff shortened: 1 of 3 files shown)


kdheepak:
It looks like this is working. I'm not sure why the newer Mac build fail? But this is good for now. Thanks for taking a shot at this @nightlark! @phlptp this is ready to merge.

kdheepak:
This commit a3680c9aa3d7c24eb60b0bb609b636e0981dd36e is the only one that is a change to HELICS src. All other changes are to the build script / .travis.yml.

PR #16, opened 2017-11-01
Udp core

phlptp:
Merge in UDP core code and other improvements and bug fixes resulting from working through that type of core.  
All tests are passing on Windows/Linux/ and Mac as far as I can tell


nightlark:
How are dropped packets detected/handled?

phlptp:
At present they are not,  that gets into a bigger question of how comm failure in HELICS in general is handled which I don't think we are in a position to address for the moment.  My purpose in the Actual UDP comms was to further test the communication mechanics for the other core types, and to get some experience working with boost asio, as well as networking with the something a little lower level than ZMQ.  To prepare for a  TCP core which will likely be more reliable in longer range networking conditions.  
The other thing is on a small network the probability of dropped packets for our current usage is pretty close to zero, so I didn't want to spend a lot of time working on something that isn't important right now.  
At some point we need to start testing HELICS in failure conditions but I don't think that time is now.  If you have ideas on reliability I am very interested.

nightlark:
With UDP sockets, usually it seems like another layer is used to add reliable and/or in-order delivery. I came across a project to implement UDT for reliability that uses boost asio UDP sockets, and there are a few other libraries that add a layer on top of UDP (most without boost).

For larger simulations between multiple computers, we will need reliable delivery so that a single dropped packet doesn't stop the entire simulation (and so that value updates and messages aren't missed). For repeatable results, we might need in-order delivery for some messages types -- order of receiving a message packet being used as a tie-breaker for queue position would be a big one. Some libraries have options to send each packet with different combinations of reliable and in-order delivery depending on what that packet type needs.

Everything compiled and the tests passed on LC. I'll try the examples and see if I can reproduce the hanging issue.

nightlark:
The pi_sender and pi_receiver examples are hanging -- the broker exits fine, but I needed to kill the two federate processes.

kdheepak:
Adding to @nightlark's comment, which I think @phptp is already aware of, this is not just an issue with this branch but even with develop.

phlptp:
the hanging could be some issue with the C-shared library, which the pi_sender and reciever use but not the broker.  I can poke at things but we really need the C-library test cases operational before I have much confidence in that library.

phlptp:
Regarding the UDP Core itself I would argue all the other modifications and extensions will involve extra overhead and I would consider them a separate communication system.  It may be a simple extension but for current purposes the UDP comms is not intended to guarantee reliability it was to get some more options and different stresses on the system code.  We can harden it later if it becomes more used.

nightlark:
I'll run the tests for this branch on LC.

phlptp:
Do we want to merge this into Develop?

nightlark:
Yes, I'll merge it now.

PR #17, opened 2017-11-06
C library halting

phlptp:
Can you check if this branch fixes the halting problem.  I had to add a closeLibrary function to the applicationAPI to shut down everything properly.  It lets the test cases run to completion on Windows and Linux.  I still had a few issues on mingw that I haven't figured out how to track yet.


kdheepak:
I tested this on my Mac using the Python examples and it worked. I updated the Python examples in this branch.

nightlark:
The pi receivers exited immediately after disconnecting from the broker; the pi senders had a pause (~5s) after disconnecting and then were forced to close -- much better than hanging forever.

phlptp:
My understanding of the shutdown process with shared libraries is still evolving particularly when it comes to threading,  I suspect there is also some compiler differences as well.  Windows seems to destroy the threads before destructing the static variables.  I think Linux might handle the ordering a bit differently.  and Different compilers on Linux might do things a bit differently as well.  Hopefully we will get the shutdown process figured out eventually.  Anyway I put a timeout in most of the shutdown loops now so it will just give up trying to shut things down properly if it has to wait too long.

nightlark:
As part of the timeout during shutdown, is freeing the zmq socket one of the things that doesn't get done properly?

phlptp:
I think it was more a problem with the zmq_context not shutting down rather than an improper shutdown of the cores.  So in case with the shared library it would shutdown the threads before the disconnect signal was processed.  This lead to the broker not receiving the shutdown signal.  That is the primary reason for the need for a close library.  so all the signals in all the threads have a chance to be processed.    Are you still getting cases where it times out?

phlptp:
Is that with both pi_sender and pi_sender2?

phlptp:
Any way you can halt a debug at the timeout message and print the stack trace for the all the process threads?

nightlark:
It is both pi_sender and pi_sender2.

Odd how I can't find a way to save or copy paste stack traces in totalview, and gdb doesn't want to run it because of issues with ld library (mismatched versions?).

Here are images of the stack traces for pi_sender threads. Two of them don't have symbol info other than epoll_wait_nocancel.

pi_sender_thread1
pi_sender_thread6
pi_sender_thread7
pi_sender_thread2
pi_sender_thread3

phlptp:
the stop should be on the std::cerr line where it prints out the message.  it could go through that loop normally a few times.  the epoll threads are part of the zmq stuff and the logger isn't really relevant.

nightlark:
The stack traces for the other threads are in the same places when it stops on the std::cerr line.

phlptp:
Ok I will try a few more things tonight and we can try some more tomorrow.

phlptp:
Lets try this again,  I made a few more updates to the destruction routine.  @nightlark  see if it works now?

nightlark:
Stack trace has the same locations for the halting problem with both pi_senders, and the nonlings_fed1.

phlptp:
can you turn on all the logging and see what the output is for all the federates.

nightlark:
Ah, it's user error for the most recent hanging. I'd been starting an extra helics_broker since I originally tried starting a pi_receiver first and thought it was expecting one when it failed to connect.

It looks like the group of commits 60915ad - 57b172c "More attempts at clearing up some tsan race conditions" fixed the halting problem. The commit prior to that group of commits (f275b5e) has issues with destroying objects.

All of the examples with the most recent commit (a9bf19b) seem to be shutting down without any issues.

phlptp:
Well that is good to hear.  I was really running out of things to try especially since I couldn't replicate the issue on any of my systems.    If you think it is ready go ahead and merge the pull requests, and we can start reducing the number of branches again.

PR #20, opened 2017-11-20
Api updates

phlptp:
This is an update to the iteration API along with several other bug fixes and internal improvements.  I believe this adds in the quick tests configuration so we should be able to run that on the CI systems in one or more of the builds.


nightlark:
What's the logic for the setMinimumChange methods in publications/subscriptions? It looks like negative changes in value have some special meaning that enables/disables change detection (based on both the previous delta value and the newly set value) separate from the method for enabling change detection.

```
void` setMinimumChange (double deltaV)
     {
         if (delta < 0.0)
         {
             changeDetectionEnabled = true;
         }
         delta = deltaV;
         if (delta < 0.0)
         {
             changeDetectionEnabled = false;
         }
     }
```

phlptp:
The intention with that function was that if the delta was negative before, just reset the enabled function, otherwise keep it as it was.  then check again with the new value.  Since change is always going to result in a positive number, if the minimum change <0 there really isn't any point in checking so the check is disabled.  However there is also a manual switch to control it.  so if there was a normal delta specified, we don't want to change the enabled setting unless it was turned off by the check.  

the expected use case is to set it at some level, then turn it on an off with the switch,  the check in that function is prevent the execution of what could potentially be an expensive operation if it is not needed.  The default delta is <0 meaning the check if off

phlptp:
I suppose I should add a bit of documentation in the code explaining that

nightlark:
Okay, that makes sense.

nightlark:
I think this is ready to merge, unless you want to add a comment to the setMinimumChange function first.

phlptp:
I will add the comment in the next chunk of code updates that will come in the next couple days.  Go ahead and merge it.

PR #24, opened 2017-11-21
Update .travis.yml with Python extension

kdheepak:


phlptp:
looks like there is no swig on Travis

kdheepak:
I submitted this PR to trigger travis build, but I didn't realize we were building all the branches. I'll update this branch and figure it out.

phlptp:
Why did we have to update cmake?  At least some of the test builds  should be at the minimum CMAKE version we specified which is 3.4.

kdheepak:
SWIG has different CMake commands for different versions of CMake that aren't compatible. I can add a new target with a separate version of CMake that builds the Python extension. Or we'd have to update the CMakeLists.txt file to include the appropriate CMake SWIG command depending on what version was being used.

phlptp:
If possible I would prefer the latter since I don't want to change the requirements on CMAKE until sometime after version 1.0.    Is it an option to have different CMAKE versions on different builds?  @nightlark ?  It might be good to have the gcc 6.0 test also run with a newer cmake?  and one of the MAC builds as well.    I am not entirely clear what is possible/easy with the Travis Stuff yet.

phlptp:
But at least some of the build tests should run with CMAKE 3.4.X  to make sure we stay compatible with the stated minimum version.

kdheepak:
This is ready to merge.

phlptp:
can you remove the message calls in CMAKE, probably best we don't clutter up the cmake output more than we have to.

kdheepak:
Good point. I've removed the message calls.

nightlark:
@phlptp An extra condition could be added to one of the build matrix entries that tells the script to fetch a different version of CMake when setting up the dependencies.

phlptp:
If that is easy to do it might be worthwhile to change the gcc 6 build to use cmake 3.9,  then we would have an extra check to make sure we are running with older and newer cmake versions in case of any check differences.

PR #26, opened 2017-11-28
Filter new api

phlptp:
This is the new Filter API and some other changes to CMAKE build scripts and other API changes.  Several redundant tests have been removed so it looks like fewer tests are active.


phlptp:
@nightlark any ideas why the pr branch is failing on gcc-6  seems to say it can't find the compiler,  but I don't know why it is working on the branch but not the pull request?

nightlark:
Latest change to develop should fix it. A newline got deleted in develop for the packages section -- I need to double check the preview when using the web editor, the cursor has a weird tendency to jump to a line that wasn't selected (usually top of the file).

PR #36, opened 2017-12-07
HELICS MATLAB extension

kdheepak:


kdheepak:
I've added a CMakeLists.txt in the swig/matlab folder that builds the MATLAB extension using CMake
I added custom commands, so this is actually SWIG independent.
If the shared library api changes, I'll have to regenerate the SWIG bindings using my custom version of SWIG.
The custom commands include the headers and link to the shared library that is created in the build folder.
See [line 30](https://github.com/GMLC-TDC/HELICS-src/blob/a14bb7c/swig/matlab/CMakeLists.txt#L30).

Issue #54, opened 2017-12-19
SWIG command produce warning for cmake 3.10


phlptp:
Looks like the version check in the python CMake files doesn't work for cmake 3.10, it produces a bunch of warnings on deprecated SWIG commands.

phlptp:
I need to verify this on on a system with cmake 3.10 and swig yet but I think that should fix it, once I verify I will   close the issue.

kdheepak:
I'm able to confirm that it works now with 3.10.

PR #58, opened 2018-01-05
Fix version compare string

kdheepak:


kdheepak:
This should address #54.

Issue #57, opened 2018-01-02
Unable to install HELICS using Boost 1.66 on OSX


kdheepak:
```
$ cmake --version
cmake version 3.10.1

$ gcc --version
Configured with: --prefix=/Library/Developer/CommandLineTools/usr --with-gxx-include-dir=/usr/include/c++/4.2.1
Apple LLVM version 9.0.0 (clang-900.0.39.2)
Target: x86_64-apple-darwin16.7.0
Thread model: posix
InstalledDir: /Library/Developer/CommandLineTools/usr/bin

$ # boost version
$ ls /usr/local/Cellar/boost/1.66.0/
INSTALL_RECEIPT.json include              lib

make
[ 14%] Built target application_api
[ 23%] Built target helics_common
[ 24%] Building CXX object src/helics/core/CMakeFiles/helics_core.dir/udp/UdpComms.cpp.o
In file included from /Users/$USER/GitRepos/HELICS-src/src/helics/core/udp/UdpComms.cpp:12:
In file included from /Users/$USER/GitRepos/HELICS-src/src/helics/core/udp/../../common/AsioServiceManager.h:30:
/usr/local/include/boost/asio/io_service.hpp:27:20: error: typedef redefinition with different types ('boost::asio::io_context' vs
      'boost::asio::io_service')
typedef io_context io_service;
                   ^
/Users/$USER/GitRepos/HELICS-src/src/helics/core/udp/UdpComms.h:23:15: note: previous definition is here
        class io_service;
              ^
In file included from /Users/$USER/GitRepos/HELICS-src/src/helics/core/udp/UdpComms.cpp:12:
/Users/$USER/GitRepos/HELICS-src/src/helics/core/udp/../../common/AsioServiceManager.h:39:34: error: incomplete type 'boost::asio::io_service' named in
      nested name specifier
    std::unique_ptr<boost::asio::io_service::work> nullwork; //!< pointer to an object used to keep a service running
                    ~~~~~~~~~~~~~^~~~~~~~~~~~
/Users/$USER/GitRepos/HELICS-src/src/helics/core/udp/UdpComms.h:23:15: note: forward declaration of 'boost::asio::io_service'
        class io_service;
              ^
/Users/$USER/GitRepos/HELICS-src/src/helics/core/udp/UdpComms.cpp:141:17: error: no matching constructor for initialization of 'udp::socket' (aka
      'basic_datagram_socket<boost::asio::ip::udp>')
[...]

PR #60, opened 2018-01-05
fix issue with boost 1.66 #57

phlptp:
this will hopefully fix the errors associated with boost 1.66


kdheepak:
Builds on my Mac with boost 1.66! Thanks!

PR #69, opened 2018-01-09
C++98 interface

nightlark:


phlptp:
@nightlark do you have any tests that use the interface?  If not we probably should have at least a couple that get run regularly in the c-interface tests

phlptp:
do you know of a way in Cmake to force the examples you built to compile with C++98?

nightlark:
There are only the ones in the examples/CppInterface folder, actual tests will look more like the c-interface tests than the application api tests.

The target_compile_options command with "-std=c++98 -pedantic" in the CMakeLists.txt for the examples is the closest to forcing compilation with C++98.

nightlark:
Strange that the tests are timing out with gcc-6 but none of the other compilers.

phlptp:
I have noticed that on occasion.  I think it is actually because gcc-6 is a little better at threading so there is more potential for timing glitches to appear.  I have been working on that test in other branches.

PR #73, opened 2018-01-11
Player recorder update

phlptp:
@nightlark and anyone else
This pull request is for a substantial update to the the players and recorders.  
In the process of updating those some work was done on the filters and finish an application API for filters.  It also fixes a few bugs and issues discovered in the timing systems.  

This will need to be merged and then synced with the C_UNIT_tests branch which will be the other major branch to finish the filter interface.  I would like this merged before we start the naming convention update.


src/helics/application_api/Federate.cpp
```diff
@@ -377,22 +376,22 @@ void Federate::setTimeDelta (Time tdelta)
...
-    coreObject->setLookAhead (fedID, lookAhead);
+    coreObject->setOutputDelay (fedID, outputDelay);
 }
 
-void Federate::setImpactWindow (Time window)
+void Federate::setInputDelay (Time window)
 {
     if (window < timeZero)
     {
         throw (InvalidParameterValue ("Impact Window must be >=0"));
```
nightlark: Include "Input Delay" in the error message and/or change argument name from window to impactWindow (or to inputDelay to match the output delay naming?).

Issue #104, opened 2018-01-20
Add C++98 API to a project so it shows up in an IDE folder and can be interacted with


phlptp:
The C++98 API is in the code base but it doesn't show in Visual studio or any other project based IDE.  
I think we just need to add the hpp files to a project of some kind so the code links, probably doesn't even need to compile as it is header only.

PR #109, opened 2018-01-22
C++98 API library and IDE targets

nightlark:
Adds a target to make the C++98 API header files show up in IDEs (Visual Studio, issue #104). Creates a helicsCpp98 interface library that can be used for target_link_libraries command.


phlptp:
I modified it so that the *_ide project actually built something using empty.cpp which makes a very tiny useless library.  The reason for this is that when I loaded it into visual studio a whole bunch of errors showed up on the static analysis due to missing header information which I couldn't find a way to add to the custom project.    This way I could include the proper header locations for the shared library and no errors show up.

PR #110, opened 2018-01-23
Set up Travis config for daily builds

nightlark:
Added a set of build configurations for a daily build that are separate from the configurations used on each commit. Will eventually be used to run code coverage and valgrind on the tests. Also changed os x builds to only take place for the master and develop branches.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,29 @@ cache:
     - cmake-3.4.3-Darwin-x86_64/CMake.app/Contents/share
     - dependencies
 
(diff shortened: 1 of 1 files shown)


kdheepak:
Nice, I didn't know you could do this. I'm still not clear on how this makes it a daily build though. @nightlark where is the cron schedule specified?

kdheepak:
I see that it is on the travis settings itself. This looks good to me.

nightlark:
@kdheepak some features they added pretty recently, back in September/October

PR #112, opened 2018-01-23
Add helics.py and _helics python library to install directory

kdheepak:


kdheepak:
I've updated the README.md and tested this locally. If this looks good I can merge it to `develop`.

kdheepak:
@nightlark That's a good idea. I'll update it the instructions.

phlptp:
@kdheepak  check that the install of the files works correctly.  I don't have access to my system where I built the python extension at the moment so I was just copying from another location.  And if it works go ahead and merge the commit and close the issue.

Issue #72, opened 2018-01-11
add Json configuration for filters


phlptp:
the json configuration for filters is not well established or tested, and a few mechanics of filters need to be worked out.

PR #118, opened 2018-01-26
Json configuration update

### LINKED ISSUE #70: Add ability to handle comments in the json configuration files for recorder and player
Right there are two different JSON readers in HELICS.  One can handle comments which is what is used for the federate configuration,  but the one used in the players cannot handle comments.  We should add the ability to strip off any comments in the player and recorder configuration file or switch JSON readers to one that handles them.

phlptp:
This pull request adds in json configurations for filters and some example files
It also changes the json reader for the players and recorders to support comments.  There are a couple more features to add yet but they are currently in development in branch thread_safety_improvements. 
This pull request will close #70 and #72


nightlark:
The changes look okay for merging. I noticed that appveyor is running a different set of tests than travis -- is that something that never got updated, or a choice made due to the failures (timeouts) on travis while running the tests?

phlptp:
It is purposeful.  Originating in the timeout problems with Travis.  I suppose eventually we will want to run some additional tests on travis,  I have been tinkering with the tests running on Appveyor somewhat, to see how they do,  At some point we can evaluate which tests to run on Travis vs Appveyor vs nightly

phlptp:
@kdheepak @nightlark  any issues with me merging this to develop?

Issue #128, opened 2018-01-30
Upgrade the boost version used in travis


phlptp:
The boost version used in Travis should be upgraded to either 1.64 or 1.65,1 for the clang and gcc 6 builds.  This will eliminate the deprecated pragma warnings in those builds and test out the updated software.  

the gcc 4.9 build should stick with 1.61 for the time being to make sure we have a test on the oldest supported versions of compiler and boost.

Issue #108, opened 2018-01-21
#pragma deprecated warnings from boost


phlptp:
```
In file included from /home/travis/build/GMLC-TDC/HELICS-src/dependencies/boost/include/boost/iostreams/detail/is_dereferenceable.hpp:12:0,
                 from /home/travis/build/GMLC-TDC/HELICS-src/dependencies/boost/include/boost/iostreams/detail/resolve.hpp:26,
                 from /home/travis/build/GMLC-TDC/HELICS-src/dependencies/boost/include/boost/iostreams/detail/push.hpp:24,
                 from /home/travis/build/GMLC-TDC/HELICS-src/dependencies/boost/include/boost/iostreams/detail/streambuf/indirect_streambuf.hpp:31,
                 from /home/travis/build/GMLC-TDC/HELICS-src/dependencies/boost/include/boost/iostreams/stream_buffer.hpp:22,
                 from /home/travis/build/GMLC-TDC/HELICS-src/dependencies/boost/include/boost/iostreams/stream.hpp:21,
                 from /home/travis/build/GMLC-TDC/HELICS-src/src/helics/application_api/ValueConverter_impl.hpp:37,
                 from /home/travis/build/GMLC-TDC/HELICS-src/tests/helics/application_api/value-converter-tests.cpp:22:
/home/travis/build/GMLC-TDC/HELICS-src/dependencies/boost/include/boost/type_traits/detail/bool_trait_def.hpp:18:78: note: #pragma message: NOTE: Use of this header (bool_trait_def.hpp) is deprecated
 # pragma message("NOTE: Use of this header (bool_trait_def.hpp) is deprecated")
```

this warning is showing up in the build logs due to the use of boost 1.61 in travis and the use of boost::iostreams.  

This is a known issue that was fixed in boost 1.62.    As far as I know the only known fix is to upgrade to boost 1.62 or higher.  The code all works fine.  

The question is whether it is worth upgrading travis to 1.62 or higher to get rid of the warnings.  Costs are that 1.61 is the lowest boost version we support that compiles everything.  so we would loose the earliest supported platform if we upgrade.

kdheepak:
We can add a new target in Travis to compile for Boost 1.61, but use the latest boost in a separate target so that we can look at the logs there. @nightlark thoughts?

nightlark:
[...]

PR #133, opened 2018-02-01
Update version of Boost used on Travis, commit keyword to update cached dependencies

nightlark:
* Uses Boost 1.65 for all builds except with gcc-4.9; configurations can be built using Boost 1.61 by adding a `MINIMUM_DEPENDENCIES=true` environment variable. Boost installation commands have been put in their own function.
* Recognizes `[update_cache]` in commit messages to purge the contents of the dependencies directory, or if the commit message also contains `boost` or `zmq`, it will only remove the dependencies mentioned in the commit message.

Addresses issues #128 and #108

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -78,7 +78,9 @@ jobs:
             - ubuntu-toolchain-r-test
           packages:
             - g++-4.9
-      env: MATRIX_EVAL="COMPILER=g++-4.9 && CC=gcc-4.9 && CXX=g++-4.9"
+      env:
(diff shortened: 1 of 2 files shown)


phlptp:
just to be clear the cache for boost and zmq will only be updated if the commit contains [update_cache] and the "boost" note will be ignored if there is no [update_cache] flag.

We will have to be careful on squash commit to remove that unless desired

nightlark:
Yea, if the `update_cache` flag isn't there, no updates to the cache will happen (unless it doesn't exist for whatever reason). Worst case, the build takes an extra 5 minutes.

Issue #135, opened 2018-02-02
Python federate ends with uncaught exception


bpalmintier:
Following the python test steps from `swig/python/README.md` the SENDER federate (only) runs fully but then errors out with:

```
PI SENDER: Broker disconnected
libc++abi.dylib: terminating with uncaught exception of type std::__1::system_error: mutex lock failed: Invalid argument
Abort trap: 6
```
Tested with:
`develop`  at ff2cc190fb16e2850a04468f1fb90d52d1541cf8
Anaconda python 3.6
Other dependencies (zmq, boost, cmake, etc.) installed via `brew`

bpalmintier:
Maybe related to #134

bpalmintier:
(Not related to #134) 
Correction: this only happens if using the external `helics_broker` call rather than the one that `pisender.py` sets up. Presumably the sender does not check if its broker got created and then later tries to stop it even if not started.

Full listing if `helics_broker 2` already running:
```
$ python pisender.py
PI SENDER: Helics version = 1.0.0.alpha (01-18-18)
Creating Broker
binding error on reply socket sleeping then will try again 
zmqError binding rep port tcp://127.0.0.1:23404Address already in use
Created Broker
Checking if Broker is connected
Checked if Broker is connected
PI SENDER: Value federate created
PI SENDER: Publication registered
PI SENDER: Entering execution mode
PI SENDER: Sending value pi = 3.142857142857143 at time 5.0 to PI RECEIVER
PI SENDER: Sending value pi = 3.142857142857143 at time 6.0 to PI RECEIVER
PI SENDER: Sending value pi = 3.142857142857143 at time 7.0 to PI RECEIVER
PI SENDER: Sending value pi = 3.142857142857143 at time 8.0 to PI RECEIVER
PI SENDER: Sending value pi = 3.142857142857143 at time 9.0 to PI RECEIVER
PI SENDER: Federate finalized
PI SENDER: Broker disconnected
libc++abi.dylib: terminating with uncaught exception of type std::__1::system_error: mutex lock failed: Invalid argument
Abort trap: 6
```

bpalmintier:
Note: as of cf5c49dffc4e357adb06886c92fdb4d2a6f175b1 the swig/python directions no longer use `helics_broker` so this is not normally seen.

kdheepak:
I'm adding @phlptp to the conversation as well, I've seen this too.

phlptp:
[...]

Issue #134, opened 2018-02-01
segmentation fault during federate disconnect or destructor


jeffdaily:
On ubuntu 16 VM.  Using latest develop head.  Built and installed debug static build of helics.  Built and installed latest ns-3 examples.  Attempted to run two instances of `./ns-3-dev-git/build/contrib/helics/examples/ns3-dev-fed-sndrcv-debug`, each with a unique federate name.  (Examples are simple and do not take command-line params, so I set the federate name to argv[1] if it exists.)  By default the examples provide full level 4 debugging.  Their code is [here](https://github.com/GMLC-TDC/ns-3-dev-git/blob/helics-1.0-alpha/contrib/helics/examples/fed-sndrcv.cc).

First, ran the broker simply
`../HELICS-build-Debug-install/bin/helics_broker 2 --loglevel=4`.

Ran the two federates.  Both federates segfault after seeing
`7221-5a2bb694-6247-4ffc-a1d2-27d17aeddd82::|| cmd:disconnect from 65537`.

I ran one of the federates with gdb to get the stack trace:

```
(gdb) where
#0  0x00007ffff4971256 in std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::compare(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) const () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#1  0x000000000059fb9c in std::operator< <char, std::char_traits<char>, std::allocator<char> > (
    __lhs=<error reading variable: Cannot access memory at address 0x2d636666342d375c>, __rhs="")
    at /usr/include/c++/5/bits/basic_string.h:4989
#2  0x00000000005a0f2f in std::less<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > >::operator() (
    this=0xa57f00 <AsioServiceManager::services[abi:cxx11]>, 
    __x=<error reading variable: Cannot access memory at address 0x2d636666342d375c>, __y="")
    at /usr/include/c++/5/bits/stl_function.h:387
[...]

PR #144, opened 2018-02-03
Possible fix for issue #135.

phlptp:
This is a potential fix for issue #135 and possibly #134.  I think it deals with one possible symptom of that problem, but I am not totally convinced it totally solved the problem.  Still working on that.

### DIFF
--- a/src/helics/common/AsioServiceManager.cpp
+++ b/src/helics/common/AsioServiceManager.cpp
@@ -156,41 +156,33 @@ AsioServiceManager::LoopHandle AsioServiceManager::runServiceLoop (const std::st
                 ptr->loopRet = std::async (std::launch::async, [ptr]() { serviceRunLoop (ptr); });
             }
         }
-        return std::make_unique<servicer> (serviceName, ptr);
+        return std::make_unique<servicer> (ptr);
     }
     throw (std::invalid_argument ("the service name specified was not available"));
 }
 
-void AsioServiceManager::haltServiceLoop (const std::string &serviceName)
+void AsioServiceManager::haltServiceLoop ()
 {
-    std::lock_guard<std::mutex> servelock (serviceLock);
-    auto fnd = services.find (serviceName);
-    if (fnd != services.end ())
+    if (running)
     {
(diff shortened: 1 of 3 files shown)


phlptp:
This change is operating on the AsioServiceManager, which is basically a convenience wrapper around boost::asio::io_service.  This gets used in a couple different circumstances,  For the UDP and TCP cores specifically, and as a timer which is part of the timeout functionality in HELICS.  The cores, and brokers trigger a timer that goes off periodically to check for lost connection right now.  When I get a little further along it should also be used as part of a deadlock detection and connection timeout for the networks.  

In any case is has a couple different uses and some of those uses require an independently executing thread (timer, asynchronous Io).  this is pretty low activity so there is no reason it can't be shared across multiple uses, which is the purpose behind the service manager.    You can create a named instance of this, and get it back later if desired, or others can use the same named instance.  (sort of like a singleton but there can be more than one and they can be destroyed).  To create them running in an independent loop you need to execute the run function in a separate thread.  Since multiple places in the same process can use the same thread, there is a counter of how many different parts are using the thread and if that goes to 0 the thread halts.  

The way it worked was that when you executed that function you got back handle.  The handle did nothing except call the close function on the AsioServiceManager when it was deleted.  This required the use of a mutex to check the main map.  

Now the part I am not so sure about.  I think what is happening is that when the program shuts down, the threads managing HELICS try to shut down gracefully.  But depending on timing and the operating system some file level globals get destroyed (the order seems to operating system dependent).  So my best guess as to what was happening was that the lock protecting the system was destroyed.  then the close call came in that tried to get the lock and bad things happened.  

what the fix does is change the counter to an atomic and as part of the object itself so A shared pointer keeps it alive.  So when the close call comes in it "shouldn't" blow up because it is making a call on the object directly.  The cost was losing the ability to close the thread by name, but I was moving away from that anyway so it isn't too big of a loss.  

I think this fixes the immediate issue here but I the whole shutdown thing with the global objects being destroyed is still not completely clear and I think the C library exacerbates it somewhat as you add in library closure processing as well.

PR #146, opened 2018-02-04
Thread safety improvements

### LINKED ISSUE #85: Intermittent failure in message filter tests
```
C:/projects/helics-src/tests/helics/application_api/test-filters.cpp(106): fatal error: in "filter_tests/message_filter_function/_1": critical check mFed->hasMessage (p2) has failed
3: Failure occurred in a following context:
3:     core_type = ipc_2;
```

This failure likely has to do with timing between filters and multiple cores.  It only shows up occasionally in release mode on low core count systems with higher latency core types, so it has thus far defied most debugging attempts.  

Additional tests are needed to put further stress on the timing system as related to filters to make the bug more apparent.

phlptp:
Wasn't planning on this branch making a pull request yet, but the fact that I might have fixed #85 means this could be a fairly important to get into the code base sooner rather than later.  Can't be totally certain since that was a very intermittent failure that I could never replicate on my system, but at least theoretically I found a place where it was possible in unusual circumstances for messages to get out of order when dealing with filters on non-local federates, and I think I fixed that possibility, which hopefully remedied that issue in Travis and appveyor.  

If we go through this week and don't see that error again on the CI systems that probably means we want to make another Alpha release with the fixes later this week.


nightlark:
Out of 10 runs on Travis, I got the intermittent failure in the message federate test once on gcc 6, a double free error in value_federate_single_transfer_publisher on gcc 4.9 and 6, and a double free error in value_federate_single_init_publish on gcc 4.9

phlptp:
Based on the results of the intervening runs, it appears that this did not fix the issue,  possibly made it slightly less likely but didn't fix it.  I will leave the PR in place as it does contain some other fixes and improvements, but doesn't resolve the issue I was hoping.

CMakeLists.txt
```diff
@@ -443,21 +443,25 @@ install (TARGETS helics-static EXPORT helics-targets
     LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR})
 
 	set(helics_static_file "${CMAKE_STATIC_LIBRARY_PREFIX}helics-static${CMAKE_STATIC_LIBRARY_SUFFIX}")
-
+	set(helics_static_file_debug "${CMAKE_STATIC_LIBRARY_PREFIX}helics-static${CMAKE_DEBUG_POSTFIX}${CMAKE_STATIC_LIBRARY_SUFFIX}")
```
kdheepak: Shouldn't it build the debug version only if the debug flag is set?

phlptp: this isn't building anything it is just setting a variable with the name.  The name gets used in a configure file for importing the exported targets.  If the debug version wasn't built there is another check in those files that clear an appropriate variable.  There were just things that weren't working very well with the import depending on the versions built,  this cleaned up some of that.

phlptp: when you install helics,  take a look in the cmake directory,  you will see a few cmake files that can be used to import all the helics libraries into another project.  This is how I do it for helics-fmi and GridDyn.

kdheepak: Okay, my bad. This looks good then.

PR #147, opened 2018-02-05
Update LICENSE

kdheepak:


bpalmintier:
Agreed. We spoke with the NREL legal team and they made these suggestsions, but we certainly need to run this by all of the legal/tech transfer folks before pulling into the main code line.

bpalmintier:
As a note the main point from the NREL team was that it is totally acceptable for copyrights to be jointly held and that this seems like the right approach for this project since it was born collaboratively. As such the wording changes expand to include PNNL, LLNL, and NREL and to rework the wording to mention the government and labs in general.

In addition our legal/TTO team thought it was unnecessary to have a DISCLAIMER since that was already in the license. This is a second key point for the law/license teams to hammer out.

Sorry for having to revisit this again, there has been some change over on the NREL legal/tech-transfer space so folks are taking a fresh eye.

phlptp:
If we are going to make any changes to license I think they have to be in before we finalize the 1.0.0 release in April.  There isn't that much work in changing the code, so this is mainly legal.

phlptp:
Looks like we have formal approval from the LEGAL departments to update the license and copyright statements.   This will be something that touches every file so we should probably schedule the merge.  Based on some of the updates we probably want to do a beta.2 on thursday/friday so that might be a good time.  The license is easy.  All the copyrights on the files have a few more potential issues.

PR #149, opened 2018-02-06
Shri/c iterative example

abhyshr:
Pull request for C example with iterative API.

### DIFF
--- a/examples/CInterface/nonlings_fed2.c
+++ b/examples/CInterface/nonlings_fed2.c
@@ -59,6 +59,13 @@ int main()
   helics_publication  pub;
   int converged;
   char sendbuf[100],recvbuf[100];
+  double y = 1.0, x = 0, /*xprv = 100,*/ yprv=100;
+  int global_conv=0,my_conv=0,other_conv; /* Global and local convergence */
+  helics_time_t currenttime=0.0;
+  helics_iteration_status currenttimeiter;
+  currenttimeiter = iterating;
+  double tol=1E-8;
+  int helics_iter = 0;
 
   helicsversion = helicsGetVersion();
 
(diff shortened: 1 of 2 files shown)


examples/CInterface/nonlings_fed2.c
```diff
@@ -56,20 +86,21 @@ int main()
   vfed = helicsCreateValueFederate(fedinfo);
   printf(" Value federate created\n");
 
-  sub = helicsFederateRegisterSubscription(vfed,"testA","double","");
+  sub = helicsFederateRegisterSubscription(vfed,"testA","string","");
```
phlptp: Why did you go with publishing "strings" vs doubles?

examples/CInterface/nonlings_fed2.c
```diff
@@ -56,20 +86,21 @@ int main()
...
-
   /* Enter initialization mode */
   status = helicsFederateEnterInitializationMode(vfed);
   printf(" Entered initialization mode\n");
-  
+  double y = 1.0, x = 0, /*xprv = 100,*/ yprv=100;
+  int global_conv=0,my_conv=0,other_conv; /* Global and local convergence */
 
-  status = helicsPublicationPublishDouble(pub, y);
+  snprintf(sendbuf,100,"%18.16f,%d",y,my_conv);
```
phlptp: Oh I see,  to add a convergence flag.  
Why is the convergence flag necessary?  Do you get situations where neither if publishing but it hasn't converged yet?  I would like to understand this a little better to make sure there is no issue with the convergence inside HELICS.

phlptp:
There was a number of new warnings that showed up on the gcc 4.9 build in travis.  Those will need to be cleaned up before we can merge.

examples/CInterface/nonlings_fed2.c
abhyshr: Without passing the convergence flag, the federate would not know if the federation has converged. The force_iteration and no_iteration flags are self-explanatory. What does the iterate_if_needed flag do?

abhyshr:
Fixed. Does gcc4.9 not allow mixing declarations and code? I've seen this error with Windows build before, not for gcc.

phlptp:
I think what is going on is that gcc 5.0 switch the default C standard to c11 which is probably why we don't see these warnings in gcc 6 or clang builds.  GCC 4.9 still had the c99 standard which when combined with the -pedantic  probably generates these warnings.  My guess is we could add a -std=XXX something to the C builds which would remove the warnings.  I would be a little hesitant with that since we want the C library to be compatible with somewhat older versions but I am not sure how far back we want to go.

examples/CInterface/nonlings_fed2.c
phlptp: iterate_if_needed gives helics control over whether to iterate or not.  Basically it will iterate if any values at the current time are received (and the iteration limit has not been exceeded) otherwise it will advance time.

phlptp: maybe it should iterate_if_updated or something better named?

abhyshr:
I'd like to understand more of the iterate_if_needed flag. What happens in the following scenario for both federates:

Federate 1: Determines that it has not locally converged, publishes its value, and calls HelicsTimeIterativeRequest with iterate_if_needed flag.
Federate 2:  Determines it has converged locally, does not want to publish its value, and calls HelicsTimeIterateRequest with iterate_if_needed flag. 

Do both the federates continue iterating in this case? 

This [paper]( https://www.modelica.org/events/modelica2015/proceedings/html/submissions/ecp15118865_PetridisClau.pdf) has a few simple examples testing different co-simulation interaction protocols (Gauss-Seidel, single-iteration Gauss-Seidel, and Newton). I'm planning to implement one of the examples using HELICS. This would be a good test of iterative API with numerical integration. What do you think?

PR #153, opened 2018-02-08
Issue 150 additions

phlptp:
add missing count functions to the C api


phlptp:
@kdheepak can you help @HimanshuJain17 do the review and merge?

phlptp:
I assume that in general any modifications to the signatures of the C interface will require regeneration of the swig generated interfaces?  Is there any way we can spit out a warning if that is needed from Travis or something like that?

phlptp:
I don't mind having one of the builds have USE_SWIG=false  (probably the gcc-6 build,  but I would prefer to not add a new build to do it, and just keep the 3 current builds and have one of them have that option

kdheepak:
Are you opposed to having 3 current builds with the `USE_SWIG=false` as an separate optional build similar to the OSX builds? Because I think over time as we get more configurations I'm not sure if we can keep to just three.

phlptp:
I am expecting we will have to expand it in the future,  that is why I want to be really sure we need it before we add any.  My understanding was the swig code generation is somewhat independent of the compiler used, which is primary difference in the 3 current current builds.  And having USE_SWIG=false on any one will test any potential issues with mismatches in the repo file and the library, therefore it is unclear to me the advantage of separate builds with just that option changed.

kdheepak:
Okay, I see your point I've made the change.

phlptp:
For the USE_SWIG option, will that apply to any of the interface build options  (BUILD_PYTHON, BUILD_MATLAB, BUILD_JAVA,  BUILD_OCTAVE in the future?)

phlptp:
Now that I am looking through your code changes it looks like it is just an option for travis.  Might it make sense to have that as an option in CMAKE, so you can turn off the use of SWIG for any of the of the library builds with a single option, or turn the swig generation on.  Might even make sense to be default off.

kdheepak:
If I understand correctly, we'd just have to rename GENERATE_PYTHON to USE_SWIG.

kdheepak:
Although, if we did that, if someone wanted to build the MATLAB extension without SWIG ( since it uses a different version of SWIG ) and the Python one with SWIG, they'd have to do two different builds and installs.

phlptp:
I would say in a majority of user cases they shouldn't need to regenerate the code through SWIG, as they are just downloading the library and building it.  If I understand correctly the only time you need to regenerate the SWIG is when the public signatures on the C shared library has changed.  As developers that may be reasonably often but even this case it will likely need to be done for all interfaces otherwise there will be a mismatch.  

I am just not seeing a use case where you would want to use swig for one component and not others except in active testing of one interface, in which case requiring two builds doesn't seem like that big of deal.

phlptp:
I don't think I realized this PR would cause such a big discussion

phlptp:
it probably makes sense to have USE_SWIG=false on one of the mac builds as well

phlptp:
@kdheepak are you going to add the updated swig generated files to the Repo?  I don't really have a matlab swig version set up.

kdheepak:
I'll do this, but won't be able to get to it till Monday.

kdheepak:
This PR now has the latest Python and MATLAB files.

PR #154, opened 2018-02-08
Cmake updates and modifications

phlptp:
add code coverage support inside the cmake files @nightlark can you check if this meshed with the coverage tools you were looking at.  

Also some other cmake improvements and updates


nightlark:
I'll try it out and see what is needed to get the results into the coverage tools.

PR #164, opened 2018-02-16
MPI Core

nightlark:
The message and value federate examples communicate successfully, there's some problems with shut down that I haven't been able to work out yet -- the federates reach the end of their respective main functions, and the destructor for the MpiComms object in the broker and federates gets called, but none of the processes exit despite all of them seeming to disconnect successfully. It also looks like the RX thread closes for all 3 processes, but the TX thread only seems to close on the broker -- I didn't see anything special getting done in the other cores to insert a disconnect message in the txQueue, but then I'd still expect at least the broker process to terminate, which would make MPI force kill the other processes in the job.


src/helics/core/mpi/MpiComms.cpp
```diff
@@ -9,38 +9,319 @@ Lawrence Livermore National Laboratory, operated by Lawrence Livermore National
...
+
+bool MpiComms::mpiCommsExists = false;
+std::mutex MpiComms::mpiSerialMutex;
+int MpiComms::commRank = -1;
+
+MpiComms::MpiComms ()
 {
+    if (mpiCommsExists)
+    {
+        std::cout << "WARNING: MPIComms object already created, unexpected results may occur" << std::endl;
```
phlptp: Does this mean that there can only be one MPI comms per process?

src/helics/core/mpi/MpiComms.cpp
```diff
@@ -9,38 +9,319 @@ Lawrence Livermore National Laboratory, operated by Lawrence Livermore National
...
 #include <memory>
 
+#include <mpi.h>
+
 namespace helics
 {
-MpiComms::MpiComms (const std::string &brokerTarget, const std::string &localTarget)
-    : CommsInterface (brokerTarget, localTarget)
+
+bool MpiComms::mpiCommsExists = false;
```
phlptp: If you want this to actually work it probably needs to be atomic otherwise it isn't going to work in multithreaded cases

src/helics/core/mpi/MpiComms.h
```diff
@@ -10,29 +10,96 @@ This software was co-developed by Pacific Northwest National Laboratory, operate
...
+std::string makePortAddress(const std::string &networkInterface, int portNumber);
+
+/** extract a port number and interface string from an address number
+@details,  if there is no port number it default to -1 this is true if none was listed
+or the interface doesn't use port numbers
+
+@param[in] address a string with an network location description i.e 127.0.0.1:34
+@return a pair with a string and int with the interface name and port number
+*/
+std::pair<std::string, int> extractInterfaceandPort(const std::string &address);
```
phlptp: Do you actually use these functions in MPI comms?  if so I would recommend using the ones in NetworkBrokerData and get rid of these definitions since they will be duplicated

src/helics/core/mpi/MpiComms.h
```diff
@@ -10,29 +10,96 @@ This software was co-developed by Pacific Northwest National Laboratory, operate
 #define _HELICS_MPI_COMMS_
```
phlptp: lets get rid of the header guards and use #pragma once

src/helics/core/mpi/MpiComms.h
```diff
@@ -10,29 +10,96 @@ This software was co-developed by Pacific Northwest National Laboratory, operate
...
+}
+#else
+namespace boost
+{
+    namespace asio
+    {
+        class io_service;
+    }
+}
+#endif
```
phlptp: As far as I can tell you are not using the boost io service stuff so we can probably remove these definitions.

src/helics/core/mpi/MpiComms.h
```diff
@@ -10,29 +10,96 @@ This software was co-developed by Pacific Northwest National Laboratory, operate
...
+    MpiComms(const int &brokerRank);
 	/** destructor*/
 	~MpiComms();
 private:
-	virtual void queue_rx_function() override;	//!< the functional loop for the receive queue
-	virtual void queue_tx_function() override;  //!< the loop for transmitting data
-	virtual void closeTransmitter() override; //!< function to instruct the transmitter loop to close
-	virtual void closeReceiver() override;  //!< function to instruct the receiver loop to close
+	int brokerRank = -1; //!< the mpi rank of the broker
+    static int commRank; //!< the mpi rank of this comm object
```
phlptp: It seems funny that this static.  Is this really a singleton class implying you can't have a core and broker instantiated in the same process?

src/helics/core/mpi/MpiComms.cpp
```diff
@@ -9,38 +9,319 @@ Lawrence Livermore National Laboratory, operated by Lawrence Livermore National
...
+{
+    if (mpiCommsExists)
+    {
+        std::cout << "WARNING: MPIComms object already created, unexpected results may occur" << std::endl;
+    }
+    std::cout << "MpiComms(" << std::to_string(brokerRank) << ")" << std::endl;
+    initMPI();
+    if (brokerRank == -1)
+    {
+        setBrokerRank (commRank);
```
phlptp: can commRank still be -1 here?

src/helics/core/mpi/MpiComms.cpp
```diff
@@ -9,38 +9,319 @@ Lawrence Livermore National Laboratory, operated by Lawrence Livermore National
...
-MpiComms::~MpiComms () = default;
+MpiComms::~MpiComms ()
+{
+    std::cout << "Disconnecting MPIComms" << std::endl;
+    disconnect ();
+    std::cout << "Finished disconnect" << std::endl;
 
-void MpiComms::queue_rx_function () {}
+    int mpi_initialized;
+    MPI_Initialized(&mpi_initialized);
```
phlptp: does this call not have to be serialized?

src/helics/core/mpi/MpiComms.cpp
```diff
@@ -9,38 +9,319 @@ Lawrence Livermore National Laboratory, operated by Lawrence Livermore National
...
+void MpiComms::serializeSendMPI(std::vector<char> message, int dest, int tag, MPI_Comm comm) {
+    std::cout << "SendMPI from " << getAddress() << " to " << dest << std::endl;
+    std::lock_guard<std::mutex> lock(mpiSerialMutex);
+    MPI_Request req;
+    MPI_Isend(message.data(), message.size(), MPI_CHAR, dest, tag, comm, &req);
+
+    int message_sent = false;
+    while (!message_sent)
+    {
+        MPI_Test(&req, &message_sent, MPI_STATUS_IGNORE);
```
phlptp: Is there any chance this gets into an infinite loop?

src/helics/core/mpi/MpiComms.cpp
```diff
@@ -9,38 +9,319 @@ Lawrence Livermore National Laboratory, operated by Lawrence Livermore National
...
-        cmd.index = CLOSE_RECEIVER;
-        transmit (-1, cmd);
+        rx_status = connection_status::error;
+        return;
     }
-    else
+    rx_status = connection_status::connected;
+
+    std::vector<char> data (10192);
+    boost::system::error_code error;
```
phlptp: not seeing these being used anywhere

src/helics/core/mpi/MpiComms.cpp
```diff
@@ -9,38 +9,319 @@ Lawrence Livermore National Laboratory, operated by Lawrence Livermore National
...
+            return;
+        }
+        if (shutdown)
+        {
+            goto CLOSE_RX_LOOP;
+        }
+
+        ActionMessage M;
+        M.from_vector (data);
+        std::cout << "message received: " << prettyPrintString(M) << std::endl;
```
phlptp: we should remove this before merging or in some sort of logging call

src/helics/core/mpi/MpiComms.cpp
```diff
@@ -9,38 +9,319 @@ Lawrence Livermore National Laboratory, operated by Lawrence Livermore National
...
+std::vector<char> MpiComms::serializeReceiveMPI(int src, int tag, MPI_Comm comm) {
+    int message_waiting = false;
+    MPI_Status status;
+
+    while (!message_waiting && !shutdown)
+    {
+        std::lock_guard<std::mutex> probe_lock(mpiSerialMutex);
+        if (!shutdown)
+        {
+            MPI_Iprobe(src, tag, comm, &message_waiting, &status);
```
phlptp: Are there any blocking MPI calls?  if so isn't holding the lock during a blocking call problematic?

src/helics/core/mpi/MpiComms.cpp
```diff
@@ -9,38 +9,319 @@ Lawrence Livermore National Laboratory, operated by Lawrence Livermore National
...
+    {
+        MPI_Test(&req, &message_sent, MPI_STATUS_IGNORE);
+    }
+}
+
+std::vector<char> MpiComms::serializeReceiveMPI(int src, int tag, MPI_Comm comm) {
+    int message_waiting = false;
+    MPI_Status status;
+
+    while (!message_waiting && !shutdown)
```
phlptp: is this a busy loop or is there some sort of waiting going on?

src/helics/core/mpi/MpiComms.cpp
```diff
@@ -9,38 +9,319 @@ Lawrence Livermore National Laboratory, operated by Lawrence Livermore National
...
+    if (rx_status == connection_status::connected)
+    {
+        shutdown = true;
+    }
+
+    tx_status = connection_status::terminated;
+}
+
+void MpiComms::closeReceiver() {
+    shutdown = true;
```
phlptp: shutdown needs to be atomic if you are going to use it like this

tests/helics/core/MpiCore-tests.cpp
```diff
@@ -12,7 +12,8 @@ Lawrence Livermore National Laboratory, operated by Lawrence Livermore National
 #include <boost/test/unit_test.hpp>
 
 #if HELICS_HAVE_MPI
```
phlptp: I think I would probably just put the conditional in the Cmake so it only adds the file if MPI is present

phlptp:
@jeffdaily   can you take a look at this as well,  I am guessing you are more familiar with MPI operations than I am

src/helics/core/mpi/MpiComms.h
nightlark: Yea, it is a singleton class since MPI assigns ranks to the process. Maybe a tag could be included to allow multiple comms objects -- possibly assigned sequentially based on the order of comm object creation?

src/helics/core/mpi/MpiComms.cpp
nightlark: Ranks are >= 0; the only exception would be if something went MPI failed to initialize.

src/helics/core/mpi/MpiComms.cpp
nightlark: I'm not using any blocking MPI calls, since MPI_THREAD_SERIALIZED doesn't allow concurrent MPI calls from different threads.

src/helics/core/mpi/MpiComms.cpp
nightlark: It is a busy loop -- do you think a call to sleep should be added in? The Iprobe call used in receive is a non-blocking check to see if any messages are waiting to be received, so if there is one Irecv can be called -- the MPI_Test call then checks to see if the call to Irecv has finished copying the received data into the buffer. This could also be restructured to get rid of the send/receive methods and instead use a separate thread that uses queues for incoming/outgoing messages and keeps all the MPI calls during normal operation in one place, eliminating the need for the mutex other than when exiting.

src/helics/core/mpi/MpiComms.cpp
nightlark: There is, if the send request never completes.

src/helics/core/mpi/MpiComms.cpp
nightlark: I'm actually not sure. MPI_Initialized is one of the few calls that can be used before MPI_Init gets called.

src/helics/core/mpi/MpiComms.cpp
nightlark: Actually, with multiple comm objects the mutex would still be needed to serialize the different threads using MPI calls.

src/helics/core/mpi/MpiComms.cpp
nightlark: I checked the MPI "standard", MPI_Initialized, MPI_Finalized, MPI_Query_Thread, MPI_Is_Thread_Main, MPI_Get_Version and MPI_Get_Library_Version are the only ones it says must always be thread-safe.

src/helics/core/mpi/MpiComms.cpp
phlptp: All the other comm types have some sort of way to wake up the thread when a message is received. Sleep isn't going to help here as that would significantly increase latency.  But the way you described it, if you can't really have multiple calls from the same process going on at the same time, I think we would need to put all the calls into a single thread, and have the busy loop deal with the transmission and reception.  

longer term you said there was another level of MPI that had some more threading capabilities but it wasn't available commonly, if that is the case then we probably want code that switches to that mode if available and falls back if not.

src/helics/core/mpi/MpiComms.h
phlptp: Not urgent for the beta but longer term I think we will need to support capabilities to have multiple HELICS cores/brokers in the same process in order to be consistent with the other core types.  Ideally switching to MPI should be mostly seamless as far the application code goes.  (startup may be another story)

src/helics/core/mpi/MpiComms.cpp
nightlark: As I've been reading more of the MPI documentation and papers on MPI threading, it seems like you're right about switching to a single MPI communication thread that gets shared by all MPIComms objects is the way to go, and to avoid multiple threads entirely. MPI's threaded mode is slower than single/funneled mode (how much depends on the system+MPI implementation) since they have locks and other synchronization happening internally with the threaded mode.

phlptp: I am guess most applications use a well defined pattern of sends and expected received, rather than an asynchronous pattern.  I am guessing this is going to take some experimentation once we get it running, especially once we get to larger scales.

nightlark:
I'm working on changes to combine the send/receive functions into their own thread to allow multiple MPI cores/brokers in a single process, and some changes to MPI initialization that will make it work better as a library that can be used in existing programs.

nightlark:
@phlptp @jeffdaily the most recent commit uses a single thread for all MPI calls and seems to be running the message federate to the same state of completion as before (disconnect messages are all received and MPI_Finalize gets called, but the processes aren't exiting).

nightlark:
The startbroker option for the examples work now with a MPI broker and cores within the same process. Terminating is still a problem. Cores disconnect but the MPI_Finalize call blocks indefinitely -- maybe it is waiting on the broker to also call MPI_Finalize?  I'm at a loss as to why the broker keeps running as if nothing has happened despite all of the connected cores having disconnected.

nightlark:
Handling pending sends now works in a way that seems to be ensuring disconnect messages reach the broker, so shut down no longer hangs -- the order that classes get destroyed in mattered a lot for shutting down without errors.

src/helics/core/mpi/MpiBroker.cpp
```diff
@@ -10,64 +10,67 @@ Lawrence Livermore National Laboratory, operated by Lawrence Livermore National
...
 
-MpiBroker::MpiBroker (bool rootBroker) noexcept : CommsBroker (rootBroker) {}
+    MpiBroker::MpiBroker(const std::string &broker_name) : CommsBroker(broker_name) {}
 
-MpiBroker::MpiBroker (const std::string &broker_name) : CommsBroker (broker_name) {}
+    MpiBroker::~MpiBroker() = default;
 
-MpiBroker::~MpiBroker () {}
+    using namespace std::string_literals;
+    static const ArgDescriptors extraArgs{ { "broker_address", ArgDescriptor::arg_type_t::string_type, "location of a broker using mpi (rank:tag)" } };
```
phlptp: I might suggest having the option of this being two separate arguments as well.  Then tag can reuse the name field if need.  Not a required change but as we evaluate and adapt this might be worth a small change to better match the other brokers

src/helics/core/mpi/MpiService.cpp
```diff
@@ -0,0 +1,314 @@
...
+namespace helics {
+
+MPI_Comm MpiService::mpiCommunicator = MPI_COMM_NULL;
+
+MpiService& MpiService::getInstance ()
+{
+    static MpiService instance;
+    instance.startService ();
+    return instance;
+}
```
phlptp: you may want this to be a shared pointer, given some of the fights I have had with the shutdown sequences on different operating systems in HELICS, if this gets used in multiple threads it is entirely possible the static variable will be deleted before the uses of it are finished, leading to termination errors and seg faults.

src/helics/core/mpi/MpiService.h
```diff
@@ -0,0 +1,73 @@
...
+    static MpiService& getInstance();
+    static void setMpiCommunicator (MPI_Comm communicator);
+
+    std::string addMpiComms (MpiComms *comm);
+    void removeMpiComms (MpiComms *comm);
+    std::string getAddress (MpiComms *comm);
+    int getRank();
+    int getTag (MpiComms *comm);
+
+    void sendMessage (std::string address, std::vector<char> message) { txMessageQueue.emplace (address, message); }
```
phlptp: should probably use std::move on the message,  otherwise I think there is an extra copy here

src/helics/core/mpi/MpiService.cpp
nightlark: If a shared pointer got added to each MpiComms object, would that prevent the instance from getting destroyed until all of the destructors for MpiComms objects finished?

phlptp: I believe so,  That is what I do in some of the similar classes.  The getInstance method won't work anymore even with a shared pointer when the threads start to close but anything that captured the pointer from earlier still has access to a valid object whereas if you just captured reference that doesn't prevent the object from getting destroyed.

src/helics/core/mpi/MpiBroker.cpp
phlptp: I think it would be a --brokername field as well.  In any case there is a name associated with a broker/core  either specified or autogenerated from a guid

nightlark: Okay. I'm not sure how well reusing it for the tag would work, since the tag has to be an integer. Maybe a hash of the name or autogenerated guid would work.

phlptp: didn't catch that it had to be an integer for the tag, thought it was a string,  in that case ignore my comment other than you should be able to specify the rank and tag separately

src/helics/core/mpi/MpiService.cpp
```diff
@@ -0,0 +1,314 @@
...
+        {
+            // Get the size of the message waiting to be received
+            int recv_size;
+            std::vector<char> buffer;
+            MPI_Get_count (&status, MPI_CHAR, &recv_size);
+            buffer.resize (recv_size);
+            
+            // Receive the message
+            MPI_Recv (buffer.data (), buffer.capacity (), MPI_CHAR, MPI_ANY_SOURCE, MPI_ANY_TAG, mpiCommunicator, &status);
+        }
```
jeffdaily: I understand you're trying to drain and discard any pending messages from MPI.  If I were being pedantic, I would say to use the status.MPI_SOURCE and status.MPI_TAG values from the MPI_Status object in the MPI_Recv(), instead of again specifying MPI_ANY_SOURCE and MPI_ANY_TAG.  There is a slight possibility that MPI could reorder the messages and you'd be receiving a different message than what probe indicated, so your size/capacity could be wrong.

src/helics/core/mpi/MpiService.cpp
```diff
@@ -0,0 +1,314 @@
...
+    }
+
+    // Finalize MPI
+    int mpi_initialized;
+    MPI_Initialized(&mpi_initialized);
+
```
jeffdaily: This check is okay, but if MPI_Finalize() was called outside of HELICS, then the above MPI_IProbe and MPI_Recv calls would likely error.

Most programs I am aware of will call MPI_Init() right inside their main(), and then MPI_Finalize() right before returning from main().  Is there any danger in HELICS calling MPI_Finalize() here?  Would it mess up an app that is also using MPI?

src/helics/core/mpi/MpiService.cpp
```diff
@@ -0,0 +1,314 @@
...
+                // Get the size of the message waiting to be received
+                int recv_size;
+                std::vector<char> buffer;
+                MPI_Get_count (&status, MPI_CHAR, &recv_size);
+                buffer.resize (recv_size);
+
+                // Post an asynchronous receive
+                MPI_Request req;
+                MPI_Irecv (buffer.data (), buffer.capacity (), MPI_CHAR, MPI_ANY_SOURCE, i, mpiCommunicator, &req);
+
```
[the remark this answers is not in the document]
jeffdaily: You're using a non-blocking receive in a blocking fashion, since you are spinning on the MPI_Test later.  You could consider using a blocking receive instead unless there is other work to do while spinning on the request.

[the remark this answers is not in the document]
nightlark: Originally the receive was happening in a separate thread from the sends and to serialize the calls it had to be non-blocking. I was thinking of sticking the receive request in an array with the buffer to be filled, then later on check for finished receives.

src/helics/core/mpi/MpiService.cpp
nightlark: It is possible, but if they're using MPI in their own program, they will want to call the sendAndReceiveMessages function as part of their own loop instead of letting HELICS create a thread (everything below MPI_THREAD_SERIALIZED wants the MPI calls to happen on the same thread as MPI_Init and MPI_Finalized get called by), or a mutex would need to get added to serialize MPI calls unless their MPI implementation has full threading support.

nightlark: I split the part that is posting receives during shutdown into its own public method, and added a check when initializing MPI to see if the application had already initialized MPI -- now when the thread is closing it will only finalize MPI if it also initialized MPI. If they're trying to use the MPI core with anything less than MPI_THREAD_MULTIPLE, they can call setStartServiceThread(false) to make so that helics doesn't spawn the thread to make MPI calls.

phlptp:
@nightlark   Ryan go ahead and merge this and close the related issues once the travis tests have passed

PR #165, opened 2018-02-20
Echo app update

phlptp:
@kdheepak you will need to update the interface code as there was a minor change in the C library.

This ended up to be a bigger project than first thought.  The Echo app which basically echos a message back to the source, brought out some issues with the timeCoordinator that I hadn't seen before, and basically forced a lot more rigor onto the timing.  Still might be some lingering issues with the other tests but is also probably related to other timing issues that occasionally popped up with the Travis tests.  It has passed all tests on my local systems now I need to see if there are any issues on Travis that still need to be dealt with.  

The major change is a differentiation in the timeCoordinators in cores/brokers with that for the federates.  The federate one has more complexity to deal with the different federate configurations.  The core/broker now functions as an aggregator, so it much simpler and has slightly different rules for signalling given its new role and not having to worry about local time requests and grants.

I also added a federateFilter example code


phlptp:
working on fixing the merge conflicts

phlptp:
There is still an issue with the travis tests causing a timeout,  in some of the filter tests.  I am looking into it, but won't get to it for a couple hours.  @kdheepak  shouldn't impact anything for the C-api,  but the issue should be addressed before the PR gets merged

phlptp:
I ended up disabling the core-tests on travis and the apps tests on appveyor.   Seems those tests though passing on local systems pretty regularly give the CI platforms trouble.

phlptp:
@kdheepak @nightlark I think this can be merged now.  pretty sure the spurious failures the application_api test that we were seeing are also fixed now.

PR #170, opened 2018-02-23
C tests

phlptp:


phlptp:
Dheepak will need to update the Python and Matlab files before merging

Issue #178, opened 2018-02-24
add caching for SWIG on Travis/appveyor


phlptp:
Given that the swig download seems to fail periodically, particularly when we are trying to get a release out.  SWIG should be cached so we are not dependent on the download site for all tests.

phlptp:
@nightlark  is this completed?

nightlark:
I think so. I'm waiting until the next commit to develop and tomorrow's cron job run to make sure the caches are updated. Also not sure that Appveyor's caching works -- we might need to host our own copy of the file.

nightlark:
Caching is also working on Appveyor.

Issue #138, opened 2018-02-02
Xcode 6.4 build on travis changed


phlptp:
The xcode 6.4 build on travis has changed to using boost 1.58.  This is causing the tests to automatically turn off and generate some error.  

the code error is easy to fix, but the other issues require some choices and change

we can either leave it at 1.58 and change some flags so it turns off the tests and doesn't run them
or change it back to a newer boost.  

We do say we support 1.58 in the core so it isn't the worst idea to have one of the builds use it

thoughts?

nightlark:
If we say we support 1.58, we should probably have at least 1 semi-regular build using boost to make sure it still compiles. I noticed the os x problem last night and was going to work on it, but worked on problem with the linux clang builds instead.

phlptp:
I am fine with leaving it at 1.58 we just need to change the flags for that build so it doesn't build the tests and doesn't run anything with ctest

kdheepak:
On osx, we can use `brew update` and then install boost 1.66 if necessary.

phlptp:
I don't actually mind using boost 1.58 in one of the tests,  probably not a bad idea if we are claiming the helics core can compile with that version.  For this build we just need to turn the test builds off so it doesn't trigger the warning message.

nightlark:
I’ll make the changes to turn off tests if we want to keep 1.58

phlptp:
I vote we keep 1.58 and drop the tests,  what is everyone elses opinion?
longer term we probably want some of the examples to run on this platform to make sure everything is working smoothly

nightlark:
We should keep it without the Boost tests, since we say that we support 1.58

PR #183, opened 2018-02-26
CI build script changes

nightlark:
* Adds caching for SWIG on Travis (#178) - also cached on appveyor, but appveyor doesn't seem to ever restore the cache; one possible solution would be finding someplace reliable to host our own copy of swig
* Adds more flexibility for specifying Boost versions for individual build configurations (#138) - uses Boost 1.58.0 for xcode 6.4, and Boost 1.61.0 for gcc 4.9
* Separates the installation of dependencies into a more general script that is used by the script for setting up dependencies specifically on Travis
* Caches less data for cmake -- removes some large files (cmake-gui) that aren't needed
* Removes use of sudo (only command using it wasn't needed), builds start up faster, and it uses EC2 so downloading the cache should in theory be faster


phlptp:
Do we have the ability to run different tests in the different builds? Specifically the daily builds should ideally be running a more extended test suite than the CI builds.  Which requires slightly different compilation flags and additional tests,  like running the sets of examples potentially using the helics runner from Dheepak?  I like where this is going, need to look a little more at the output and configuration yet.

nightlark:
We don't have anything special for running a different set of tests yet, but something like the USE_SWIG variable could be used to an extended set of tests and using a different set of compilation flags.

I guess defining tests isn't necessarily just levels like quick/extended, since running examples as a test could be independent of running the quick/extended set of tests or running things using the helics runner, though some sets of tests like quick/extended might have compilation flag requirements that make them unusable together (though quick is a subset of extended).

phlptp:
For example right now the travis tests run on the CI builds, which is fine,  But on the daily build I would like be running the application_api_tests instead potentially built without the QUICK_TEST_ONLY flag,  this will be particularly important for the coverage tests since that should ideally run all the tests in ctests to get an accurate coverage report

PR #184, opened 2018-02-27
Update java swig bindings

kdheepak:


phlptp:
Do we have any way to test the java bindings in travis?   if not how difficult would that be?

afisher1:
I'm confused by this. It looks like you are adding the built java bindings to the repository rather than having the bindings being built during src build.

kdheepak:
The idea is to allow users to build the extension without installing SWIG. There's more changes that are required here since the CMakeLists.txt file needs to be updated to allow an option to build without SWIG.

nightlark:
@phlptp it might require changing the language in the config file to java. I think since we already install our own c++ compilers for most of the builds, switching shouldn't break the existing build. As long as we stick to the pre-installed JDKs, I don't think the switch would add significantly to the build time.

phlptp:
Well I think as Andy puts together a few tests of the Java interface it would be good to have a way to test them on a regular basis either a daily run or as part of the CI.  I also think for all the language interfaces we should be able to build without needing swig as a requirement.  Python, Java, Matlab, Octave and any others we may decide we need.  I don't know what else there might be in the future.

nightlark:
It might need to be a daily run. Switching the language type to java breaks all of the builds, it looks like they are all using gcc 4.8, maybe the apt-get to install other compilers failed or something is messing up the CC/CXX env variables.

phlptp:
Well  I guess we will need to deal with the environment setup when we actually have some tests that could be run
@kdheepak sounds like there is more that needs to be done on the PR,  what do you want to do with it now?  Make the updates to Cmake in this PR,  merge it now and add stuff later?

kdheepak:
I'll make the changes to the CMakeLists.txt and get back on this.

phlptp:
I think these files are now out of date given the recent changes to SWIG #191.  We should probably regenerate the files and add them to that PR and may add the changes to cmake and close this one without merging.

kdheepak:
@phlptp I suggest we leave this PR open. I'll add the CMakeLists.txt changes here, and then rebase on to develop. Users can still use SWIG to generate the Java files, so this change is not required at the moment. We need to more #191 before Friday, since I would require those changes for the MATLAB tutorial. We can delay this PR to after Friday and we'll be fine.

kdheepak:
Yes, I'll be able to recreate the Java bindings once I rebase this branch on top of develop.

phlptp:
@kdheepak,  I propose rejecting this PR and putting a little more work into [interface_cmake_update](GMLC-TDC/HELICS-src/tree/interface_cmake_update) and maybe merging with @afisher1 's work on the JAVA tests.  I started to play with how to generate the jar files.  I also think @afisher1  said that one of the java files needed to be modified to include some line.  So we might need a python script to make that modification before packaging.  This isn't working yet, I think I still have something wrong with the custom command things since it gives me errors when I run on Visual studio.  I also haven't tested the octave stuff yet.  But it would be nice to have all the interfaces be consistent in the CMAKE configuration

kdheepak:
Okay that sounds good. I'll close this one.

PR #191, opened 2018-02-28
Interface update

kdheepak:


swig/matlab/pireceiver.m
```diff
@@ -1,4 +1,4 @@
-loadlibrary(GetFullPath('~/local/helics-1.0.0a/lib/libhelicsSharedLib.dylib'));
+loadlibrary(GetFullPath('~/local/helics-develop/lib/libhelicsSharedLib.dylib'));
```
phlptp: Just noticing that these files are specific to your local system and are specific to a mac.  Is there anything we can do to make them more generic?  and why are we loading this library vs something specific to the matlab mex file as this looks like it loads the c shared library?  I would have thought the mex wrapper would take care of that?

phlptp:
probably should update the python and Java files as well since they will all be affected before merging

swig/matlab/pireceiver.m
kdheepak: I was going to update the documentation. The only way I was able to get the MATLAB extension working was to load the shared library manually before calling any HELICS functions. This is why having a `helics-config` binary would be extremely useful.

phlptp:
@kdheepak are you planning on adding any more to this PR?

kdheepak:
No, we can merge this. I'm not sure why the tests on the Mac failed. I'm restarting them on Travis. Locally on my machine, they seem to work fine.

phlptp:
@kdheepak You can go ahead and merge at your convenience.  Do you think we should get out an update to the Beta before friday as well?

Issue #195, opened 2018-02-28
rework the test cases so the travis tests use the same files as application_api_tests


phlptp:
The tests are becoming a little more problematic to maintain as fixes are not propagated from the travis tests to the application_api tests and vice versa.  With a little rework I think the tests can be made to use the same files so that won't be an issue and we only have a single set of actual test files with with separate compilations and executable.

Issue #175, opened 2018-02-24
Convert all the value tests in the shared library tests to use fixture


phlptp:
The ValueFederate tests should be migrated over to use the test Fixtures to ensure proper clean up and more rigorous testing, now that the fixture is completed.

Issue #173, opened 2018-02-24
merge value federate tests in shared library tests


phlptp:
Currently there are two ValueFederate_test1 files that do much of the same thing.   These two tests need to be merged to a single file

PR #197, opened 2018-03-01
Test cleanup

phlptp:
clean up the tests a little,  remove travis_test and replace with near equivalent key_tests which is a subset of the application_api_tests which is a subset of the extended_api_tests.  These all get built now so can be run independently.  The CMAKE options for the different tests have been removed from CMAKE and from the CI configurations.  
The value and message test are now split into multiple files depending on which level of test they are part of rather than all mixed through each other.  

This addresses issues #195 and #173 and part of #175.


.travis.yml
```diff
@@ -140,13 +140,14 @@ script:
...
-  # 6 = travis-tests
+  # 3 = key_tests
+  # 4 = application api tests
+  # 5 = extended_api tests
+  # 6 = shared_library tests
+  # 7 = helics-apps tests
   - ctest --verbose --timeout 120 -I 0,0,0,1
   # - ctest --verbose --timeout 120 -I 0,0,0,2
-  - ctest --verbose --timeout 120 -I 0,0,0,6
+  - ctest --verbose --timeout 120 -I 0,0,0,3
```
nightlark: There probably isn't a significant advantage to this on Travis/Appveyor, but if you're okay with the same timeout being applied to each test, the command could be changed to `ctest --verbose --timeout 120 -I 0,0,0,1,3` to run tests 1 and 3.  On Bamboo, one set of tests timed out and then it continued on with the next set of tests -- it seems to make getting dashboard results easier, since otherwise the different ctest runs overwrite the report file that gets output.

phlptp: I will probably tweak it a little more later once we have a better idea of typical run times.  I suspect the common tests should complete pretty quickly so that one might be reduced.  The application_api test probably require a little longer than 120 s, but I imagine it would be platform dependent

Issue #193, opened 2018-02-28
helics-config binary


phlptp:
It'd be great if we could have a helics-config binary be generated when someone builds/installs HELICS.

helics-config --includes could return -I/path/to/helics_install/include
helics-config --lib could return -L/path/to/helics_install/lib

kdheepak:
I can tackle this.

phlptp:
@kdheepak I made some modifications to the helics-config.  I think it should mostly work now.  Need to check with a package installer yet to make sure.

PR #203, opened 2018-03-08
Add helicsConfigMain.cpp

kdheepak:
This adds a `helics-config` as a command line executable. 

```bash
$ cmake -DCMAKE_INSTALL_PREFIX=/Users/$(whoami)/local/helics-develop/ .. && make -j 8 && make install

$ helics-config --help
Usage: helics-config --prefix|--includes|--libs|--help

$ helics-config --prefix
/Users/$USER/local/helics-develop

$ helics-config --includes                                                                                                  
/Users/$USER/local/helics-develop/include
 
$ helics-config --libs
/Users/$USER/local/helics-develop/lib

```


kdheepak:
Resolves #193

src/helics/apps/helicsConfigMain.cpp
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    for (int i = 1; i < argc; ++i) {
+        std::string arg = argv[i];
+        if ( (arg == "-h") || (arg == "--help") ) {
+            show_usage(argv[0]);
+        } else if ( arg == "--prefix" ) {
+            std::cout << HELICS_INSTALL_PREFIX << std::endl;
+        } else if ( arg == "--includes" ) {
+            std::cout << HELICS_INSTALL_PREFIX << kPathSeparator << "include" << std::endl;
```
phlptp: Are you planning on this doing something in the future?  right now it doesn't look it spits out any actual information?

phlptp:
what exactly is this supposed to print out, I think I am still a little unclear as to the information desired out of this executable.

kdheepak:
In the MATLAB or Java federate, we typically need to load the shared library. If this exists, one could get the location of the library folder from this executable, and then append the `helicsSharedLib` to it.

src/helics/apps/helicsConfigMain.cpp
kdheepak: Doesn't this print to the stdout? That was the intended purpose here.

phlptp:
in that case you probably actually want to get the actual install location vs the assuming it is in lib or include off the prefix.  Which it will be in most cases but that is controllable.  and I am not sure how that will all work on an install so you may actually need to pick up relative paths and such.  possibly something for later if you think this is useful as is.

kdheepak:
I still don't understand why you think this will not work. `HELICS_INSTALL_PREFIX` is set to the value of `CMAKE_INSTALL_PREFIX` during the make process.

phlptp:
the include directory on the install is controlled by ${CMAKE_INSTALL_INCLUDEDIR}  in most cases  it is set by default to "include" but it is modifiable.  And the other problematic case is if this were packaged up as part of a binary distributable module.

kdheepak:
I see. That makes sense. So we can add more `#define` statements to the header file and make it work. What do we do for the case where it is a packaged binary distribution?

kdheepak:
See [python-config.sh.in](https://github.com/python/cpython/blob/3.7/Misc/python-config.sh.in) for a similar implementation.

nightlark:
Some of the other python sysconfig stuff seems to use environment variables and platform-specific conventions.

phlptp:
I suppose one way would be to have it search around a little for some possible relative paths

phlptp:
that might be more complicated than we need at present though

phlptp:
looks like a compile bug is showup on Appveyor that will need to be fixed as well

phlptp:
@kdheepak where are you at with this PR?

kdheepak:
I think we have to discuss how this should be done.

phlptp:
I don't think this is going to get done by the 1.0 release  we should target it for some effort over the next couple months.

phlptp:
@kdheepak can you update this with the latest develop branch and fix some of the compile errors from appveyor.  then maybe we can update this with some search functionality and get it integrated.

kdheepak:
I've rebased this to the latest develop and fixed the merge conflicts. We may not need this if we support [pkg-config](https://people.freedesktop.org/~dbn/pkg-config-guide.html). Thoughts?

kdheepak:
Should I go ahead and merge this?

phlptp:
Can you actually try it and make sure it prints out what you might expect for it to be useful?

phlptp:
It it does that the I think you can do a squash and merge

kdheepak:
I'm getting this error when I try to build. 

```
CMake Error at /Users/$USER/miniconda3/share/cmake/ZeroMQ/ZeroMQConfig.cmake:44 (include):
  include could not find load file:

    /Users/$USER/miniconda3/share/cmake/ZeroMQ/ZeroMQTargets.cmake
Call Stack (most recent call first):
  config/cmake/addZeroMQ.cmake:15 (find_package)
  CMakeLists.txt:248 (include)


CMake Error at /Users/$USER/miniconda3/share/cmake/ZeroMQ/ZeroMQConfig.cmake:46 (get_target_property):
  get_target_property() called with non-existent target "libzmq".
Call Stack (most recent call first):
  config/cmake/addZeroMQ.cmake:15 (find_package)
  CMakeLists.txt:248 (include)


CMake Error at /Users/$USER/miniconda3/share/cmake/ZeroMQ/ZeroMQConfig.cmake:47 (get_target_property):
  get_target_property() called with non-existent target "libzmq".
Call Stack (most recent call first):
  config/cmake/addZeroMQ.cmake:15 (find_package)
  CMakeLists.txt:248 (include)


CMake Error at /Users/$USER/miniconda3/share/cmake/ZeroMQ/ZeroMQConfig.cmake:48 (get_target_property):
  get_target_property() called with non-existent target "libzmq-static".
Call Stack (most recent call first):
  config/cmake/addZeroMQ.cmake:15 (find_package)
  CMakeLists.txt:248 (include)



```

phlptp:
interesting,  what else is in that directory share/cmake/ZeroMQ/?  
It would seem like a bug on whoever put that Config file in there to not include the target file along with it

kdheepak:
Works great!

screen shot 2018-05-23 at 3 46 31 pm

 Do we need both `--std` and `--flags`?

kdheepak:
Do we also need both `--install` and `--prefix`?

phlptp:
strictly speaking no,. but std is inherent flags could include stuff not strictly needed for any linking application.  prefix and install could be different if you oddly configured the install.  I can double check that

PR #206, opened 2018-03-09
Remove trailing whitespace

kdheepak:
Also add empty line at the end of a line. This helps with cleaner diff.


phlptp:
Did you just add a new line at the end of every file.  I noticed a few that now seemed to have two new lines now.

kdheepak:
I removed all changes to the `ThirdParty` folder. I also removed changes to the `swig/matlab/+helics/*` and `swig/python/helics_wrap.c`, since they are autogenerated. It should have only added one new line to the end of the file (I strip files of all new lines at the end of a file, and then add just one new line.)

PR #209, opened 2018-03-10
Interface cmake update

phlptp:
This is mostly an update to the cmake code around swig.
I changed the options so there is a single DISABLE_SWIG option to turn off swig generation.  Though I think Python and Octave are the only ones this works for at present.  

Octave generation is enabled.  This works on Visual Studio,  won't work on MSYS unless Octave is built inside MSYS.  I think it will work with Mingw though I haven't been able to get HELICS built on an Independent Mingw installation, granted I haven't tried that hard.  I also haven't tried on Linux yet.  

It builds a java jar file, and make the modification that @afisher1 mentioned automatically.  Someone with a little more experience in JAVA will have to figure out if that works or not.  It does make the jar file but I don't know how to use that.  Also tweaked the .i file to work a little better with java.

I removed the boost header includes from the public interface files in HELICS.  Boost libraries are still required for operation but you won't need to include the boost header files.    

started migrating towards using target_include_directories instead of the generate include_directories directives.


phlptp:
@kdheepak I don't think I follow what you are doing with the python interface build without swig?  Can you explain it a little more?

phlptp:
can you check if that is still working with the DISABLE_SWIG I wasn't entirely sure what I was doing in the python interface.

kdheepak:
I had to change the `configure` step for `setup.py` to be not `COPYONLY`. It then works fine on my local machine.

phlptp:
were you actually changing something in the file with configure?

kdheepak:
Yes, I need to set the version number in the Python extension to match that in HELICS.

kdheepak:
Any reason for moving the `matlab/+helics` folder to `matlab/interface/+helics`. I'd like to leave it as `matlab/+helics` instead, because that's what swig generates by default when you run the commands in the README.md in that folder.

phlptp:
mostly to simplify moving the files over and to clearly separate the interface flles from those used by the build system.  didn't have a strong preference so we can move them back if needed.  Now that I think about the only one that gets copied over to the build directory is the .cpp file so most of my reason isn't really valid for the Matlab interface

phlptp:
for the java interface we do copy over the files since I do generate the .jar file using those files, we can play with it a little though

phlptp:
@afisher1 can you take a look at this I want you perspective on the java interface build, particularly the jar file creation, I don't really have a good way of testing or the know how to test that.

phlptp:
I am not doing a squash on this one due to the earlier merge from develop of all the copyright updates

PR #210, opened 2018-03-14
fix a few minor issues in interface and update variant library

phlptp:
fix an issue with octave and matlab interface on linux and update variant.hpp to newer version
also tweak the java interface to remove lib in front of library name in load library


swig/java/addLoadLibraryCommand.cmake
```diff
@@ -5,6 +5,7 @@ if (NOT LIBRARY_FILE)
 set(LIBRARY_FILE JNIhelics)
 else()
 get_filename_component(LIBRARY_FILE ${LIBRARY_FILE} NAME_WE)
+STRING(REGEX REPLACE "^lib" "" LIBRARY_FILE ${LIBRARY_FILE})
```
kdheepak: This should be UNIX only, right? We will still want the `lib` prefix. Alternatively, we can just add this to the documentation and ask users to load the shared library before calling any of the HELICS functions.

phlptp: I don't really know how the loadLibrary function actually works.  I know some things like that actually search for prepended versions as well.  So I don't know.  It woudn't be that hard to make it only run on UNIX, but obviously if this is needed then on UNIX it finding the name without the prepended lib so it might need to do the same in windows,  someone needed to test it out really.

kdheepak:
I think this commit does not change the `helics.java` generated by swig. I just tested this and the `helics.java` does not have the `System.loadLibrary` function.

PR #211, opened 2018-03-14
Test zmq dependency version

nightlark:
Installs tagged v4.2.3 zmq release instead of latest commit. Avoid potential breakage of zeromq version that is in development.


phlptp:
Do we know why appveyor is suddenly looking for boost in 1_65_1 directory instead of the 1_63 it normally does on Appveyor?

nightlark:
I'd say something changed on Appveyor's end -- I just tried rebuilding a commit in a branch that last used 1.63 and passed. Now it is trying to use 1.65.1 instead.

phlptp:
looks like it is just broken,  I tried with direct find of the different boost versions, skipping 1_65_1 and that failed too so 1_63_0 doesn't work anymore either as far as I can tell, I will try one more thing

PR #213, opened 2018-03-14
Install MPI dependency and build

nightlark:
* Adds options to install-dependency script for building Open MPI and MPICH
* Builds MPI core on gcc-4.9.3 configuration


phlptp:
How do we intend to use this in travis?  Daily build or part of the every commit CI?

nightlark:
I think it would be good to have a configuration that at least builds the MPI core for commits made to develop/master (same with other cores/interfaces).

phlptp:
can we have just one configuration that does that?   or do they all need to build mpi for that to work?  If it is just one, I am fine with adding a 4th configuration for that.

nightlark:
Right now it is configured to build with MPI for only the gcc-4.9.3 configuration on each commit, regardless of the branch. That can be changed to limit it to only building with MPI if the commit (or PR) is to the develop/master branch -- though we should have a way to turn on MPI for all configurations (Linux, not OS X) for certain branches -- maybe if the branch name contains the term mpi?

phlptp:
I think my preference would be on the gcc 6.3 build instead of the 4.9.4 build. since we are already stressing that one with the oldest supported build.  And there seems to be a bunch of cmake warnings coming in that should be cleared up before merging with develop.  Are there any tests we can run on Travis to verify it is working?

nightlark:
There are some examples that could be run. A test suite for testing with the MPI core would be better -- I think with non-blocking HELICS calls it would be similar to the other test suites, and give an example of how someone could integrate use of the MPI core with an existing program that uses MPI.

phlptp:
Did the clang+mpi work, and if so once you roll that back we can approve this

nightlark:
Yep, it worked.

PR #215, opened 2018-03-16
Shared test update

phlptp:
update to test cases and some additional functions in the shared library


nightlark:
Appveyor is giving some errors: `$<COMPILE_LANGUAGE:...> may not be used with Visual Studio generators.` The CMake docs say that "portable buildsystems would not use this expression, and would create separate libraries for each source file language".

Issue #201, opened 2018-03-04
Add build to travis using clang 5.0 using --std=c++17


phlptp:
I think we should add a new build to travis that runs with the C++17 flags.  
This will require building boost with cpp17 flags and turning on some options in HELICS.

I want to make sure we work with the different settings and I think this is worth adding a new build to Travis.

phlptp:
This will be dependent on #196

nightlark:
Can it be added to the daily builds?

Issue #196, opened 2018-03-01
Add option to build boost with different toolsets


nightlark:
Need a way to override the default toolset used to build boost.

phlptp:
At one point I had some scripts in the scripts directory that built boost, looks like dheepak removed them for some reason.  But they were part of a set of scripts I am working on to build a whole bunch of different configurations and compilers.  The way I ended up building boost with the different compilers was actually remapping the compilers to the different ones for example mapping clang++ to clang++-4.0 along with all the other associated programs.  The same with GCC.  This does require sudo access.  The other option seemed to be writing some new configuration in bjam files, which I tried but could never get to work

kdheepak:
Sorry about that! I saw they weren't being used and removed them. I probably misunderstood what they were for. I can add them back in if required.

phlptp:
We can probably put them in a separate folder if would make things cleaner,  I use them as part of some scripts I was working on with google cloud to run a large number of different configurations, but I haven't put most of that into the repo yet.  but we do refer to some of those scripts in the documentation and presentations so it would probably be good to add them back in some fashion

nightlark:
It's too bad they can't pick up and use what's in the CC and CXX vars like most other build systems.

phlptp:
is it possible to just create gcc, g++, ld, etc symlinks in the local directory(or directory you call b2 in) that overrides the one in /usr/bin then you can just use with-toolset=XXXX
or the same with clang,  then you might not have to modify anything in the bjam files

nightlark:
Symlinks might work. Or if needed, adding the directory with the symlinks to the start of PATH.

nightlark:
I tried the symlink trick with two versions of gcc on LC, it also requires adding the directory with the symlink to PATH. On LC only the --with-toolset option needs changing, since the module loading of various compilers takes care of updating the system PATH already; Travis and other systems might need to create a symlink to make sure Boost uses the right version.

PR #216, opened 2018-03-19
Added CI compilation options

nightlark:
* Boost toolset can be set to gcc/clang/icc based on the compiler being used (#196)
* C++17 flag can be set for building HELICS (and Boost)
* Adds a daily clang-5 build with C++17 standard (#201)
* Adds a daily gcc-5 build
* install-ci-dependencies.sh can be used on non-Travis systems by passing it the install directory for dependencies as an argument -- python 3 needs to already be installed


scripts/install-dependency.sh
```diff
@@ -147,9 +148,43 @@ install_cmake () {
...
 install_path=$3
+
+compiler_toolset=$4
+if [[ -z $compiler_toolset ]]; then
+    compiler_toolset=gcc
+fi
+
+if [[ "$CXX_STANDARD" == 17 ]]; then
+    echo "Install dependency with C++17 flag requested"
+    cmake_cxx_standard_option=-DCMAKE_CXX_STANDARD=17
```
phlptp: is this variable used anywhere?  codacy is showing a warning about it

nightlark: It was for any dependencies using cmake as their build system - but that was just zeromq, but it didn’t like getting compiled with c++17, and is being used as a shared library anyway.

.travis.yml
```diff
@@ -134,14 +160,14 @@ install:
 
 script:
   - mkdir build && cd build
-  - export TRAVIS_DEPENDENCY_DIR=${TRAVIS_BUILD_DIR}/dependencies
-  - export HELICS_DEPENDENCY_FLAGS="-DZeroMQ_INSTALL_PATH=${TRAVIS_DEPENDENCY_DIR}/zmq -DBOOST_INSTALL_PATH=${TRAVIS_DEPENDENCY_DIR}/boost"
+  - export HELICS_DEPENDENCY_FLAGS="-DZeroMQ_INSTALL_PATH=${CI_DEPENDENCY_DIR}/zmq -DBOOST_ROOT=${CI_DEPENDENCY_DIR}/boost"
```
phlptp: probably need to tweak this line to get the boost path to work correctly again using `BOOST_INSTALL_PATH` instead of `BOOST_ROOT`

PR #226, opened 2018-03-27
Build Java interface on Travis + Darwin Boost.Build fixes

nightlark:
* Builds the Java interface on Travis
* Changes some bash script syntax to make behavior on Darwin consistent with Linux (ex - quotes surrounding variable names in case statements)
* Further attempts to get Boost to build with the right toolset based on the compiler in use -- seems to work okay now, at least no more undefined symbol errors on OS X
* Adds an environment variable that can be set to enable debug info for the dependency installation script


scripts/install-dependency.sh
```diff
@@ -119,8 +123,13 @@ install_boost () {
     wget --no-check-certificate -O ${boost_version_str}.tar.gz http://sourceforge.net/projects/boost/files/boost/${boost_version}/${boost_version_str}.tar.gz/download && tar xzf ${boost_version_str}.tar.gz
     (
         cd ${boost_version_str}/;
-        ./bootstrap.sh --with-toolset=${boost_toolset} --with-libraries=date_time,filesystem,program_options,system,chrono,timer,test;
-        ./b2 -j2 link=shared threading=multi variant=release cxxflags=${BOOST_CXX_FLAGS} > /dev/null;
+        ./bootstrap.sh --with-libraries=date_time,filesystem,program_options,system,chrono,timer,test;
```
phlptp: I don't know if we will ever build the matlab or octave interface on Travis, but those default to requiring the boost static libraries.  Nothing needed right now but something to keep in mind.

phlptp:
I want to see this build on the mac builds before approving

scripts/install-dependency.sh
nightlark: I'm not sure matlab on Travis will ever be possible, unless MathWorks collaborates with Travis to make it available. Octave is a bit of a big install, but can work.

Which of the cmake options USE_BOOST_STATIC_LIBS, BOOST_STATIC, and Boost_USE_STATIC_LIBS are needed?

nightlark: First result for matlab on travis ci gives a job posting by MathWorks -- so they might be working on a way to integrate with Travis, but there aren't many details (cost, separate service, etc). I might have found a project that uses matlab on Travis by hosting their own special image with a secure variable that probably protects their license key -- but that setup would be hard to maintain and might not be able to build it for pull requests.

phlptp: I was not expecting matlab or octave to be done on travis at least not anytime soon, but if you are using some of the same scripts on bamboo or other places  that is more where this comment would play a role.  As far as options go
the only public variable is USE_BOOST_STATIC_LIBS  to force using static libs.  Using that or BUILD_MATLAB or BUILD_OCTAVE automatically triggers Boost_USE_STATIC_LIBS and BOOST_STATIC is the variable that gets set in the config file for use in HELICS.

nightlark: So to build a copy of HELICS with matlab/octave, I'd need to build boost with link=static, and then run cmake with USE_BOOST_STATIC_LIBS=ON, and BOOST_INSTALL_PATH=<boost_install_dir>?

phlptp: I don't think you would need to set `USE_BOOST_STATIC_LIBS`.  the cmake scripts would set `Boost_USE_STATIC_LIBS` to default to ON if `BUILD_HELICS` were set to ON

phlptp: What are we expecting the typicall build times to be once boost gets cached

phlptp:
I would do a merge request

scripts/install-dependency.sh
nightlark: Typical build times should be about the same as before. Building the Java interface didn't seem to impact build time by much -- at least not a noticeable impact given some of the large variations in build times that already exist.

PR #227, opened 2018-03-27
Thread safety improvements

phlptp:
This is another part of the thread on this branch.  There are some changes here that should make the fixes to the filters work a little nicer, which is why I am submitting a PR now.  

The goal is to isolate the core thread to minimize the number of syncrhonization locks between the API facing calls and the processing loop.  Previously there was locks taken on pretty much every federate interaction.  This PR partitions the internal federate object maps between the API callable functions and the internal loop.  A number of other things were fixed along the way to getting there. 

The future iteration of the branch will do the same thing to the interface object representations.  which should reduce the number of locks taken quite a bit further.  

I have said this before but it may have fixed some of the long standing issues that occasionaly come up on travis and appveyor.


phlptp:
@nightlark  can you resolve the conflict once you merge #226,  I am not totally clear on what the resolution is right now.

phlptp:
A number of the file changes are library updates to cereal and libguarded which are merges from the upstream source.  

Another change is more work on the integration of named_points and bool into the helics well supported types.  Additional tests are required to verify operation and then add the functionality to the C interface

nightlark:
Yea, I'll resolve the conflict. I want the last mac build to finish to make sure the cache gets updated. Huge backlog on Travis for OS X builds right now.

phlptp:
@kdheepak can you take a look at this, see if you notice any issues?

PR #232, opened 2018-04-02
Update python tests suite

kdheepak:


kdheepak:
Some functions in the SWIG Python interface need to be updated for use in Python. I've marked these tests to be skipped for now.

phlptp:
Do you have a list of functions which are not working? and have you tried anything in the *.i files to get them working?

kdheepak:
`helicsPublicationGetKey` and `helicsSubscriptionGetKey` were also having an issue. It needs to return a string instead of taking a `str` as an argument.

phlptp:
cstring.i is already included in the helicsPython.i  but I think you saw that.  Anything specific for python should be in that file.

swig/helics.i
```diff
@@ -17,10 +17,13 @@
 %apply helics_time_t *OUTPUT {helics_time_t*};
 %apply (char *STRING, size_t LENGTH) { (const char *data, int len) };
 %cstring_output_maxsize(char *str, int maxlen);
+%cstring_bounded_output(char* str, 1024);
```
phlptp: cstring stuff has to be in the helicsPython.i since java doesn't work with cstring.i

kdheepak:
Merged into #233 

screen shot 2018-04-04 at 10 23 37 am

PR #233, opened 2018-04-03
App updates

phlptp:


phlptp:
we may want to merge this one before #232 since it has some other fixes to the python interfaces and a few minor updates to the C library

kdheepak:
I'll make sure I rebase that branch on top of this after this is merged, to check if everything works.

PR #238, opened 2018-04-04
Fix swig interfaces

kdheepak:


nightlark:
I think the generated swig bindings in the repo for Java needed updating.

phlptp:
@nightlark, I think we go the pr to pass on travis, we should probably merge it then finish up the last few items

PR #240, opened 2018-04-09
Create HELICS installer and artifact on AppVeyor

nightlark:
Install artifact is available at: https://ci.appveyor.com/project/nightlark/helics-src/build/artifacts
Appveyor can upload artifacts to a variety of places with either a deploy section in the config file, or by running curl/ftp to upload files to wherever we want them.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -29,6 +29,9 @@ build:
   verbosity: minimal
 
 after_build:
+  - cd build
+  - cpack
+  - cd ..
 
 test_script:
   - cd build
@@ -45,6 +48,7 @@ test_script:
   - ctest -C Release --verbose --timeout 90 -I 0,0,0,2
   - ctest -C Release --verbose --timeout 240 -I 0,0,0,4
   #- ctest -C Release --verbose --timeout 90 -I 0,0,0,7
+  - cd ..
 
 artifacts:
-
+  - path: build/Helics-1.0.0-beta.2-win64.exe
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -30,7 +30,7 @@ build:
 
 after_build:
   - cd build
-  - cpack -R ${APPVEYOR_BUILD_VERSION}
+  - cpack -B $PWD/installer-output
   - cd ..
 
 test_script:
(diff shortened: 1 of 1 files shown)


appveyor.yml
```diff
@@ -45,6 +48,7 @@ test_script:
   - ctest -C Release --verbose --timeout 90 -I 0,0,0,2
   - ctest -C Release --verbose --timeout 240 -I 0,0,0,4
   #- ctest -C Release --verbose --timeout 90 -I 0,0,0,7
+  - cd ..
 
 artifacts:
-
+  - path: build/Helics-1.0.0-beta.2-win64.exe
```
phlptp: This will likely break when we change the version number, since cpack file creation is based on the CMAKE HELICS_VERSION variable.  It would be ideal if this can key off that, but I would settle for making the name here a variable that can be updated whenever the version number inside the cmake is updated

kdheepak: I have the same comment, we shouldn't hardcode.

nightlark: Can the CMake script set an environment variable with the output name of the installer?

kdheepak:
We'd have to make this file a templated file, see other .in files as an example.

appveyor.yml
phlptp: `set( ENV{HELICS_VERSION} ${HELICS_VERSION} )`
would set an environmental variable to the helics version code

not sure how you use that in appveyor but the -win64.exe should be pretty constant

phlptp:
if you want the full file name from cmake that would probably be a little different

appveyor.yml
nightlark: I tried setting the environment variable using CMake, also with PARENT_SCOPE -- it doesn't seem to be setting the environment variable for the script that ran the cmake command.

appveyor.yml
```diff
@@ -30,7 +30,7 @@ build:
 
 after_build:
   - cd build
-  - cpack
+  - cpack -R ${APPVEYOR_BUILD_VERSION}
```
phlptp: Is this variable something appveyor creates?

nightlark: It's a variable created by appveyor, I'm hoping it is set by `version: 1.0.0-beta.2.{build} `

appveyor.yml
phlptp: Probably the script just picks up environment variables that were present when the script was started.

PR #241, opened 2018-04-09
Cloning filter update

phlptp:
This deals with cloning filters much better and can handle remote destination filters, which was kind of a pain and I had been putting off dealing with for a while.  It required some timing tweaks to the time-coordinators.  

This may also resolve a few more issues with spurious failures in the CI tests.


phlptp:
I think I partially added some of the python tests which is why that is failing on the push

PR #245, opened 2018-04-11
New examples

phlptp:
Adding c examples for message federates and combination federates. The goal is to have them be able to interact with the C++ equivalent.  I haven't run full tests yet.

### DIFF
--- a/examples/CInterface/CMakeLists.txt
+++ b/examples/CInterface/CMakeLists.txt
@@ -69,7 +69,7 @@ target_compile_options(hello_world_sender PRIVATE -ansi -Wall -pedantic -Werror)
 target_compile_options(hello_world_receiver PRIVATE -ansi -Wall -pedantic -Werror)
 endif()
 
-
+add_subdirectory(messageFed1)
 
 
 foreach(keyfile IN LISTS KEY_LIBRARY_FILES)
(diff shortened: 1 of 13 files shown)


nightlark:
Should an entry be added to the cmake file for copying the run[3].sh files to the build directory? Some of the examples have the scripts copied, others don’t.

phlptp:
Someday I would like to make sure those scripts are actually working.  I think in this case I just copied from another directory,  I guess we probably should delete them since I don't think they will work at present.  And they are unix specific.  We should probably invest some time in the examples repo and make sure the scripts work and everything runs smoothly when that is operational.  What is your opinion should the *.sh be deleted or just left and we will deal with them later?

nightlark:
I think dealing with paths and the directory to execute the run.sh scripts from is similar to the problem of running the examples as part of the test on Travis, except spread out over a large number of nearly identical files - lots of things can break from a small name or location change. Having a cleaner solution in an examples repository would be better than having a bunch of outdated scripts lying around in this one.

Most of the run.sh scripts don’t seem to work on LC without changing the -name and -target arguments to —name and —target.

I just got an idea for running the examples we keep in this repo on Travis without dealing with hard-coded paths.

PR #265, opened 2018-04-18
Thread safety improvements

phlptp:
pull request to trigger mac builds,  should not be merged


phlptp:
@kdheepak if we are going to do a bug fix release this weekend this should be merged as it does fix some issues and clears up the Xcode 7.3 issue

PR #283, opened 2018-04-26
Add python2 generation

phlptp:
add python2 interface generation to HELICS.  The generation is incompatible with the python 3 interface so only can be active at a time.


swig/python2/python_maps2.i
```diff
@@ -0,0 +1,132 @@
...
+      PyErr_SetString(PyExc_ValueError,"List elements must be numbers");
+      free($1);
+      return NULL;
+    }
+  }
+}
+
+%typemap(argout) (const double *vectorInput, int vectorlength)
+{
+}
```
shwethanidd: Question: Why is this an empty function?

swig/python2/python_maps2.i
```diff
@@ -0,0 +1,132 @@
...
+
+%typemap(freearg) (const double *vectorInput, int vectorlength) {
+   if ($1) free($1);
+}
+
+// typemap for vector output functions
+
+%typemap(arginit) (double data[], int maxlen, int *actualSize) {
+  $1=(double *)(NULL);
```
shwethanidd: Question: Why is $1 set to NULL?

phlptp: Mostly to prevent compiler warnings for uninitialized values.  in swig inputs $1, $2, etc are references to the actual objects that go into the C call which in this case is double *.  We actually set it to something in the (in) typemap but that can error, and if that happens then if $1 was not set to NULL it could attempt to free it, which would be bad.

swig/python2/python_maps2.i
phlptp: these are not functions (per se) they are small segments of code that happen in specific parts of a process.  So this section (argout) basically states that there is no output related to these arguments.

Issue #284, opened 2018-04-26
updates to minDelta and Period not acknowledges after execution entry


phlptp:
updates to minDelta and period in a federate after entry to execution mode are not acknowledged and recognized in future time grants.  Bug reported by @afisher1.

PR #285, opened 2018-04-26
Additional timing tests

phlptp:
Fix for #284,  the configure message was not being processed when stuck in the queue, also deleted some commented out code.

### DIFF
--- a/src/helics/core/FederateState.cpp
+++ b/src/helics/core/FederateState.cpp
@@ -1036,40 +1036,7 @@ iteration_state FederateState::processActionMessage (ActionMessage &cmd)
         }
 
         break;
-        /*
-    case CMD_REG_DST_FILTER:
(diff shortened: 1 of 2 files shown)


src/helics/core/FederateState.cpp
```diff
@@ -1087,6 +1054,9 @@ iteration_state FederateState::processActionMessage (ActionMessage &cmd)
             return iteration_state::next_step;
         }
         break;
+    case CMD_FED_CONFIGURE:
+        processConfigUpdate(cmd);
+        break;
```
kdheepak: Shouldn't we add a `default` case here?

phlptp: there is one, it is just at the top of the function
```
case CMD_IGNORE:
    default:
        break;
```

PR #287, opened 2018-04-27
V1.0.2

phlptp:


.travis.yml
```diff
@@ -191,7 +191,7 @@ script:
     # - echo "$LD_LIBRARY_PATH"
     # - echo "$DYLD_FALLBACK_LIBRARY_PATH"
   - python3 -m helics
-  - python3 -m pytest -v ../tests/python_helics
+  - python3 -m pytest -sv ../tests/python_helics
```
kdheepak: Do we want this to always show captured stdout and stderr? It could make the tests verbose.

phlptp: at least until we fully resolve the issues with the python tests,  doesn't seem to output much right now

PR #288, opened 2018-04-27
fix to allow recorders to terminate

phlptp:
this allows recorders with no defined stop time to terminate if everything connected has halted.


kdheepak:
Do the other apps need this as well? The player for example?

phlptp:
It would have impacted the tracer, but the fix is at the timeCoordinator level so it applies to that as well.  It only seemed to apply if a stop time was not specified and the message timing worked out in a particular way from the other federates.

PR #297, opened 2018-05-02
Cmake configure update

phlptp:
Some fixes and improvements on the install to support HELICS-FMI/GridDyn

general improvements and simplifications to the CMAKE files.


config/cmake/addMPI.cmake
```diff
@@ -0,0 +1,44 @@
...
+  else()
+    set(MS_MPI_ARCH_DIR x86)
+   endif()
+  set(MPI_CXX_LIBRARIES "C:/Program Files (x86)/Microsoft SDKs/MPI/Lib/${MS_MPI_ARCH_DIR}/msmpi.lib")
+   set(MPI_C_INCLUDE_PATH "${MPI_CXX_INCLUDE_PATH}")
+   set(MPI_C_LIBRARIES "{${MPI_CXX_LIBRARIES}")
+	set(MPI_C_FOUND YES)
+	set(MPI_CXX_FOUND YES)
+ else()
+    find_package(MPI)
```
nightlark: It looks like find_package should work for detecting MSMPI and MPICH2 installs on Windows -- was it not working on your system?

phlptp: No, it found the mpiexec but didn't find the header files, or get appropriate compilers.  I don't think this is sufficient but the find_package(MPI) didn't seem to work very well on windows.  I am going to play with it some more as we go along.  Not entirely sure what is in place for windows was appropriate either but it got further along than just with the find_package

nightlark: Maybe a problem with the MSMPI installer not always setting environment variables. What about calling find_package(MPI) first for all systems, and then as a fallback if things aren't found, use the hard-coded default paths?

phlptp: I updated it to use  the findMPI,,  which is still a bit finicky on Windows.  Sometimes it works, other times it doesn't so well.

phlptp:
Need to correct spelling on a few names in the contributors md and add links to the Readme for contributors and roadmap.

phlptp:
@kdheepak  I made the changes and added a CONTRIBUTING.md to the root directory.  Can you take a look?

kdheepak:
I think it looks good. We can also merge CONTRIBUTORS.md -> CONTRIBUTING.md.

phlptp:
Why would we want to merge them?  they are very different purposes and I am not sure what purpose combining them would serve.  CONTRIBUTORS is an acknowledgement of the actual contributors and used software packages.  CONTRIBUTING is instructions on how new people can contribute.

kdheepak:
I was thinking it would decrease the number of files in the root directory. I also thought I saw that being done in other projects, but looking again now it seems that most have it separate (Pandas, Bokeh, NumPy etc). So we can also have them as separate files. Also, once this is merged I'll make sure that the CONTRIBUTING.md file is used as a page in the docs, so that's another reason to have it separate that I didn't consider before.

Issue #299, opened 2018-05-03
Change broker arguments to automatically detect interfaces 


phlptp:
Need a simple option to add to broker command arguments to specify that the broker should work externally.  And add automatic determination of federate interfaces based on broker arguments.

PR #301, opened 2018-05-04
Broker network specs updates

phlptp:
This improves the autodetection for the local interfaces.  For cores/brokers that link with a broker, the network should be detected based on the broker.  For root brokers.  it will still default to local, but a flag like --ipv4 or --ipv6 or --all can be specified to simplify the network connection interface.  without requiring --interface=XXX.XXX.XXX.XXX:port or something to that effect.  zmq will default to using tcp

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -100,7 +100,7 @@ jobs:
           packages:
             - clang-3.6
       env:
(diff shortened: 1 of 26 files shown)


phlptp:
Links Issue #299

phlptp:
@nightlark, @kdheepak   do you have any opinions on what the arguments should be for the different interface options,  
currently
 --local
 --ipv4
 --ipv6
 --all

PR #303, opened 2018-05-07
Named point functions

phlptp:
fully integrate named point functions into the library.  Named points are {"string", double} pair objects.  They are fully translateable to other types, and have dedicated functions in Publications and Subscriptions.  as well as the interface libraries.  Also added boolean support functions as well.  

Also added a number of tests, and realized to fully test all the translations will require quite a few more more tests, but those will come.


phlptp:
That is probably a good idea.  I should check to make sure the tests are in place for the boolean functions for the C shared library as well, and maybe add some of the functions to the C++98 API.  So a little more work on this commit yet before it is ready.  @kdheepak can you work on the python tests?

kdheepak:
I probably won't be able to get to the tests this week. Do we want to merge this and then work on the tests as part of a future update?

phlptp:
I am inclined to merge it,  The C interface tests are in place, so I say we make an issue about some additional interface tests for the new functions, and we work on that over the next couple weeks.

kdheepak:
That sounds good to me.

Issue #320, opened 2018-05-12
Move Broker executable generation code out of core directory


phlptp:
Right now the broker executable code is in the core directory, which doesn't really fit with the rest of the code base.  So the code and corresponding cmake should be moved to the Apps directory, where it fits better.  

this move should correspond with the other activities around the brokers 
link with #162

PR #323, opened 2018-05-14
Broker app

phlptp:
add brokerApp,  move some cmake around,  remove HELICSImport.cmake and move functionality into HELICSConfig.cmake


phlptp:
link with issue #320

Issue #304, opened 2018-05-07
Add getSize function to Subscription object


phlptp:
The C api makes occasional use of the size of an object to perform allocations.  This function could be made available directly from the Subscription object in a much simpler and faster way than is done currently.  

The issue is to add this functionality to the subscription object itself.

PR #330, opened 2018-05-24
Subscription update

phlptp:
adds some fixes for subscriptions and the getSize functions issue #304.  Supporting that required some other changes.


phlptp:
@nightlark this should be it for this PR.  this now includes the version number update and cleans up a few more warnings.  Can you look it over and see if you see any issues?  Then we can merge this and migrate to master

src/helics/application_api/Subscriptions.cpp
```diff
@@ -81,13 +81,98 @@ void Subscription::handleCallback (Time time)
...
+{ 
+    getAndCheckForUpdate();
+    auto out = getValue<std::string>();
+    return out.size()+1;
+}
+
+size_t Subscription::getVectorSize () 
+{ 
+    getAndCheckForUpdate();
+    auto out = getValue<std::vector<double>>();
```
nightlark: Does this still work if the publication is for a vector of `std::complex<double>` (one of the overloaded options for the `publish` method in the application api)? There might also be some confusion if a federate manually packs a vector of a different data type into the payload.

phlptp: the getValue function will engage all the translation machinery, so any type will get translated to an appropriate vector (doubles, strings, complex, complex vector.  In the case of complex vector it will translate it to a vector of doubles which will have 2x the length of std::vector<double>.    When I started this I thought it would work out a bit cleaner, but my original idea had a lot of corner cases which would fail, so I had to fall back to just doing a full translation to vector<double>

phlptp: I think I might in a later update make the getVectorSize a template function with the default be double.  Then if someone wanted to get the size of the complex<double> vector that would be possible.  But for the moment the primary use case is in the C-api which only has support for vector<double>

PR #343, opened 2018-06-06
Realtime trials

phlptp:
adds a realtime flag for federates.  This will slow down a fast federate and force grant time to federates that are being slowed down by dependencies, along with a warning.  

controlled by a flag and two parameters rt_lead and rt_lag  which control which points in time around the grant time a federate would have extra actions taken either a delay or force-grant.  This uses OS facilities for timekeeping so its accuracy at a given time point is questionable.  I would think if you need a harder real time requirement using a federate with a solid real time clock would do a little better at time keeping.  But for human real time applications this should work well.  

Some of the facilities developed here might also be used later for deadlock detection and some other features.


tests/helics/core/MessageTimerTests.cpp
```diff
@@ -0,0 +1,124 @@
...
+BOOST_AUTO_TEST_SUITE(messageTimer_tests)
+
+using namespace helics;
+
+BOOST_AUTO_TEST_CASE(basic_test)
+{
+    std::mutex mlock;
+    ActionMessage M;
+    auto cback = [&](ActionMessage &&m) {std::lock_guard<std::mutex> locker(mlock);
+    M = std::move(m); };
```
nightlark: This and lines 35-36 might be easier to read if the body of the lambda is on its own (indented) lines separate from the curly-braces.

Issue #319, opened 2018-05-11
Build C++ shared library as part of a daily build


phlptp:
This might be able to be added to one of the existing daily build.  related to issue #318

nightlark:
So that would be a build with BUILD_CXX_SHARED_LIB on, and BUILD_C_SHARED_LIB off?

phlptp:
I think once the cxx_shared_rework is integrated you should be able to do both simultaneously

PR #346, opened 2018-06-08
Build the C++ shared library on Travis and Appveyor

nightlark:
Builds the C++ shared library (re: #319) for every build (seems to have minimal effect on build time, could be moved to a single build). Moves initial export of environment variables for helics cmake flags to the install section.


.travis.yml
```diff
@@ -168,11 +168,13 @@ before_install:
 
 install:
   - source scripts/install-ci-dependencies.sh
+  - export HELICS_DEPENDENCY_FLAGS=""
+  - export HELICS_OPTION_FLAGS="${OPTION_FLAGS}"
 
 script:
   - mkdir build && cd build
-  - export HELICS_DEPENDENCY_FLAGS="-DZeroMQ_INSTALL_PATH=${CI_DEPENDENCY_DIR}/zmq -DBOOST_INSTALL_PATH=${CI_DEPENDENCY_DIR}/boost"
-  - export HELICS_OPTION_FLAGS="-DBUILD_C_SHARED_LIB=ON -DBUILD_HELICS_EXAMPLES=ON -DBUILD_PYTHON_INTERFACE=ON -DBUILD_JAVA_INTERFACE=ON -DEXAMPLES_WARNINGS_AS_ERROR=ON -DPYTHON_LIBRARY=$(python3-config --prefix)/lib/libpython3.6m.${SHARED_LIB_EXT} -DPYTHON_INCLUDE_DIR=$(python3-config --prefix)/include/python3.6m/"
```
phlptp: just to be clear, the -DBUILD_HELICS_EXAMPLES=ON was removed because that is the default setting, correct?

nightlark: Yes. I think most of the flag setup will get moved later to the install section, to simplify builds with a custom script section.

kdheepak:
We can probably upload the builds back to GitHub as well, on tagged releases.

Issue #334, opened 2018-05-27
clean up errors for missing required publications


phlptp:
The errors/logs generated from missing required publications are not very clean or consistent.  This should be cleaned up.

Issue #329, opened 2018-05-23
When to error on duplicate publications


phlptp:
There are a couple situations I was running into I was wondering what peoples opinions were.

Say a federate creates a duplicate publication- the direct registration call clearly should be an error since two publications cannot have the same name.  

However, the more interesting question is the wrapper object for publications (and by extension the interface wrappers).  Those have the ability to detect an error and get a reference to the original publication, and just have two publication objects reference the same actual publication.  My inclination is to allow this and not error if the types are matching.   I was wondering what other peoples thoughts on this situation were?

PR #347, opened 2018-06-08
Duplicate errors

phlptp:
add duplicate error detection for publications and endpoints and a allow Publication objects to be duplicates to reference the same publication.  Issues #334 and #329


phlptp:
@nightlark I think this one is ready for review now, ended up being kind of a rabbit hole once I enabled errors on missing publications.

Issue #344, opened 2018-06-07
Move HELICS test cases to use boost::test labels


phlptp:
Use labels in the test cases and executables instead of having 3 separate test executables that use the same code.  
This will simplify the builds and probably make them quicker and give us a little more control in the testing

Issue #261, opened 2018-04-15
Used named tests instead of numbers in travis and appveyor


phlptp:
We are getting enough tests now that it is probably advisable to use names in the ctest calls instead of numbers.  The numbers are likely to be inconsistent and increasingly difficult to maintain so we should move to named tests for less brittleness in the CI execution.  
This will involve a few minor changes to the Travis and appveyor yml files and probably some renaming of the actual tests.

phlptp:
This is finished but will be merged after 1.2 release

Issue #260, opened 2018-04-14
Enable the python tests to be run through ctest


phlptp:
right now the python test scripts are run as separate commands.  We should migrate these to be able to run under ctest Ideally as a single ctest rather than a bunch of them, or we should move towards a non numerical based ctest execution in the CI scripts

PR #352, opened 2018-06-13
Test labels

nightlark:
* Adds named labels for ctest invocations instead of test index numbers (#261)
  * Test cases run by using the label should mostly match the tests run when using index numbers
  * Adds continuous, daily (nightly?), and release labels -- broken tests need fixing before these will work
* Adds boost test labels (#344)
  * Adds ci, daily, release labels to test cases (could get tedious, may not be something we want to do -- or it might be easier to manage with only ci/daily labels, and default to all tests otherwise; doesn't work well with using utf::disabled)
  * Single executable for application api test suite with ctest targets for running key, normal, or extended set of tests
* Adds a Python label for running the python tests using ctest (#260)


phlptp:
I think I would say the release should be all tests?    How would we trigger a release test on Travis?

nightlark:
For the release tests it could get triggered for PRs/pushes to master, unless the tests take more than ~30-40 minutes to run (I think a fresh build with all dependencies needs about 20 minutes).

tests/helics/application_api/CMakeLists.txt
```diff
@@ -39,41 +38,30 @@ heat-transfer-tests.cpp
...
-
-add_test(NAME key-tests COMMAND key-tests --log_level=message --report_level=short)
-add_test(NAME application-api-tests COMMAND application-api-tests --log_level=message --report_level=short)
-
-add_test(NAME extended-api-tests COMMAND extended-api-tests --log_level=message --report_level=short)
-
+add_test(NAME key-api-tests COMMAND application-api-tests --run_test=@key --log_level=message --report_level=short)
+add_test(NAME application-api-tests COMMAND application-api-tests --run_test=!@extended --log_level=message --report_level=short)
+add_test(NAME extended-api-tests COMMAND application-api-tests --log_level=message --report_level=short)
+set_property(TEST key-api-tests PROPERTY LABELS KeyApi)
```
phlptp: One of the other differences between the tests was the number of the test_case scenarios run in the DATA_TEST_CASES.  I think this is lost now with this setup.  I like the direction this is going but we might need to rethink how that is done since the extended test case is now very different from what it was.  see the testFixtures.hpp for the definitions.  I don't immediately know how we might deal with that other than to duplicate the data test cases and have one use the quick_test definitions, and another add the extended versions.

phlptp: with this you are defining QUICK_TESTS_ONLY for all the tests, which is different than before.

tests/helics/application_api/CMakeLists.txt
```diff
@@ -39,41 +38,30 @@ heat-transfer-tests.cpp
...
-
 target_compile_definitions(application-api-tests PRIVATE "-DTEST_DIR=\"${CMAKE_CURRENT_SOURCE_DIR}\"")
-
-target_compile_definitions(extended-api-tests PRIVATE "-DHELICS_BROKER_LOCATION=\"${HELICS_BROKER_LOC}\"")
-target_compile_definitions(extended-api-tests PRIVATE "-DHELICS_INSTALL_LOCATION=\"${CMAKE_INSTALL_PREFIX}\"")
-
-target_compile_definitions(extended-api-tests PRIVATE "-DTEST_DIR=\"${CMAKE_CURRENT_SOURCE_DIR}\"")
 target_compile_definitions(application-api-tests PRIVATE "-DQUICK_TESTS_ONLY")
 
-target_compile_definitions(key-tests PRIVATE "-DKEY_TESTS")
```
phlptp: also the number of tests run under KEY_TESTS is different than before as well, since this definition is no longer applied

tests/helics/application_api/CMakeLists.txt
nightlark: The quick tests one is a problem, since that flag is also used to define an option that can comment out certain lines of a test. There isn’t any good way to control the subset of data driven test cases run, it seems to generate the cases at compile time — maybe something could be done with manual registration to pass in an argument listing the cores to use.

tests/helics/application_api/CMakeLists.txt
nightlark: I thought I’d marked all of the cases covered by KEY_TESTS with the “key” label. Could the change in the data test case have affected the number of tests run? Another question - are the key tests intended to be a subset of the quick/regular tests? If there are some that are key test only, that could add a lot of labels.

phlptp: It is driven by the data test cases,  basically a different test set applied to the data test cases.  Ignoring the possibilities of TCP_DISABLE and no ZMQ_ENABLED.  KEY_TESTS were a strict subset.

phlptp:
I Like the direction this is going but I also think it is going to make a mess if we try to sort it all out before we do release 1.2.  So after we get the next release out, we can sort out the test cases, probably make a few more tests cases and split out the ci tests from the full tests.  I think we should drop to two levels of tests,  

Does that make it so we would only need to label the "CI" tests?  so we would only have 1 label?  

I am going to close the related issues since we have a PR for them but we won't merge this until after the 1.2.0 release is out.

nightlark:
It should be possible to bring it down to one label; for the data test cases, I think the "duplicated" tests could have all of the logic in a shared function call, then the boost macro for creating the tests could be copied and use different lists of core types (disjoint set instead of subset/superset used now).

kdheepak:
Let me know if you want me to resolve these merge conflicts.

phlptp:
I think this PR will not be part of 1.2 so we can resolve them after we get 1.2 out.  then we can rework the test cases somewhat after that point.  But this PR changes things enough that I didn't want to try to sort it out before we get the 1.2 release out.  So lets just hold off and deal with this later

phlptp:
@nightlark I removed the release/daily labels, and added CI labels and a few other changes to the tests.  There will probably be more tinkering with which tests are in the CI over time, but for the moment I am ok with it.  I don't think I got the ctest configuration right yet, and the travis test calls are not correct.  I wasn't totally clear on how that worked yet.  

The basics is that there is now a ci label.  All tests with the "ci" label should be run on the commit.  All other tests should run daily.  

We may at some point adopt some other kinds of tests that we don't run as often but I think those would be more like scaling/performance tests that I wouldn't want to run on travis/appveyor anyway.

phlptp:
@nightlark 
the same ctest command for travis doesn't work on appveyor.
I was trying to get the same tests running on appveyor as on travis but it shows up as
```
cd ..
cd build
ctest --output-on-failure --timeout 240 -L Continuous
Test project C:/projects/helics-src/build
No tests were found!!!
cd ..
```
So it doesn't seem to find any tests,  do you have any idea why it isn't finding any tests?

I am trying to track down the issue with the shared_lib cpp test on travis,  not really having all that much luck

nightlark:
I know I saw that problem with the appveyor not working with the same command before — the fix might have been removing the space and using -LContinuous instead.

phlptp:
@nightlark  Now the tests run on appveyor,  the java tests don't seem to be loading correctly on Visual studio,  Any ideas on what might be wrong.  Something about not finding the class.

phlptp:
also looks like we probably need to back off a little on the application API tests for appveyor,  they are now taking around 7 minutes

phlptp:
So as far as I can tell the issue with the cpp_shared tests is a namespace conflict issue with the cpp98 headers having the same names as the internal ones.  Not sure if that indicates something is wrong with the name hiding in linux, but it does explain why it worked just find on Windows.  I changed the cpp98 headers to be in a helics98 namespace to resolve the issue

phlptp:
we need to resolve the issue with the tests being run in appveyor yet.  and the Java tests in appveyor

nightlark:
The Java tests should work on both Travis and Appveyor now.

nightlark:
I wonder if add_jar for each of the tests might've been easier than what we are doing. It seems like UseJava should have a better way of working with individual class files.

phlptp:
yeah, beginning to wonder if the java tests on appveyor are worth it.

phlptp:
@kdheepak @nightlark looks like java on appveyor worked finally.  Looks to be an issue with the paths on Macs for finding with the helics-apps-tests for some reason.  Then we are getting close.

nightlark:
It's good that we have some idea of the platform differences with java/javac; I think anyone using Java for something more than a simple example will be using Gradle/Maven/Ant as a build tool that would take care of the differences.

nightlark:
One of the mac builds has a failed test that tried to load libzmq 5.1.3 dylib -- the copy of zmq that it picked up during configuration is 4.2.3

phlptp:
I was trying to mess around with the rpath but it didn't seem to make any difference in the mac failure.  Do we have have idea how we might fix the issue with the zmq loading?

nightlark:
I tried to figure out where it was getting a version number for zmq that doesn't exist. Maybe we can get something set up to upload the tests (or run cpack) as an artifact that Dheepak can take a look at.

phlptp:
what changed so we now are getting compiler faults?  
```
g++-4.9: internal compiler error: Killed (program cc1plus)
Please submit a full bug report,
with preprocessed source if appropriate.
See <file:///usr/share/doc/gcc-4.9/README.Bugs> for instructions.
make[2]: *** [src/helics/application_api/CMakeFiles/application_api.dir/Federate.cpp.o] Error 4
```

nightlark:
The changes that look like they affect the preprocessor are:

FilterAdditionalTests:
`#include for FilterOperations.hpp moved`
```
-BOOST_AUTO_TEST_CASE(message_random_drop_dest_object1, *utf::label("ci"))
+BOOST_DATA_TEST_CASE (message_random_drop_dest_object1, bdata::make (core_types), core_type)
```

ValueFederateTestTemplates:
`#include for testFixtures.hpp moving, addition of boost test headers`

phlptp:
Digging into this a little more I think the most likely cause that I have seen for this error is that gcc is running out of memory on the VM.    Since gcc4.9 is somewhat more memory intensive than 6 or clang this affects that compiler more than the others.  Is there any way to easily monitor memory consumption during compilation on Travis?

nightlark:
There doesn't seem to be an easy way to monitor memory consumption on Travis (several requests for the feature though) -- `watch -n 5 free` blocks the make command, and it doesn't work when put in the background. I think we would need to write a script to monitor free memory that outputs the information even when run in the background.

phlptp:
I think I resolved the compiler faults or it fixed itself.  
We still need to resolve the xcode 7.3 issue with the zmq library?  @kdheepak  any progress on that?  

I am also considering splitting up the Application API test into two program,  one for the actual api tests, and one for system tests that look at helics system issues (timing tests, iteration tests and things like that)

nightlark:
@phlptp I noticed that the build of ZeroMQ on Linux has a libzmq.5.1.3.so symlink to libzmq.4.2.3.so;  it is odd that it just broke on the os x 7.3 build

kdheepak:
I have XCode 9 on my Mac, and when I build HELICS it is using ZMQ v5 ( `/usr/local/lib/libzmq.5.dylib` ). I can download XCode7.3 to test this.

CMakeLists.txt
```diff
@@ -294,29 +295,47 @@ set(CMAKE_SKIP_BUILD_RPATH  FALSE)
 set(CMAKE_BUILD_WITH_INSTALL_RPATH FALSE)
 
 set(CMAKE_MACOSX_RPATH 1)
-set(CMAKE_BUILD_RPATH "${AUTOBUILD_INSTALL_PATH}/bin;${AUTOBUILD_INSTALL_PATH}/lib;${AUTOBUILD_INSTALL_PATH}/lib64")
+set(CMAKE_BUILD_RPATH "./;${AUTOBUILD_INSTALL_PATH}/bin;${AUTOBUILD_INSTALL_PATH}/lib;${AUTOBUILD_INSTALL_PATH}/lib64")
```
kdheepak: Why was this change necessary? I'm going to try to undo this change, and experiment with a few other changes on builds on the Mac Travis.

phlptp: It was partly to address the issue with the shared library not finding libraries located in the same directory, which was an issue with the shared_library for some of the interfaces.  basically include the current path in the RPATH.

kdheepak:
The tests fail for a different reason now. For making the previous errors go away on Travis osx, I had to run `make install` and delete the build executables.

PR #353, opened 2018-06-14
Fix setup.py

kdheepak:
Allows installation of Python extension from the command line after setup.py is configured.

Once this PR is merged, users will be able to run the install command from the build folder even if they do not install HELICS.

If users want to build the Python extension from source, and have already installed HELICS, all they have to do is create a new `CMakeLists.txt` file in the `swig/python` folder with the following lines:

```cmake
configure_file(setup.py.in setup.py)
configure_file(helics_wrap.c helics_wrap.c COPYONLY)
configure_file(helics.py helics.py COPYONLY)
```

and run the following:

```bash
cmake .
HELICS_INCLUDE=/path/to/helics-installation-v1.1.1/include/ HELICS_LIB=/path/to/helics-installation-v1.1.1/lib/ pip install -e .

```

e.g. on my Mac, I was able to run this.

```
HELICS_INCLUDE=/usr/local/Cellar/helics/1.1.1/include/ HELICS_LIB=/usr/local/Cellar/helics/1.1.1/lib/ pip install -e .
```

and use the HELICS Python extension.


phlptp:
@nightlark does this conflict with anything you were doing?

nightlark:
It changes some of the same lines, but I only altered the default values used when no environment variable exists, so it should be easy to add any that are still relevant.

phlptp:
looks like there are few issues with the non-swig build of python that should be resolved before we can merge

PR #372, opened 2018-06-26
Update python tests

### LINKED ISSUE #305: Python tests for named_point and bool functions
We need some tests in place for the named_point and boolean functions added to the C interface

kdheepak:
- Adds named_point and bool tests

Fixes #305


phlptp:
So  I think there was a combination of an actual bug in the string translation for named points, and the fact that it was getting triggered when it probalby shouldn't have been through the getStringSize function, and this only occurred in the python tests but not the other tests that did mostly the same thing.  

In any case it took a little while to find but both the actual bug, and the improper triggering are fixed now in this PR.

kdheepak:
Great job sleuthing!

PR #374, opened 2018-06-28
Fix matlab installation issues

### LINKED ISSUE #336: Add `helicsStartup.m` to swig/matlab
This will eliminate the need for MATLAB users to run `loadlibrary(GetFullPath('path/to/helics_install/lib/helicsSharedLib.so'))`

We can make this a cmake template, so it always points to the installed directory.

```matlab
% helicsStartup.m
loadlibrary(GetFullPath($CMAKE_LIB_PATH))
```

### LINKED ISSUE #361: Move HELICS-src/swig/matlab/libhelicsSharedLib.h to /path/to/install/matlab/libhelicsSharedLib.h

kdheepak:
- Fixes #336 
- Fixes #361 
- Improves documentation for MATLAB install


swig/matlab/helicsStartup.m
```diff
@@ -0,0 +1 @@
+loadlibrary(GetFullPath('./libhelicsSharedLib.dylib'));
```
phlptp: this seems mac specific,  is this going to be useful in general or just on a mac?

kdheepak: Ah shoot you are right. That is Mac specific. I'll fix it now.

docs/installation/language.md
```diff
@@ -110,6 +112,21 @@ make -j8
 make install
 ```
 
+Add the install directory path to the MATLAB files to your PATH, and then open MATLAB.
+
+Now in MATLAB, run the following:
+
+```matlab
+loadlibrary(GetFullPath('path/to/helics/install/libhelicsSharedLib.dylib'));
```
phlptp: this again is for mac only,
I will add a section on the mex setup after this is merged

swig/matlab/helicsStartup.m
phlptp: might need to be a cmake configure file

phlptp: I don't seem to need that command on windows, it mostly just finds it if it is in the same directory.  Is there something we could put in the rpath to not need this on other platforms

swig/matlab/helicsStartup.m
```diff
@@ -1 +1,2098 @@
-loadlibrary(GetFullPath('./libhelicsSharedLib.dylib'));
+# This file was automatically generated by SWIG (http://www.swig.org).
```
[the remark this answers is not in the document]
kdheepak: I think I committed the wrong file! Let me take a look again.

Issue #189, opened 2018-02-28
Test Suite for Matlab


phlptp:
We will need a test suite in Matlab that executes all functions in the shared library interface to Matlab.    The C library tests make sure that the function calls actually do what they are supposed to.  The python test suite needs to ensure those functions got translated correctly.  So it can be a subset of the shared library tests, but if we want to fully support it we need the tests running.   And it should be run regularly (daily?) In the case of Matlab, it probably isn't going to be run on Travis but potentially at one of our local CI test systems.

phlptp:
This test suite once written can probably also be used for octave once we have that interface operations with a few minor modifications since I expect the code will look very similar

nightlark:
matlab.unittest.plugins.TAPPlugin or matlab.unittest.plugins.XMLPlugin look like they can output files that Bamboo would be able to parse for checking test results.

kdheepak:
We can also run octave tests on Travis as an alternative. I'm unsure at this point whether the c Mex interface would work with octave.

phlptp:
I think the normal swig has an octave interface generation capability, so I suspect while the generated functions will be similar, the underlying code generation is probably different so will be a very different build process.

phlptp:
I am moving the development of the full test suite to 1.1, we need to create it but I also think we need some way of regularly testing it and that isn't going to work on Travis the issue is in the same group as the multimachine testing

phlptp:
this will probably parallel the effort around Octave since most of the tests should be pretty similar

phlptp:
@nightlark can we run matlab tests on bamboo?

nightlark:
Yes, Quartz has a copy of Matlab 2013a (8.1) installed (‘module load matlab’ to use). I’m asking LC support about getting a copy of 2018 installed for another project

phlptp:
Matlab did a major upgrade in the unit test capabilities in 2013b so I am leaning towards using that framework in Matlab.    I know the Matlab interface works back before then but the tests probably won't.  We will likely need a check in cmake to pick up the version of Matlab.

phlptp:
#375 is a start to completing this issue.  The filter and messageFederate tests as well as better CTEST integration need to be written yet.

phlptp:
#380 finishes this issue

PR #375, opened 2018-06-29
Matlab unit tests

phlptp:
Adds the first set of Matlab unit Tests,  fixes the getVector Failure in Matlab, and updates some documentation related to Matlab, Issue #189.

I think the rest of the tests and adding them to Ctest will come in a later revision.  But for now this tests the ValueFederate interface and makes a few fixes.  

Also included are some fixes to the python tests that I came across while using them as a template.


kdheepak:
Nice job catching these!

phlptp:
@kdheepak I am going to run a couple more checks at my office, when I get there, to make sure the vector fix in Matlab runs on multiple systems.  Then I will merge.

PR #380, opened 2018-07-03
Additional matlab tests

phlptp:
add more matlab tests, and a few tweaks to the python tests

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -74,8 +74,8 @@ else()
 endif()
 
 cmake_dependent_option(BUILD_C_SHARED_LIB "Build the Shared Libraries with a C interface" ON "NOT INTERFACE_BUILD" ON)
-cmake_dependent_option(DISABLE_SWIG "disable the use of the swigh executable to generate interface code and use repository code instead" OFF "INTERFACE_BUILD" OFF)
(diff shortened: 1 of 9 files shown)


CMakeLists.txt
```diff
@@ -74,8 +74,8 @@ else()
 endif()
 
 cmake_dependent_option(BUILD_C_SHARED_LIB "Build the Shared Libraries with a C interface" ON "NOT INTERFACE_BUILD" ON)
-cmake_dependent_option(DISABLE_SWIG "disable the use of the swigh executable to generate interface code and use repository code instead" OFF "INTERFACE_BUILD" OFF)
-	option(BUILD_CXX_SHARED_LIB "Build a Shared Libraries of the CXX interface" OFF)
+option(DISABLE_SWIG "disable the use of the swig executable to generate interface code and use repository code instead" OFF)
```
kdheepak: Why is this not a `cmake_dependent_option` anymore?

phlptp: The main reason is if it doesn't show up and you can't check it in the GUI the same time you check one of the INTERFACE_BUILD options you end up with the cmake error if it can't find swig.  This got rather annoying with the MATLAB INTERFACE BUILD on machines with out the special version of swig.  So it just simplified matters to not have that be a dependent option.

PR #384, opened 2018-07-11
Fix c api reference files for documentation

kdheepak:
- Update conf.py with correct relative path
- Update copyright symbols


kdheepak:
@nightlark I'm not sure why the appveyor build failed. I'm going to go ahead and merge because this shouldn't actually cause any issue with HELICS, these changes are only whitespace / in the docs folder.

Issue #379, opened 2018-06-30
Mac binary distribution


phlptp:
This should use the CPack stuff in cmake that is already setup
probably using the [DragNDrop](https://cmake.org/cmake/help/latest/module/CPackDMG.html) generator.

guessing a few generator specific fields will need to be added

though there might be some new generators as well if we were ok requiring a new cmake for the mac builds which isn't out of the question.  

This might be simple enough that it could be done and included with the 1.3 release.

nightlark:
[CPackPackageMaker](https://cmake.org/cmake/help/v3.12/module/CPackPackageMaker.html) has been around since at least version 3.0 of CMake and can be used to create a component-based installer similar to Windows.

phlptp:
That sounds like a good approach,  looks like there is only additional variable we would need to add.  Is PackageMaker something that is easily available(included) or something that would have to be downloaded?  @kdheepak  do you know much about it?

kdheepak:
The main thing is that the dependencies are also coupled with the installation, i.e. boost and zmq libraries should be included. If CPackPackageMaker can do that that would be ideal. I've not looked into this much yet, but I'll start to. I have been doing a bunch of reading to find what needs to be done in general on a Mac. I'll post information I find on this thread.

PR #390, opened 2018-07-20
Add macos DMG building using cpack 

kdheepak:
Addresses #379 

Run the following from the build folder to create the DMG 

```
mkdir build-osx 
cd build-osx
cmake -DENABLE_PACKAGE_BUILD=ON -DCPACK_BINARY_BUNDLE=ON -DBUILD_PYTHON_INTERFACE=ON -DCMAKE_INSTALL_PREFIX="/Users/$USER/local/helics-develop/" ../
make -j 8 && make install
cpack -C Release
```


kdheepak:
One needs to copy `Info.plist` and `HELICS.ico` to the build folder, before running `cpack -C Release`. This can be done through CMake in the configuration step.

kdheepak:
If we can compile the shared library linking the static library for ZMQ, Boost and the standard library that would be ideal. Users would need to download just one file and everything would ideally just work.

phlptp:
@kdheepak 
do you know the commands for compiling the standard library statically in clang.  I think the current code is missing that piece.  it woulds on Windows and gcc currently.  But if we knew what was required we might be able to tweak that for APPLE as well

phlptp:
if we need to do the copy step that probably should be done in cmake rather than leave it up to the users

CMakeLists.txt
```diff
@@ -532,6 +532,11 @@ else (WIN32)
 	set(CPACK_PACKAGE_ICON "${CMAKE_SOURCE_DIR}/docs/img/HELICS.ico")
 endif(WIN32)
 
+if (APPLE)
+    set(CPACK_BUNDLE_NAME "libhelics")
+    set(CPACK_BUNDLE_PLIST Info.plist)
+    set(CPACK_BUNDLE_ICON HELICS.ico)
```
phlptp: can these be set to the full path instead of needing to copy them over?

kdheepak: Yes, I will do that.

kdheepak:
This worked! I can create a dmg now using static ZMQ (that builds from source). I have the flag set for static Boost and static standard library, but I don't know if that is actually doing anything at the moment.

config/Info.plist
```diff
@@ -0,0 +1,30 @@
...
+	<key>CFBundleInfoDictionaryVersion</key>
+	<string>1.0</string>
+	<key>CFBundlePackageType</key>
+	<string>APPL</string>
+	<key>CFBundleShortVersionString</key>
+	<string>%VERS%</string>
+	<key>CFBundleVersion</key>
+	<string>%VERS%</string>
+	<key>NSHumanReadableCopyright</key>
+	<string>© 2017 HELICS</string>
```
phlptp: should this be 2018?

config/Info.plist
```diff
@@ -0,0 +1,30 @@
...
+	<key>NSPrincipalClass</key>
+	<string>NSApplication</string>
+</dict>
+</plist>
```
phlptp: Is there anything in this file that is likely to change over time (other than the copyright once a year)

phlptp:
what compiler are you using?

kdheepak:
I'm using Clang. I still need to figure out what what all the fields in the plist file should be.

phlptp:
my guess is you are not linking to the static c++ libraries in that case.

Issue #393, opened 2018-07-20
Help output for recorder doesn't list all options


phlptp:
The help output from the recorder is missing a number of options and explanations.

Need to double check the other apps

PR #394, opened 2018-07-21
App help update

phlptp:
Issue #393 update help output from apps and code-format


src/helics/apps/Player.cpp
```diff
@@ -1,857 +1,863 @@
...
-        return stringOps::removeQuotes (stringToDecode);
-    }
-    // move is required since you are returning the rvalue and we want to move from the rvalue input
-    return std::move (stringToDecode);
-}
+/*
+Copyright © 2017-2018,
+Battelle Memorial Institute; Lawrence Livermore National Security, LLC; Alliance for Sustainable Energy, LLC
+All rights reserved. See LICENSE file and DISCLAIMER for more details.
+*/
```
kdheepak: Why is every line in this file changed? Could the encoding of file have changed here when you made changes to the file?

phlptp: MIght have changed the line endings, when I did a clang-format on it?

PR #406, opened 2018-07-30
Test octave build on linux

phlptp:
Mostly some tweaks to get the octave interface compiling on Linux systems.  I ran some initial tests but a full test suite is needed in the future.  Though will probably wait until the 2.0 API is settled.


phlptp:
We already have the folder.   I am guessing it will require separate typemaps since the .oct octave API is totally distinct from the MEX API.  I think the unit test framework is different too unfortunately.  It looks like there are ways to use .mex files in octave but from what I have seen the octave API isn't too bad and probably far simpler than the corresponding mex files.    I think longer term we want to make the code look the same from the user side but the building of the interface files is very different, and we probably want to to keep it that way rather than trying to shoehorn them together.

PR #423, opened 2018-08-01
Run cmake-format and fix copyright

### LINKED ISSUE #46: Clean CMake formatting

kdheepak:
Fixes #46

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,7 +1,7 @@
 #
-# Copyright © 2017-2018, Battelle Memorial Institute; Lawrence Livermore
-# National Security, LLC; Alliance for Sustainable Energy, LLC All rights
-# reserved. See LICENSE file and DISCLAIMER for more details.
+# Copyright © 2017-2018, Battelle Memorial Institute; Lawrence Livermore National
+# Security, LLC; Alliance for Sustainable Energy, LLC All rights reserved. See LICENSE
+# file and DISCLAIMER for more details.
 #
 
 cmake_minimum_required(VERSION 3.4)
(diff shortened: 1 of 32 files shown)


CMakeLists.txt
```diff
@@ -191,387 +248,461 @@ target_link_libraries(helics_base INTERFACE Boostlibs::core)
...
-HIDE_VARIABLE(FMT_DOC)
-HIDE_VARIABLE(FMT_INSTALL)
-HIDE_VARIABLE(FMT_PEDANTIC)
-HIDE_VARIABLE(FMT_TEST)
-HIDE_VARIABLE(FMT_USE_CPP14)
+hide_variable(FMT_DOC)
+hide_variable(FMT_INSTALL)
+hide_variable(FMT_PEDANTIC)
+hide_variable(FMT_TEST)
+hide_variable(FMT_USE_CPP14)
```
phlptp: Does this work,  are macros case sensitive?

swig/java/CMakeLists.txt
```diff
@@ -11,79 +11,97 @@ include(UseJava)
...
+  if(${CMAKE_MAJOR_VERSION}.${CMAKE_MINOR_VERSION} VERSION_GREATER 3.7)
+    swig_add_library(JNIhelics TYPE MODULE LANGUAGE java SOURCES helicsJava.i)
+  else()
+    swig_add_module(JNIhelics java helicsJava.i)
+  endif()
 
-	SWIG_LINK_LIBRARIES(JNIhelics helicsSharedLib)
-	SWIG_LINK_LIBRARIES(JNIhelics ${JAVA_LIBRARIES})
+  swig_link_libraries(JNIhelics helicsSharedLib)
+  swig_link_libraries(JNIhelics ${JAVA_LIBRARIES})
```
phlptp: Does capitalization matter here for these functions

phlptp:
What is your opinion on the formatting job,  for the most part it seems to do a good job,  Some things seem a bit verbose and might be clearer on few lines, but I might be willing to live with it?

CMakeLists.txt
kdheepak: I tested it and it built fine, no errors during the `make -j 8` process.

swig/java/CMakeLists.txt
kdheepak: I think in CMake `commands` are case insensitive but arguments are case sensitive.

phlptp: I don't think I realized that was true for user defined functions as well as the built in ones.

kdheepak:
I would prefer the following:

```
set(helicsShared_headers
    helics.h
    api-data.h
    ValueFederate.h
    MessageFederate.h
    MessageFilters.h
    helicsCallbacks.h)
```

to be formatted as the following instead:

```
set(
    helicsShared_headers
    helics.h
    api-data.h
    ValueFederate.h
    MessageFederate.h
    MessageFilters.h
    helicsCallbacks.h
)
```

Because if you were going to add something to it later as the last argument, the git diff would show just the right line, as opposed to showing two lines have changed, it is more noise.

There are other small things that I would like changed. 

My opinion on the matter of formatting has been that it is easier to pick a formatter that does a good enough job to make code consistent and readable, and save many person hours arguing amongst fellow developers on the right way to format code. So that's the main reason I'm in favor of having auto formatters, and not having to worry / discuss code formatting. However, if the auto formatter produces unreadable code, then that's unacceptable. I don't think that is the case here.

We can also probably make requests for the changes we want on https://github.com/cheshirekow/cmake_format, it seems to be actively developed.

kdheepak:
They have some options that allow for setting `line-width` and `dangle-parens`. This is what I'm running the following on all files at the moment.

```
cmake-format -i {filename} --line-width 88 --dangle-parens=True
```

phlptp:
I tend to agree with you, we can live with a bit of non-idealness for an automated solution.  And I think it is a definite improvement.  

If it does this should be a separate pr, but will this run on the .cmake files in the config directory?

kdheepak:
I can update my script to make it run on the .cmake files in the config directory as well. At the moment they are not formatted.

phlptp:
lets wait until this clears,  then do that in a separate PR.  It is a lot of files to go through and those might be a little more error prone.

PR #424, opened 2018-08-02
Test fixes

phlptp:
mostly to figure out what the issue with the occasional test case failure is.


phlptp:
I think this is ready to go now.  Cleans up a few warnings on some newer compilers, and might fix a couple test case issues.

Issue #427, opened 2018-08-08
binding error on internal comms socket


kdheepak:
I'm unable to create a new broker in an interactive session repeatedly.

```python
# broker.py
import time
import helics as h

def main():
    initstring = "2 --name=mainbroker"
    # Create broker #
    print("Creating Broker")
    broker = h.helicsCreateBroker("zmq", "", initstring)
    print("Created Broker")

    print("Checking if Broker is connected")
    isconnected = h.helicsBrokerIsConnected(broker)
    print("Checked if Broker is connected")

    if isconnected == 1:
        print("Broker created and connected")

    try:

        while (h.helicsBrokerIsConnected(broker)):
            time.sleep(1)

    except KeyboardInterrupt:
        h.helicsCloseLibrary()
```

If I Ctrl+C this process, this results in the following the next time I try to call the `main` function.

```
In [2]: from broker import main
In [3]: main()
Creating Broker
binding error on internal comms socket
Created Broker
Checking if Broker is connected
Checked if Broker is connected

```

cc @bpalmintier

phlptp:
If you wait longer does it work again?  or is it permanently blocked (until python is terminated)?

phlptp:
Have you tried this on Windows and does it behave in the same way?

bpalmintier:
I am having similar issues under MATLAB, using OSX. (Also seeing `binding error on internal comms socket` errors on the command prompt used to start MATLAB's GUI, so likely coming from underlying C library?

Still working on fully repeatable sequence, but it might be something related to not unbinding from the port in ZMQ. See this [Stack Overflow Post](https://stackoverflow.com/questions/24323124/bind-publish-unbind-repeat) which though Python based, may hint at some of the same issue.

kdheepak:
This [post](https://stackoverflow.com/questions/12397175/how-do-i-close-an-open-port-from-the-terminal-on-the-mac) suggests killing the process

```
$ lsof -i :23405
COMMAND     PID     USER   FD   TYPE             DEVICE SIZE/OFF NODE NAME
python3.6 98590 $USER   34u  IPv4 0x83961aa5794b40cb      0t0  TCP localhost:23405 (LISTEN)
```

There must be a way to close the port in ZeroMQ, or if we are already doing it we have to make sure we catch these edge cases.

kdheepak:
If I try to close the library it hangs the process the next time I run main.

```
In [7]: main()
Creating Broker
binding error on internal comms socket
[...]

Issue #425, opened 2018-08-03
Ease of use updates for C/SWIG interface


bpalmintier:
Current configuration requires a few extra steeps that could be streamlined (some might propagate up to C++) for easier use. Initial suggestions:
1. Once a federate exists, it takes 3 functions to successfully get rid of it and in many cases to effectively exit the program (or get back to the command prompt in MATLAB, etc.). (terminate, free, and close library) Seems there could be a single function that does all of these and hence can easily be used to exit from error conditions, etc.
2. Must specify the number of federates at initialization. Seems this could default to 1 

Will post others as I work through some MATLAB (via SWIG) updates.

phlptp:
Terminate is a call into HELICS to disconnect a federate from the federation.  It is still queriable and usable after this call.  All created subscriptions are still usable for instance to get the last data.  

In the C library, free will release the small wrapper around a federate.  It may or may not actually destroy the underlying federate Object depending on whether it was cloned or not.  

The closeLibary call destroys everything that has been instantiated thus far and shuts down all threads.  

These are all very distinct operations, and I am a little unclear what operations should be performed in various cases if you were to make a single call.    CloseLibrary will destroy everything and shut it down though that is kind of a blunt instrument.  

The 2.0 API is still open for enhancements so any suggestions are welcome.

bpalmintier:
@phlptp I agree that these are different functionality and think we should certainly keep this fine grained control for advanced users/use cases, but in practice many federates will be standalone programs or otherwise have a simple configuration where all of these need to be performed during shut down. Having a unified function that does them all in this common sequence would make clean shutdowns and particularly multiple paths to shutdowns (such as on errors, etc.) much cleaner. See also the discussion on #429 for an example

phlptp:
In the 2.0 release we will be moving to a single output API, with an error object, so I think that will be a little more intuitive for the interface API's, and get rid of the status output.

phlptp:
The destroy function was added in #444, and will also be included in HELICS 2.0

bpalmintier:
Great!

PR #429, opened 2018-08-08
Matlab example update

bpalmintier:
I finished overhauling the pisender/receiver in MATLAB including comments, error checks, etc. and a new helper function. 

Would propose we merge this into src and then relocate them to the examples repo, since this fixes a few bugs. And before merging, also would be great if someone else can verify working on a few platforms.


bpalmintier:
Note: since no underlying code changes that impact C shared library made, the easiest way to test is copy all `*.m` files to the matlab folder in an existing install folder. Then follow directions in `pisender.m`

bpalmintier:
@phlptp:
* Yep, develop is what I meant. I have re-updated

@kdheepak: 
* The dvi file is found in the SWIG folder. I figured it came from something internal, but can remove.
* I would like to include all of the *.m files now so that anyone looking in the history gets a valid set of MATLAB functionality. After merging I think we excise the examples. I also would like to include the close Federate in the src repo for now. It seems like something that belongs in the C library, but for now seems a better fit for src, since it will be a core, common functionality for most MATLAB users.

swig/matlab/helicsEndFederate.m
```diff
@@ -0,0 +1,19 @@
...
+        warning('Non-zero status (%d) returned in call to helicsFederateFinalize. Ignoring.', status)
+    end
+
+
+    helics.helicsFederateFree(fed_info);
+end
+
+%Note: helicsCloseLibrary is unrelated to the MATLAB library loading (from
+%helicsStartup)
+helics.helicsCloseLibrary();
```
phlptp: By including this call it will destroy all existing federates/cores/ brokers.  So if you have anything else from HELICS that you would want to keep this isn't going to work.

swig/matlab/helicsEndFederate.m
```diff
@@ -0,0 +1,19 @@
+function helicsEndFederate(fed_info)
```
phlptp: wondering if helicsCloseFederate might be a better name?

phlptp:
At some point I think we will want a Matlab based class wrapper around HELICS, that would take care of some of the swig like oddities.  But I don't want to spend time on that before the 2.0 version is close to final at least as far as the API.  Otherwise it will be a lot of duplicate work.  Anyway other than the closeLibrary call everything looks good.  just not sure that won't cause more issues than the program solves.

swig/matlab/helicsEndFederate.m
phlptp: Ok,  then lets add the parameter,  add some more details in the help results of the function stating when it is appropriate to use the function, and when it might not be appropriate, and the explanation of the argument.  

Also as part #427 I will be altering the closeLibrary function to go through and disconnect or terminate the federates, core, and brokers.

swig/matlab/helicsEndFederate.m
bpalmintier: Maybe `helicsDestroyFederate` that would be symmetrical with the the various "create" actions that made it in the first place.

Incidentally, next code change (after move to examples repo) I want to make to this is to replace the programmatic build up of the federate with creating it from JSON, which seems like the paradigm we want to steer folks toward.

Overall (and after moving this code to examples) I could see there being 3 simple examples, many/all of which we might opt to duplicate for other languages over time:
1. Simple Value Federate: we have this one: `pisender/pireceiver`
2. Simple Message Federate: perhaps `hellosender`/`helloreceiver` which might also include the use of a delay filter.
3. Modestly more complex combo federate, that in particular shows the use of a value and message stream for similar data (such as physical voltage vs. measured voltage). Ideally this might also show the use of a built-in player for one of the data streams.

All of these would use helicsDestroyFederate--likely multiple times during error checks--and all would want to run helicsDestroyFederate. It is not until we get to more advanced example that we'd need to skip closing the library. Such as:
* Future example with multiple federates sharing the same broker, built up programmatically in a loop

phlptp: I can add those functions in the C interface for the 2.0 alpha release.  There would probably also be a corresponding helicsDestroyCore,  helicsDestoryBroker as well.  

You can also use TOML now if you so desire and that is easier.

bpalmintier: Renamed

bpalmintier:
Note for those coming from #425: comments around `helicsDestroyFederate` may be hidden as out of date under `swig/matlab/helicsEndFederate.m`

phlptp:
@kdheepak   any other changes you want to see otherwise I will go ahead and merge this.

kdheepak:
I think it looks good. Even no CMakeLists.txt changes are required I think, because the glob for *.m files will get all these changes.

phlptp:
there may be some issues when swig actually works to generate the matlab but I don't think that is an issue related to this PR.

bpalmintier:
All sounds good, I'll let you go ahead and merge. (Unless it is OK for me to do so)

kdheepak:
Merged! Thanks for taking this on!

bpalmintier:
Great.

PR #430, opened 2018-08-08
Add ISSUE_TEMPLATE.md

kdheepak:
This PR adds a ISSUE_TEMPLATE for GitHub issues.


kdheepak:
I wanted to merge it to `master` since it only contains the ISSUE_TEMPLATE and needs to be on that branch to show up when users create a new issue. We can merge to develop, but would have to wait till the next release to add the templates for new issues. Should I change it to `develop` instead?

phlptp:
Might have to restart some builds.  So they pass

Issue #420, opened 2018-08-01
Tsan execution


phlptp:
Set up a travis run with tsan enabled likely daily on develop.

Issue #419, opened 2018-08-01
code coverage and test cases


phlptp:
Which test cases are we running for the code coverage?  We need to be running the non-ci tests to get full coverage estimates.

nightlark:
This and #420 will be done with changes to how tests are run on Travis.

PR #436, opened 2018-08-17
Travis thread sanitizer and test execution changes

nightlark:
* Adds a thread sanitizer run to the daily builds (#420)
* Runs the full test suite for the daily code coverage build (#419)
* Uses static linking for Boost and ZeroMQ
* Names the build configurations on Travis


tests/java_helics/CMakeLists.txt
```diff
@@ -140,6 +140,7 @@ set_property(
     LABELS
     Java
     Continuous
+    Coverage
```
phlptp: I doubt the Java, matlab, python test will add much on the coverage side, so wondering if this is necessary?  Are we checking coverage of the python/ Java code?

nightlark: No, we aren't checking coverage of the Python/Java code. I think the only contribution it adds now is if those interfaces use some part of the shared library that the shared library test cases don't cover.

phlptp: I think I would prefer knowing if the shared_library tests didn't cover a particular aspect of the C library.  @kdheepak any thoughts?

phlptp:
looks like there might be some library issues with ZMQ on the mac builds for the tests

nightlark:
It looks like the same problem we had a few weeks ago. Would be nice if macOS just left environment variables alone.

Switching to static linking seems easier than trying to appease macOS dylib loading.

nightlark:
- [ ] Build dependencies using sanitizer flags
- [ ] Build static ZeroMQ library

nightlark:
@phlptp Using ZMQ and Boost as static libraries is working -- I think this will also be useful for uploading the HELICS library artifacts somewhere to use in builds triggered in other repositories (so they don't need to rebuild it again).

phlptp:
@nightlark  where is the call to b2 in the code?  
I know I have done it before with a standalone script see scripts/build_boost_gcc.sh    I have put the sanitizer flags in the cxxflags variable.  They do need to be in quotes.

nightlark:
https://github.com/GMLC-TDC/HELICS-src/blob/travis-test-changes/scripts/install-dependency.sh#L139

I used scripts/build_boost_gcc.sh for reference -- I've tried a number of ways to get the quotes in the constructed cxxflags string (escaping the quotes when constructing it, quoting the variable in the b2 line), but bash has some quirks with how it splits and passes arguments; I was getting errors from b2 about things like `PIC"` not being recognized as a file name. I know I've seen similar things in .travis.yml with passing constructed arguments to cmake.

nightlark:
@phlptp Okay, I figured it out. I had to get rid of the quotes around the value cxxflags is set to, but use quotes when expanding the constructed "cxxflags=..." string variable. I think quotes around the parameters cxxflags was set to were getting messed up by bash splitting on whitespace.

phlptp:
@nightlark are you planning on doing anything with this anymore or is it ready to merge?

nightlark:
It is either ready to merge, or we can change the set of tests run with Valgrind and possibly remove the XCode 9.x build that slipped into the config file then merge. I think the current thread sanitizer output is as good as it will get for now.

PR #457, opened 2018-09-11
MSYS2 Build Instructions

afisher1:
This PR adds documentation for windows platforms on how to build GridLAB-D and HELICS together using MSYS2.


docs/installation/windows.md
```diff
@@ -1,161 +1,263 @@
...
+
+All packages can be installed by typing the following:
+```bash
+$ pacman -S --needed base-devel mingw-w64_x86_64-toolchain mingw-w64-x86_64-cmake mingw-w64-x86_64-boost mingw-w64-x86_64-qt5 mingw-w64-x86_64-dlfcn mingw-w64-x86_64-xerces-c
+```
+One issue that the boost package has when building with GridLAB-D with HELICS on MSYS2 is all the boost libraries have a -mt in their names. In order for GridLAB-D to find these libraries the -mt must be removed from the libboost*-mt.dll files located in /mingw64/bin. Work is in progress to remove this modification in MSYS2 builds.
+
+After all this has been done /mingw64/bin must be added to the PATH environment variable.
+```bash
+$ export PATH=$PATH:/mingw65/bin
```
phlptp: should be ming64 and this should be in the system path already, so I am very unclear why this is necessary.  That is one of the system directories and is part of the default path.  Everything in that directory is available from the command line directly and is easily found.

docs/installation/windows.md
```diff
@@ -1,161 +1,263 @@
...
+- mingw-w64-x86_64-qt5 (only if you want to be able to run cmake-gui which this guide recommends.)
+- mingw-w64-x86_64-dlfcn
+- mingw-w64-x86_64-xerces-c
+- mingw-w64-x86_64-zeromq
+
+All packages can be installed by typing the following:
+```bash
+$ pacman -S --needed base-devel mingw-w64_x86_64-toolchain mingw-w64-x86_64-cmake mingw-w64-x86_64-boost mingw-w64-x86_64-qt5 mingw-w64-x86_64-dlfcn mingw-w64-x86_64-xerces-c
+```
+One issue that the boost package has when building with GridLAB-D with HELICS on MSYS2 is all the boost libraries have a -mt in their names. In order for GridLAB-D to find these libraries the -mt must be removed from the libboost*-mt.dll files located in /mingw64/bin. Work is in progress to remove this modification in MSYS2 builds.
```
phlptp: This is problematic.  the -mt on the boost libraries indicate that the libraries are mutlithread, which has meaning on windows systems.  It should be the default, but in the boost nomenclature if it doesn't have this it will cause some libraries to not use it and fail.    It will also prevent proper updates from occurring in pacman.    In the short term if this is needed a far better solution would be to add symlinks to that  directory so the original file are not modified.  And if this is necessary we should provide a script that adds these symlinks.  
We should not be modifying the name of system files!!

docs/installation/windows.md
```diff
@@ -1,161 +1,263 @@
...
+```bash
+$ git checkout -b develop origin/develop
+```
+
+### Compiling HELICS From Source ###
+Change directories to HELICS_ROOT_DIR. Create a directory called helics-build. This can be accomplished by using the mkdir command. cd into this directory. Now type the following:
+```bash
+$ cmake-gui ../
+```
+If this failes that is because mingw-w64-x86_64-qt5 was not installed. If you did install it the cmake gui window should pop up. click the Advanced check box next to the search bar. Then click Configure. A window will pop up asking you to specify the generator for this project. Select MSYS Makesfiles from the dropdown menu. Then click the Specify native compilers check box and click next. Enter C:/msys64/mingw64/bin/gcc.exe for the C compiler and C:/msys64/mingw64/bin/g++.exe for the C++ compiler and click finish. Once the Configure process finished several variables will show up highlighted in red. Since this is the first time setup the Boost and ZeroMQ library. Below are the following cmake variables that need to be verified.
```
phlptp: If we want this to be as easy as possible there should be a script provided that calls cmake directly with the appropriate variables.

docs/installation/windows.md
afisher1: you are absolutely right I'm fixing configure.ac in gld to do something different in windows for searching for the boost libraries.

afisher1:
That's what I was thinking. The only part I think is appropriate is the HELICS compilation instructions.

docs/installation/windows.md
afisher1: it is only if people started up the MSYS2 MinGW 64-bit terminal. If they just start the MSYS2 MSYS terminal it is not and you have to add it. This is just me trying to cover bases.

afisher1:
@phlptp I've made some revised changes to the document. please take a look. I've removed all the GridLAB-D stuff from the document and made corrections and clarifications addressing some of you comments.

docs/installation/windows.md
```diff
@@ -1,161 +1,227 @@
...
+- mingw-w64-x86_64-cmake
+- mingw-w64-x86_64-boost
+- mingw-w64-x86_64-qt5 (only if you want to be able to run cmake-gui which this guide recommends.)
+- mingw-w64-x86_64-zeromq
+
+All packages can be installed by typing the following:
+```bash
+$ pacman -S --needed base-devel mingw-w64_x86_64-toolchain mingw-w64-x86_64-cmake mingw-w64-x86_64-boost mingw-w64-x86_64-qt5
+```
+After all this has been done /mingw64/bin must be in the PATH environment variable. If it isn't then it must be added.
```
phlptp: Should add a note about the mingw-64 shell vs the msys shell, and if using the mingw-64 shell this isn't needed.

docs/installation/windows.md
```diff
@@ -1,161 +1,227 @@
...
+
+### Compiling HELICS From Source ###
+Change directories to HELICS_ROOT_DIR. Create a directory called helics-build. This can be accomplished by using the mkdir command. cd into this directory. Now type the following:
+```bash
+$ cmake-gui ../
+```
+If this failes that is because mingw-w64-x86_64-qt5 was not installed. If you did install it the cmake gui window should pop up. click the Advanced check box next to the search bar. Then click Configure. A window will pop up asking you to specify the generator for this project. Select MSYS Makesfiles from the dropdown menu. Then click the Specify native compilers check box and click next. Enter C:/msys64/mingw64/bin/gcc.exe for the C compiler and C:/msys64/mingw64/bin/g++.exe for the C++ compiler and click finish. Once the Configure process finished several variables will show up highlighted in red. Since this is the first time setup the Boost and ZeroMQ library. Below are the following cmake variables that need to be verified.
+
+* BUILD_CXX_SHARED_LIB should be checked as GridLAB-D dynamically links with the shared c++ library of HELICS
+* Boost_CHRONO_LIBRARY_DEBUG/RELEASE C:/msys64/mingw64/bin/libboost_chrono.dll
```
phlptp: on my system all the boost libraries have the -mt extension,  It will also depend on whether you are using static libraries or dynamic ones.   So there maybe should be a note that the exact boost libraries will depend on the other options and the particular system and file location.  What matters is that they are found

docs/installation/windows.md
afisher1: While it would make it easier on the developer I don't think the burden should be on us to provide a script to go out and find if someone installed the prerequisites in an nonstandard location.

docs/installation/windows.md
afisher1: I've edited the file providing a note about the mingw-64 shell.

docs/installation/windows.md
afisher1: I edited the file to reflect the standard MSYS2 boost package. unless they are compiling boost themselve which isn't necessary in the MSYS2 shell as there is a boost package that is new enough to build HELICS.

docs/installation/windows.md
phlptp: I think this comment was related to the changing of boost-mt files to not have that extension,  But I think that was resolved so the comment isn't relevant any more.

phlptp:
I will merge this once it builds

Issue #460, opened 2018-09-12
clang sanitizer build is an allowed failure


phlptp:
@nightlark  can we specify that the clang-thread-sanitizer build is an allowed failure.  Right now the log length is causing a failure.  It is and will be very useful but it shouldn't be required on the cron jobs.

PR #461, opened 2018-09-12
Allow clang thread sanitizer cron build to fail

nightlark:
* Issue #460 allowing clang thread sanitizer cron build as a failure
* Travis doesn't support wild cards for allowable failures, entry must be an exact match


nightlark:
* Addresses issue #460 allowing clang thread sanitizer cron build as a failure
* Travis doesn't support wild cards for allowable failures, entry must be an exact match

Issue #458, opened 2018-09-12
Issue template suggested update


bpalmintier:
FEATURE REQUEST

The current issue template starts with 
```
* **I'm submitting a ...**
  - [ ] bug
  - [ ] feature request
  - [ ] support question
```
Which github treats as a todo list and expects all of the check boxes to be complete in its progress bar. Would suggest simply  having the list and adding a comment to delete all but the desired issue type.

* **I'm submitting a ...**
  - [ ] bug
  - [ ] feature request
  - [ ] support question

### Bug

* **What is the current behavior?**

* **What is the expected behavior? What is the motivation / use case for changing the behavior?**

* **What are the steps to reproduce this bug? Please provide a minimal working example of the bug if possible.**

* **Please tell us about your environment:**

  - HELICS version: 

```
$ helics_app --version
```

  - Operating System: [all | Windows | Mac | Linux]
  - Language Extension: [all | Python | MATLAB | Java]

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, gitter, etc)

PR #467, opened 2018-09-22
Update issue templates

phlptp:
Split out the issue templates into 3 templates that can used instead of the single issue template.


phlptp:
Using the github templates to create bug/issue reports instead of the single template we had been using.  I copied most of the information over.  The idea is to merge this to master, than copy it back to develop and possibly tweak it a little further and remove the current template.  This is to address #458.

phlptp:
@kdheepak unless you see some major issue, just go ahead and merge this when you get a chance.  then we should create a PR from master into develop, so these changes get back into the develop branch, and we can make any further edits there.

PR #473, opened 2018-10-01
Fix docs

kdheepak:


docs/installation/language.md
```diff
@@ -27,7 +27,7 @@ IPython 6.2.1 -- An enhanced Interactive Python. Type '?' for help.
 In [1]: import helics
 
 In [2]: helics.helicsGetVersion()
-Out[2]: '1.0.2 (04-28-18)'
+Out[2]: '1.x.x (XX-XX-XX)'
```
phlptp: should probably have an x as the first number too since that will be updated soon

docs/installation/language.md
```diff
@@ -109,14 +109,14 @@ cmake -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX="C:\local\helics-develop
...
 ```
 >>mex -setup
 ```
-and follow the prompted instructions.  
+and follow the prompted instructions.
 
-Reconstructing the files requires a specific version of SWIG with MATLAB support.  IF swig is not used then adding `-DDISABLE_SWIG=ON` to the cmake command will bypass swig and use the included interface files directly.  If any modifications to the C library were made then swig must be used to regenerate the files appropriately, otherwise the existing files can be used.  The CMAKE scripts will detect if the swig is the appropriate version and act accordingly.    
+Reconstructing the files requires a specific version of SWIG with MATLAB support.  IF swig is not used then adding `-DDISABLE_SWIG=ON` to the cmake command will bypass swig and use the included interface files directly.  If any modifications to the C library were made then swig must be used to regenerate the files appropriately, otherwise the existing files can be used.  The CMAKE scripts will detect if the swig is the appropriate version and act accordingly.
```
phlptp: 'IF' should be 'If'

phlptp:
I will merge once the tests pass

PR #474, opened 2018-10-05
Fix codecov and appveyor badges to use master branch

nightlark:

# Description

Fixes display of Appveyor build status and code coverage badges to show the status of the master branch instead of develop.

# Proposed changes

Changes URLs for badge images and links for Appveyor and codecov from develop to master.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,11 +28,14 @@
   </tr>
   <tr>
   <td>Gitter</td>
-  <td colspan="2" align="center">[![Gitter chat](https://badges.gitter.im/GMLC-TDC/HELICS-src.png)](https://gitter.im/GMLC-TDC/HELICS-src)</td>
+  <td colspan="2" align="center">
+  <a href="https://gitter.im/GMLC-TDC/HELICS-src"><img src="https://badges.gitter.im/GMLC-TDC/HELICS-src.png" alt="Gitter Chat" /></a>
+  </td>
   </tr>
   <tr>
   <td>Documentation</td>
-  <td colspan="2" align="center"> [![Documentation](https://img.shields.io/badge/docs-ready-blue.svg)](http://gmlc-tdc.github.io/HELICS-src)</td>
+  <td colspan="2" align="center">
+  <a href="http://gmlc-tdc.github.io/HELICS-src"><img src="https://img.shields.io/badge/docs-ready-blue.svg" alt="Documentation" /></a>
   </tr>
 </table>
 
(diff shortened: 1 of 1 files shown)


phlptp:
There is no issue with the changes, the question is really should we have the develop or master branch showing the badges?

phlptp:
I implemented Dheepak's suggestion of a table rather than choosing between the two branches [table](https://github.com/GMLC-TDC/HELICS-src/tree/badge-fix). Of course it doesn't work correctly yet

phlptp:
Should be working now,  had to go with full html table instead of mixing markdown and html, which worked just fine in my previewer, but not on Github.

nightlark:
A table in markdown didn’t work?

phlptp:
As far as I can tell the markdown tables down support column merging and if you want to do anything remotely complicated you need to use html

nightlark:
Okay, I see what you mean. I guess the only way to do only markdown would be duplicating badges or having badges that aren't part of the tables.

Should the Travis badges link to the page for the build history of the corresponding branch?

phlptp:
I played with it with just markdown and I couldn't get it to look right.   
Also on the travis linking, I thought of doing that but that is probably one of the links i use most frequently and I like it to just go to the summary list of all the recent builds.    We can debate it, or may have the develop branch go to the develop history, or something like that, but the master should probably just go the list of builds.  It isn't like we allow the master to not pass very often.

kdheepak:
I'd prefer the table to be transposed, but other than that it looks good!

phlptp:
I don't know, I kind of like the vertical orientation,  It definitely shows up nicer on a small screen like a phone, and if we were to add some columns or rows it has space to grow easily.

phlptp:
While we are tweaking the readme, is there anything else we want to change on it?

bpalmintier:
I like the current vertical orientation, but am flexible.

For the readme, other suggestions:

* We use the idea of examples twice, ones in Getting started (external repo) and also in Documentation. Is the documentation referring to old embedded repo folders? If so let's delete, if not perhaps we can find a new name?
* Add mention of the HELICS book documentation effort?
* Need to refresh the tool list. Guessing there are some additions, and I think GridLAB-D may now be part of the main release.
* For the tools, might we need some form of version-based table? at least for 1.3.1 vs 2.0?

bpalmintier:
Should we spin out a different README issue?
Another item for it: Additional publications.

phlptp:
How about I merge this and add an issue for further corrections and modifications?

PR #491, opened 2018-10-19
Version update

phlptp:
# Description

Version update,  rename one of the recently added functions, update to contributors, and changelog


nightlark:
Should also update the README to remove the part about GridLab-D having support in the feature/1024 branch.

nightlark:
I think it should link to https://github.com/gridlab-d/gridlab-d/tree/develop instead, since it isn't in their master branch yet.

PR #501, opened 2018-11-01
Python swig interface updates

kdheepak:
- [X] I have reviewed the [Contributing Guidelines](../CONTRIBUTING.md)
- [X] I agree to license my contributions under the same license as in this repository
- [X] I have verified that the tests pass

# Description

Adds support for Python3 swig interface for HELICS 2.0

### DIFF
--- a/tests/python_helics/test_message_federate.py
+++ b/tests/python_helics/test_message_federate.py
@@ -51,6 +51,7 @@ def mFed():
     while (h.helicsBrokerIsConnected(broker)):
         time.sleep(1)
 
+    h.helicsFederateInfoFree(fedinfo)
     h.helicsFederateFree(mFed)
     h.helicsCloseLibrary()
 
--- a/tests/python_helics/test_message_federate.py
+++ b/tests/python_helics/test_message_federate.py
@@ -38,7 +38,8 @@ def mFed():
     # Set one second message interval #
     h.helicsFederateInfoSetTimeProperty(fedinfo, h.helics_time_property_time_delta, deltat)
 
-    # h.helicsFederateInfoSetLoggingLevel(fedinfo, 1)
+    h.helicsFederateInfoSetIntegerProperty(fedinfo, h.helics_int_property_log_level, 1)
+
 
     mFed = h.helicsCreateMessageFederate("TestA Federate", fedinfo)
 
--- a/tests/python_helics/test_message_federate.py
+++ b/tests/python_helics/test_message_federate.py
@@ -19,10 +19,10 @@ def mFed():
         pass
 
     # Create Federate Info object that describes the federate properties #
-    fedinfo = h.helicsFederateInfoCreate()
+    fedinfo = h.helicsCreateFederateInfo()
 
     # Set Federate name #
-    h.helicsFederateInfoSetFederateName(fedinfo, "TestA Federate")
+    h.helicsFederateInfoSetCoreName(fedinfo, "TestA Federate")
 
     # Set core type from string #
     h.helicsFederateInfoSetCoreTypeFromString(fedinfo, "zmq")
@@ -36,19 +36,17 @@ def mFed():
     # setTimedelta routine is a multiplier for the default timedelta.
 
     # Set one second message interval #
-    h.helicsFederateInfoSetTimeDelta(fedinfo, deltat)
+    h.helicsFederateInfoSetTimeProperty(fedinfo, h.helics_time_property_time_delta, deltat)
 
-    h.helicsFederateInfoSetLoggingLevel(fedinfo, 1)
+    # h.helicsFederateInfoSetLoggingLevel(fedinfo, 1)
 
-    mFed = h.helicsCreateMessageFederate(fedinfo)
+    mFed = h.helicsCreateMessageFederate("Federate 1", fedinfo)
 
     yield mFed
 
-    status = h.helicsFederateFinalize(mFed)
-    assert status == h.helics_ok
-    status, state = h.helicsFederateGetState(mFed)
+    h.helicsFederateFinalize(mFed)
+    state = h.helicsFederateGetState(mFed)
     assert state == 3
-    assert status == h.helics_ok
     while (h.helicsBrokerIsConnected(broker)):
         time.sleep(1)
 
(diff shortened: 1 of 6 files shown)


phlptp:
I would delete the `typemap(XXX) (char *outputString, int maxlen)`  typemaps they are not used anymore, and are part of the issue with the namedpoint.

tests/python_helics/test_message_federate.py
```diff
@@ -19,10 +19,10 @@ def mFed():
         pass
 
     # Create Federate Info object that describes the federate properties #
-    fedinfo = h.helicsFederateInfoCreate()
+    fedinfo = h.helicsCreateFederateInfo()
 
     # Set Federate name #
-    h.helicsFederateInfoSetFederateName(fedinfo, "TestA Federate")
+    h.helicsFederateInfoSetCoreName(fedinfo, "CoreA Federate")
```
phlptp: these are very different calls,  in one you are setting the name of the core to connect to and in the previous version you were setting the desired name of the federate.

kdheepak: Yes. The name of the federate now does not seem to be set by a function anymore. It is set when creating the federate. 

```
HELICS_EXPORT helics_federate helicsCreateValueFederate (const char *fedName, const helics_federate_info_t fi, helics_error *err);
```

This takes the federate name.

kdheepak:
> I would delete the `typemap(XXX) (char *outputString, int maxlen)` typemaps they are not used anymore, and are part of the issue with the namedpoint.

I did see that but left it in there because there was the following:

```
src/helics/shared_api_library/MessageFederateExport.cpp
449:bool checkOutArgString(char *outputString, int maxlen, helics_error *err)
452:    if ((outputString == nullptr) || (maxlen <= 0))

src/helics/shared_api_library/internal/api_objects.h
176:bool checkOutArgString (char *outputString, int maxlen, helics_error *err);
```

I see now that these are not exported. So I'll remove them.

tests/python_helics/test_message_federate.py
kdheepak: See this line [here](https://github.com/GMLC-TDC/HELICS-src/pull/501/files#diff-78391ea64970da9b0bb6eb7148788be9R43).

kdheepak:
All Python tests pass

screen shot 2018-11-01 at 1 46 46 pm

kdheepak:
This should not be merged yet. I think this segfaults instead of throwing an error.

kdheepak:
This is ready for review / merging.

tests/python_helics/test_message_filter.py
```diff
@@ -23,42 +23,37 @@ def AddBroker(core_type="zmq", number_of_federates=1):
...
-    assert status == 0
+    h.helicsFederateInfoSetTimeProperty(fedinfo, h.helics_time_property_time_delta, deltat)
 
-    status = h.helicsFederateInfoSetLoggingLevel(fedinfo, 1)
-    assert status == 0
+    h.helicsFederateInfoSetIntegerProperty(fedinfo, h.helics_int_property_log_level, 1)
 
-    mFed = h.helicsCreateMessageFederate(fedinfo)
+
+    mFed = h.helicsCreateMessageFederate(name_prefix, fedinfo)
```
phlptp: You may want to free the federateInfo if you are done with it, otherwise it will be a small memory leak.

tests/python_helics/test_value_federate.py
```diff
@@ -108,58 +97,49 @@ def test_value_federate_runFederateTestNamedPoint(vFed):
...
+    h.helicsPublicationPublishNamedPoint(pubid, testValue1, testVal1)
 
     # double val;
-    status, value, val = h.helicsSubscriptionGetNamedPoint(subid)
-    assert status == 0
-    assert value == defaultValue
-    assert val == defVal
+    val = h.helicsInputGetNamedPoint(subid)
+    # assert value == defaultValue
+    assert val == ['start of a longer string in place of the shorter one and now this should be very long', 5.3]
```
phlptp: Should you use the named variable here, so you don't have to change the string in two places if you want a different string?  Or is there some reason this won't work

tests/python_helics/test_message_filter.py
kdheepak: I'm assuming it has to be freed only at the end of the program, and not immediately after creating the Federate. Is that correct?

phlptp: I don't believe the fedinfo maintains a pointer internal to the library like federates/cores/brokers so to have no leaks it would need to be freed at some point after you are done using it.    You can use it multiple times if you like.    it isn't that big of object and it is not likely to be created a whole bunch of times so it is a small issue if you forget but the test code should be good usage.  
So as long as you keep the object around you can free it whenever.

kdheepak: Okay, I added free fedinfo functions along with the other free functions.

tests/python_helics/test_value_federate.py
kdheepak: I made this change.

kdheepak:
Do not merge this, at the moment the tests don't pass locally.

kdheepak:
Okay, this is ready to be merged now.

tests/python_helics/test_message_federate.py
```diff
@@ -51,6 +51,7 @@ def mFed():
     while (h.helicsBrokerIsConnected(broker)):
         time.sleep(1)
 
+    h.helicsFederateInfoFree(fedinfo)
```
phlptp: you aren't double freeing here are you?    otherwise this seems a little inconsistent with the other tests?

kdheepak: Why is this a double free? I think this is identical to what we are doing in the other Python tests.

phlptp: It looks like in some places you call a separate free function ' FreeFederate(fFed, ffedinfo)'  that frees the fed info, but I wasn't sure how that interacted with this free?  It may not, I am just trying to clarify things.

kdheepak: Yes, this FreeFederate calls helicsFederateInfoFree, so it's all good. Only the `test_message_filter.py` using FreeFederate because it made it easier to create multiple federates in a single test this way.

phlptp:
Ok, then I think the only thing, is why the tests don't seem to be running in Travis?

PR #503, opened 2018-11-06
Python interface cmake build

nightlark:

# Description

Builds the Python interface without swig as a CMake library to avoid errors on systems that have read-only pip installations.

### DIFF
--- a/interfaces/python/CMakeLists.txt
+++ b/interfaces/python/CMakeLists.txt
@@ -82,81 +82,57 @@ if(ENABLE_SWIG AND SWIG_EXECUTABLE)
 else()
   message("Building without swig")
 
-  add_library(helicsPYTHON MODULE helics_wrap.c)
-  set_target_properties(helicsPYTHON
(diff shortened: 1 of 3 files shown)


nightlark:
I think all of our builds use the oldest supported cmake version (3.4.3), but only one of the builds uses the pre-generated swig headers.

phlptp:
Should we have a build with a newer cmake?

nightlark:
Actually, I was wrong -- the Windows build on Appveyor uses CMake 3.12, but it uses swig. I don't think the added time for an additional build on every commit is worth it (maybe for PRs to master), since most of the failures are from the generated files not getting updated; one of the builds detecting changes that result in different generated files would be useful.

phlptp:
Someday I would love to see an automated system that checks and regenerates the files for the non-swig builds and makes a PR with the updated files.

phlptp:
but you are right we probably don't need a separate build for that test.

PR #505, opened 2018-11-07
Add install include path to imported helics-static and helics-shared cmake targets

nightlark:

# Description
Adds the install include path as part of the interface for the helics-static and helics-shared CMake targets imported by projects using find_package in config mode to locate HELICS. Allows using only CMake targets without needing to use the HELICS_INCLUDE_DIR variable to add an include directory.

### DIFF
--- a/src/helics/CMakeLists.txt
+++ b/src/helics/CMakeLists.txt
@@ -23,6 +23,10 @@ add_library(
 target_link_libraries(helics-static PUBLIC helics_base)
 # add and alias library to match the find_package
 add_library(HELICS::helics-static ALIAS helics-static)
+target_include_directories(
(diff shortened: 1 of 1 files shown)


phlptp:
Is there anything else to be added along these lines or should I merge it?

nightlark:
Do you think a similar change should be made to the helics apps target?

phlptp:
The Helics_app-target should have a dependency on the static library already,  I would think they should propagate from what you have already.

nightlark:
Alright -- I think this is okay for merging then. If a use case comes up later with helics apps that has problems with the include directory propagating to the helics app target then a similar change can be made.

PR #507, opened 2018-11-08
Fix strip protocol

kdheepak:

# Description

Without this fix, `stripProtocol` returns `/127.0.0.1` instead of `127.0.0.1` when passed in `tcp://127.0.0.1`.


phlptp:
It looks like we are missing test cases for 
```
/** strip any protocol strings from the interface and return a new string
@example tcp://127.0.0.1 -> 127.0.0.1*/
std::string stripProtocol(const std::string &networkAddress);
/** strip any protocol strings from the interface and return a new string*/
void removeProtocol (std::string &networkAddress);

/** add a protocol url to the interface and return a new string*/
std::string addProtocol (const std::string &networkAddress, interface_type interfaceT);

/** add a protocol url to the interface modifying the string in place*/
void insertProtocol (std::string &networkAddress, interface_type interfaceT);
``` 
they should probably be added to 
networkInfoTests.cpp

kdheepak:
This is ready to be reviewed now.

Issue #510, opened 2018-11-09
Function naming for C interface


kdheepak:
This is the list of current names:

```
h.HELICS_CORE_TYPE_DEFAULT: <class 'int'>
h.HELICS_CORE_TYPE_INTERPROCESS: <class 'int'>
h.HELICS_CORE_TYPE_IPC: <class 'int'>
h.HELICS_CORE_TYPE_MPI: <class 'int'>
h.HELICS_CORE_TYPE_TCP: <class 'int'>
h.HELICS_CORE_TYPE_TCP_SS: <class 'int'>
h.HELICS_CORE_TYPE_TEST: <class 'int'>
h.HELICS_CORE_TYPE_UDP: <class 'int'>
h.HELICS_CORE_TYPE_ZMQ: <class 'int'>
h.HELICS_CORE_TYPE_ZMQ_TEST: <class 'int'>
h.HELICS_DATA_TYPE_ANY: <class 'int'>
h.HELICS_DATA_TYPE_BOOLEAN: <class 'int'>
h.HELICS_DATA_TYPE_CHAR: <class 'int'>
h.HELICS_DATA_TYPE_COMPLEX: <class 'int'>
h.HELICS_DATA_TYPE_DOUBLE: <class 'int'>
h.HELICS_DATA_TYPE_INT: <class 'int'>
h.HELICS_DATA_TYPE_NAMEDPOINT: <class 'int'>
h.HELICS_DATA_TYPE_RAW: <class 'int'>
h.HELICS_DATA_TYPE_STRING: <class 'int'>
h.HELICS_DATA_TYPE_TIME: <class 'int'>
h.HELICS_DATA_TYPE_VECTOR: <class 'int'>
h.cvar: <class 'swigvarlink'>
h.data_t: <class 'type'>
h.data_t_swigregister: <class 'builtin_function_or_method'>
h.force_iteration: <class 'int'>
h.helicsBrokerAddDestinationFilterToEndpoint: <class 'builtin_function_or_method'>
h.helicsBrokerAddSourceFilterToEndpoint: <class 'builtin_function_or_method'>
h.helicsBrokerClone: <class 'builtin_function_or_method'>
h.helicsBrokerDataLink: <class 'builtin_function_or_method'>
h.helicsBrokerDestroy: <class 'builtin_function_or_method'>
h.helicsBrokerDisconnect: <class 'function'>
h.helicsBrokerFree: <class 'function'>
[...]

PR #513, opened 2018-11-15
synchronize c and C++ api more,  make sure all symbols from the C api…

phlptp:
… have helics_ nomenclature,  remove _t from some types for consistency.   add helics_enums (for all libraries) and helics_definitions for (C++)

### Description

update symbol naming in the C and C++ api's and remove some definitions that may conflict

### Proposed changes

- add defs namespace in C++ with symbol definitions instead of #defines
- add helics_enums with the baseline definitions that get used in C and C++
- update all files that used any of these symbols
- change C types to have helics_ in front,  mainly message_t to helics_message and helics_time_t to helics_time
- make the symbols the same to the extent possible between C and C++


phlptp:
Link to #510

src/helics/MessageFederates.hpp
```diff
@@ -0,0 +1 @@
+
```
afisher1: This file appears to be empty. What is the purpose of this file?

src/helics/ValueFederates.hpp
```diff
@@ -0,0 +1 @@
+
```
afisher1: Another empty file?

src/helics/core/helics_definitions.hpp
```diff
@@ -0,0 +1,94 @@
+/*
```
afisher1: Shouldn't src/helics/helics_enums.h be included in this file?

phlptp: yes it should,  must have just included that in every file that also used this.

src/helics/ValueFederates.hpp
phlptp: see above, and ApplicationApi.hpp will include all the files related to the Application API in one shot.

src/helics/MessageFederates.hpp
phlptp: I am not quite done yet, but this will be a convenience header to include all the files related to MessageFederates in the application API

nightlark:
Okay, so it is a matlab thing.

src/helics/core/helics_definitions.hpp
afisher1: Ok. Seeing as how this is a core header it won't be used at the application api layer it shouldn't affect the interface but I do suggest this be added before we forget.

phlptp: I added it, and I think this should be used by the application API, it will be included with the files for the application_api,  I doubt it would be included in isolation but it definitely needed that header.

afisher1: Well I meant normally external tools will most likely just use helics_enums.h. I normally wouldn't recommend them having to include headers form the core.

phlptp:
I think the documentation is one thing we need to put a little effort into before the 2.0 final comes out

PR #514, opened 2018-11-16
Add baseline elements for info field.

phlptp:

### Proposed changes

-
-
-

Adding the pull request for easier discussion and comments


src/helics/application_api/ValueFederateManager.cpp
```diff
@@ -437,6 +437,40 @@ bool ValueFederateManager::getPublicationOption (const Publication &pub, int32_t
     return coreObject->getHandleOption (pub.handle, option);
 }
 
+std::string ValueFederateManager::getInputInfo(const Input &imp) const {
```
phlptp: We should be able to deal with this directly in the Value Federate since the info is contained in the core object no reason for the extra layer of calls here.

src/helics/core/BasicHandleInfo.hpp
```diff
@@ -58,6 +58,7 @@ class BasicHandleInfo
     const std::string key;  //!< the name of the handle
     const std::string type;  //!< the type of data used by the handle
     const std::string units;  //!< the units associated with the handle
+    const std::string interface_info;
```
phlptp: this should not be const,  the others are because they cannot change once constructed,  info should be able to like `used` and `flags`

src/helics/application_api/ValueFederate.cpp
```diff
@@ -138,37 +146,42 @@ void ValueFederate::registerValueInterfacesJson (const std::string &configString
...
+    if (doc.isMember ("publications")) // TODO: add info field loading
     {
         auto pubs = doc["publications"];
         for (const auto &pub : pubs)
         {
             auto key = getKey (pub);
 
-            Publication *pubAct = &vfManager->getPublication (key);
-            if (pubAct->isValid ())
+            auto &pubAct = vfManager->getPublication (key);
```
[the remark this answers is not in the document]
eberleim: My bad, I was going for uniformity, and hadn't had a chance to do any testing. Is this problem not shared by the `vfManager->getSubscription(key)` on line 190? It was simply having two sections of code that "appear" to be equivalent in purpose be implemented two different ways.

src/helics/core/CommonCore.cpp
```diff
@@ -3744,4 +3744,21 @@ void CommonCore::processMessageFilter (ActionMessage &cmd)
...
+    {
+        auto handleInfo = getHandleInfo (handle);
+        if (handleInfo != nullptr)
+        {
+            return handleInfo->interface_info;
+        }
+        return emptyStr;
+    }
+
+    void CommonCore::setInterfaceInfo(helics::interface_handle handle, std::string info) {
```
phlptp: And this would be the trouble spot?  
Here you need to use a modify function on handles directly like,  may need some checks as well to make sure handle is valid
```
handles.modify ([&](auto &hdls) { hdls.getHandleInfo (handle.baseValue ())->interface_info = info; });
```

phlptp: the handles is protected by a ordered_guarded object which only allows modifications through the modify function so you need to write a little lambda function that does the modifications

src/helics/core/HandleManager.cpp
```diff
@@ -36,6 +36,21 @@ BasicHandleInfo &HandleManager::addHandle (global_federate_id_t fed_id,
     return handles.back ();
 }
 
+//    BasicHandleInfo &HandleManager::addHandle (global_federate_id_t fed_id,
```
phlptp: I would prefer not to have this function

src/helics/core/HandleManager.hpp
```diff
@@ -82,6 +82,8 @@ class HandleManager
     const BasicHandleInfo *getInput(const std::string &name) const;
     federate_id_t getLocalFedID (interface_handle id_) const;
 
+    BasicHandleInfo *getinterfaceInfo();
```
phlptp: This one doesn't seem necessary either

src/helics/application_api/ValueFederate.cpp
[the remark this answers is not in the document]
phlptp: It isn't because subAct is never assigned to.  I suppose we should make all those consistent since you can either use pointers or just make a new reference when calling the register function.    Just can't reassign a reference.   I should probably make that a little safer since that might easy to do.    Will have to think about that one a little more.

src/helics/application_api/MessageFederate.cpp
```diff
@@ -119,6 +119,7 @@ void MessageFederate::registerMessageInterfacesJson (const std::string &jsonStri
         {
             auto eptName = getKey (ept);
             auto type = jsonGetOrDefault (ept, "type", std::string ());
+            auto info = jsonGetOrDefault (ept, "info", std::string ());
```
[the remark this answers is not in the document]
phlptp: the case that would be problematic for this is if the info was defined in a JSON file as a JSON formatted string.  In that case I am not entirely certain the code would function properly.    I might be able to make it work as such though, but we do need to take that into account that info might be a JSON structure in a JSON file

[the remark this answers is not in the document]
eberleim: I implemented the changes (once I figured out how to do the modify, the rest of it was fairly trivial and a lot of the garbage I created trying to dig through it could be cleaned up), and you're right, a true JSON string breaks it because of the quotes, however I was able to put an escaped JSON string in without any trouble. I will push up those changes to the branch, and we can evaluate how we want to handle those, without them necessarily interfering with the beta window any further.

phlptp:
Remaining items to do before we can merge 
[ ] add some test cases
[ ] add the calls on the Input, Publication, Filter, and Endpoint objects
[ ] add the corresponding C library functions
[ ] add test cases for the C API calls

src/helics/application_api/Federate.cpp
```diff
@@ -1155,4 +1155,12 @@ void Federate::setFilterOption (const Filter &filt, int32_t option, bool option_
...
 }
 
+    void Federate::setInfo(interface_handle handle, const std::string& info) {
+        coreObject->setInterfaceInfo(handle, info);
+    }
+
+std::string const &Federate::getInfo(interface_handle handle) {
+    return coreObject->getInterfaceInfo(handle);
+}
+
```
phlptp: I am wondering whether this is best or if a series of overloads might be better.  Using the interface_handle directly  does simplify the code so I am not sure but it might be inconsistent with the rest of the API.  Any thoughts?

src/helics/application_api/ValueFederateManager.hpp
```diff
@@ -161,10 +161,6 @@ class ValueFederateManager
     /** get an option values for a publication*/
     bool getPublicationOption (const Publication &pub, int32_t option) const;
 
-    /** register a callback function to call when any subscribed value is updated
-    @details there can only be one generic callback
-    @param[in] callback the function to call
-    */
```
phlptp: why was this deleted?

tests/helics/application_api/ValueFederateAdditionalTests.cpp
```diff
@@ -414,6 +414,59 @@ BOOST_DATA_TEST_CASE (test_async_calls, bdata::make (core_types), core_type)
...
+    sub1.setInfo(std::string("sub_test1"));
+    sub2.setInfo(std::string("sub_test2"));
+    sub3.setInfo(std::string("sub_test3"));
+
+    vFed1->enterExecutingMode();
+
+    auto info1 = vFed1->getInfo(pubid1.getHandle());
+    auto info2 = vFed1->getInfo(sub1.getHandle());
+    auto info3 = vFed1->getInfo(sub2.getHandle());
+    auto info4 = vFed1->getInfo(sub3.getHandle());
```
phlptp: This should probably call the getInfo directly on the object vs the fed call

phlptp:
Added a few items to the checklist, and added an example for inputs in the C API.  You should be able to replicate that for the publications/filters/endpoints in the C api

src/helics/application_api/ValueFederateManager.hpp
eberleim: It got caught up in an overzealous delete of some functions which had been placed directly above it. I've restored it on local and committed it. I will push that up with the other fixes when they're ready.

tests/helics/application_api/ValueFederateAdditionalTests.cpp
eberleim: I'll add direct requests to the subs themselves, but I think I'll keep the versions that request against the federate, since it gives us better coverage of possible use cases.

src/helics/application_api/Federate.cpp
eberleim: Would it be more consistent to have a template method in the Federate that takes one of the object types and calls getHandle() internally then passes it to the core? That occurred to me as an option last night.

phlptp: I have thought about it too, but what I am trying to avoid is having to include the definitions of Inputs, Publications, Endpoints, and Filters in the header file that defines the Federate.    There may be a way to have the template definition in a separate file but that gets a little complicated.

phlptp: There are a couple other functions that are identical between all the interfaces.  I thought about making all the interfaces share a base class but that has similar issues to the templates

tests/helics/application_api/test_files/example_combo_fed.json
```diff
@@ -45,8 +45,9 @@
 },
 {
 "key":"pub2", // the name of the publication
- "type":"double" // the type assocaited with a publication (optional)
+ "type":"double", // the type assocaited with a publication (optional)
  //no global:true implies this will have the federate name prepended like comboFed/pub2
+ "info":{"field1":45,"field2":99}
```
eberleim: While I also prefer the "pass in real JSON" on principle, I'm concerned about the issues this might cause for the time being, do we want to try to support this functionality immediately? My impression was we were going to attempt to enforce string-format for now.
something like:
`"info":"{\"field1\":45,\"field2\":99}"`

tests/helics/application_api/ValueFederateAdditionalTests.cpp
phlptp: that sounds fine we should test both ways

tests/helics/application_api/test_files/example_combo_fed.json
phlptp: There seemed to be a way of doing it with minimal code in the call to getValueOrDefault with json so I just added it.    What issues are you concerned with?

eberleim: As long as it doesn't cause encoding issues, I suppose it's a non-issue. My only real concern was that we may see a lot of "I gave it JSON but got an escaped string back and had to add a JSON decoder on my end to read it" which is the behavior I would expect, but might not be intuitive, or something of that nature.

phlptp: I think we will just have to make sure we document what it does and doesn't do properly.  If it was a json structure to begin with it should come out as a json interpretable string.    In C++ you can probably just stick whatever data you wanted in there since std::string isn't limited to text characters.  That isn't going to work on C interfaces though

phlptp:
@eberleim I think way we did this you need to approve the PR.    I am rerunning the travis results, there might be an issue with the cache from the added functions, which is why the python stuff isn't working

eberleim:
I'm doing the C test cases right now, they're nearly complete. 
I'll see if github lets me reassign it as a "I want to merge" item.

eberleim:
@phlptp I've pushed up the C API Tests, so I think we now have all the pieces we needed.

phlptp:
go ahead and approve it,   The push failure is due to the out of date api files,  I will be finalizing all of those in the final batch once this and the others are merged.   So I think you can go ahead an approve and merge it when you are ready

PR #522, opened 2018-11-25
Close routes

phlptp:

### Description

Adds a removeRoute function to the commonCore and coreBroker API's and calls the functions when a route is disconnected.  This should make the sporadic TCP errors a little less likely since the client connections should in most cases be closed first.


src/helics/common/AsioServiceManager.cpp
```diff
@@ -147,32 +144,31 @@ AsioServiceManager::LoopHandle AsioServiceManager::runServiceLoop (const std::st
...
-    ++runCounter; //atomic
+    ++runCounter;  // atomic
 
     bool exp = false;
     if (running.compare_exchange_strong (exp, true))
     {
         auto ptr = shared_from_this ();
-        std::packaged_task<void()> serviceTask ([ptr=std::move(ptr)]() { serviceProcessingLoop (ptr); });
-     //   std::cout << "run Service loop " << runCounter << "\n";
+        std::packaged_task<void()> serviceTask ([ptr = std::move (ptr)]() { serviceProcessingLoop (ptr); });
```
kdheepak: This is a closure, correct? Why wouldn't you `std::move` the `ptr` in the closure body? My understanding is that you rvalue cast in the arguments for the closure, right? I thought it would have been written as below.

```
std::packaged_task<void()> serviceTask ([ptr = std::move (ptr)]() { serviceProcessingLoop (std::move(ptr)); });
```

It's probably just an extra copy, so it's fine. I'm just curious.

src/helics/core/ActionMessage.cpp
```diff
@@ -189,7 +189,7 @@ std::string ActionMessage::packetize () const
     // don't forget to flush the stream to finish writing into the buffer
     s.flush ();
     // now generate a length header
-    auto sz = static_cast<int32_t> (data.size ());
+    auto sz = static_cast<uint32_t> (data.size ());
```
kdheepak: Why are we using unsigned here? From Google's cpp style guide: 

> You should not use the unsigned integer types such as uint32_t, unless there is a valid reason such as representing a bit pattern rather than a number, or you need defined overflow modulo 2^N. In particular, do not use unsigned types to say a number will never be negative. Instead, use assertions for this. [1]

[1] - https://google.github.io/styleguide/cppguide.html#Integer_Types

phlptp: the next lines are bit shifts and & operations.    clang-tidy flag it as a code issue since you are doing bit operations and packing a raw data structure.

phlptp: and that should be done on unsigned values

kdheepak: Okay great!

src/helics/common/AsioServiceManager.cpp
kdheepak: There's a bunch of other places in the code that I see a similar change required.

src/helics/common/AsioServiceManager.cpp
```diff
@@ -182,52 +178,51 @@ AsioServiceManager::LoopHandle AsioServiceManager::startServiceLoop ()
             if (running.compare_exchange_strong (exp, true))
             {
                 auto ptr = shared_from_this ();
-                std::packaged_task<void()> serviceTask ([ptr=std::move(ptr)]() { serviceProcessingLoop (ptr); });
+                std::packaged_task<void()> serviceTask (
+                  [ptr = std::move (ptr)]() { serviceProcessingLoop (ptr); });
```
kdheepak: Same change here?

PR #525, opened 2018-12-07
Custom serialization

phlptp:
Reduce headers for ActionMessage and use a custom serialization instead of cereal. 
This should improve performance somewhat, though most cases it probably won't be noticeable.    But it also improves compile time a little.  

### Description

Use a custom serialization of the ActionMessage, which reduces the size a little and should improve performance for some cases.

### DIFF
--- a/src/helics/common/timeRepresentation.hpp
+++ b/src/helics/common/timeRepresentation.hpp
@@ -21,9 +21,9 @@ All rights reserved. See LICENSE file and DISCLAIMER for more details.
 #include <chrono>
 #include <cmath>
 #include <cstdint>
+#include <iosfwd>
 #include <limits>
 #include <type_traits>
-#include <iosfwd>
 #ifdef __cpp_lib_chrono
 #define CHRONO_CONSTEXPR constexpr
 #else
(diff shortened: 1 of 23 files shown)


Jacobhansens:
I tested this branch with the use case I have been working on and I get the following error message from the broker:

```
root (1)::Federation Summary> 
	8 federates
	8 brokers/cores
	1375 publications
	0 inputs
	0 endpoints
	0 filters
<<<<<<<<<
commWarning||1778-5xQBZ-8tWlu-OFsJc-ksFVo:unknown route and no broker, dropping message
commWarning||1778-5xQBZ-8tWlu-OFsJc-ksFVo:unknown route and no broker, dropping message
```

Any idea what could be causing this?

Jacobhansens:
I just looked at the results I get and something is definitely being dropped in my simulation with these changes. I am checking to see if I made a mistake on my end.

I will also pull the latest commit and see if the more detailed message will help me narrow it down

phlptp:
I thought about adding a test to that effect but I am not sure that would actually test much since the test would be most about the ability to pretend we know what a big endian system would do.    I have heard it is possible to make a mac compile with power pc emulation so maybe we should try that at some point and have it link with an x86_64 system.

PR #526, opened 2018-12-07
Set property function change

phlptp:

### Description

use an overload set for setProperty on the federate and FederateInfo structures instead of separate functions.    And allow in most usable instances the properties to interchange across the types.  The separate functions still need to remain in place for the C interface, though at some point the language interfaces may want to adopt the overload.


phlptp:
appveyor is working fine.  The codacy check is showing as failed due it not being able to do the analysis.    @afisher1  this change will require some updates to the GridLab-d code for setting properties, just so you are aware.

afisher1:
I'm aware of that @phlptp thanks!

PR #528, opened 2018-12-08
Global values

phlptp:

### Description

add the ability to set a global string that can be queried through the query interface

### Proposed changes

add an ability to set a global field programatically or through a configuration file.  This field can be used with configuration information or other information.  

-
-
-


nightlark:
Is the loop around todo comment related to globales?

phlptp:
yes,  I am not entirely sure how to do that with toml files yet.  It wasn't working like i expected it to so I need to investigate that further, and it wasn't critical so I put it off.  It requires being able to loop over all the elements in the globals section and extract the name and value.  I don't think that will be the preferred way of specifying them so I don't think not having it yet is that critical.

Issue #529, opened 2018-12-09
Building HELICS without tests still requires Boost unit test framework


nightlark:
**Describe the bug**
The CMake script for HELICS wants the Boost unit test framework libraries to be present, even when building tests is disabled.

**What is the expected behavior?**
The unit test framework libraries for Boost should not need to be found if tests are not getting built.

PR #530, opened 2018-12-10
Some updates to CMAKE to allow disabling individual core types and re…

phlptp:

### Description

This PR addresses #529 by changing the list of boost required libraries to not include unit_testing_framework if BUILD_TESTING is not enabled.

It also adds hidden option to selectively disable particular core types  by DISABLE_<core_type>_CORE option in CMAKE


nightlark:
Should the option for the MPI core be renamed to DISABLE_MPI_CORE (default OFF) instead of ENABLE_MPI, to match the options for the other cores?

phlptp:
I went back and forth on that.  I left it as is since the ENABLE_MPI is a positive action and a visible option.  DISABLING the various core type is the same notion of action against the default and they are HIDDEN options that can also be triggered by various build conditions and particular circumstances.  So I definitely don't want them showing in the cmake-gui normally which is different than the ENABLE_MPI and ZMQ options.

PR #533, opened 2018-12-12
Cleanup unused headers and remove unneeded compiled Boost libraries

nightlark:

### Description

Removes some unused headers and compiled Boost libraries that aren't used, reduces the number of Boost libraries that need to be compiled while building for CI or installed with apt.

### Proposed changes

- Remove some headers that are included but not used from several source files
- Remove unused compiled Boost libraries; chrono isn't used (std::chrono is used instead), timer isn't needed (for C++11 compilers, std::chrono is used internally), and compiled parts of date_time aren't used
- Remove unused compiled Boost libraries from libraries to build on Travis and in build instruction documents

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -275,9 +275,6 @@ set(
   program_options
   filesystem
   system
-  date_time
-  timer
-  chrono
 )
 if (BUILD_TESTING)
 	list(APPEND BOOST_REQUIRED_LIBRARIES unit_test_framework)
--- a/config/cmake/addBoost.cmake
+++ b/config/cmake/addBoost.cmake
@@ -1,5 +1,5 @@
 ##############################################################################
-#Copyright © 2017-2018,
+#Copyright Â© 2017-2018,
 #Battelle Memorial Institute; Lawrence Livermore National Security, LLC; Alliance for Sustainable Energy, LLC
 #All rights reserved. See LICENSE file and DISCLAIMER for more details.
 ##############################################################################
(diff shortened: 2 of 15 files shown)


phlptp:
So if I understand all this correctly 
the `BOOST_DATE_TIME_NO_LIB` takes care of the date_time
and that may have depended on chrono, and somewhere along the line that might have gotten removed as well.  
The timer dependency likely came from the use of the timeouts on boost test, which have been removed since they didn't work correctly on windows.  
Also I think i was using boost/chrono at some point in the ipc queue for the timeouts

nightlark:
Yea, `BOOST_DATE_TIME_NO_LIB` removes the attempt to auto-link to the optional compiled part of the library. `chrono` is what made me check the other libraries, since date_time wasn't using it and std::chrono was used in HELICS.

Eventually I think I'd like to try getting the core/application api down to only using Boost headers -- it would make integrating HELICS with other build systems (i.e. ns3/waf) simpler and remove the need to potentially mess with b2/bjam for different compilers/toolsets/static builds by users.

phlptp:
I came across and undocumented flag someone mentioned about just using the system error codes header that didn't require the rest of the system library, so that might be possible.    Filesystem, I am not sure how much it is used in the core library.  I know it is in the test code so it may be possible to pull that out of the core and only include it if tests are built.    

Program options are used all over the place so that would require a lot of work and another good option to replace it.    The other option would be to include a copy of it in the helics source code, that would have the advantage of letting us use the latest features, but it is a decent amount of code to include.

nightlark:
Program options is definitely the big one. Switching to the more frequently updated Asio library (header-only) would be another way to remove the Boost system dependency, it just uses a different namespace.

PR #535, opened 2018-12-12
Close interfaces

phlptp:

### Description

add close functions to the interfaces, and removeTarget functions.  This allows interfaces to be closed dynamically and targets removed.  This is one step along the way of allowing more dynamic federations.


src/helics/application_api/Federate.hpp
```diff
@@ -40,12 +40,12 @@ class CloningFilter;
...
-    enum class states : char
+    /** the allowable operation modes of the federate*/
+    enum class modes : char
     {
         startup = 0,  //!< when created the federate is in startup state
-        initialization = 1,  //!< entered after the enterInitializingMode call has returned
-        execution = 2,  //!< entered after the enterExectuationState call has returned
+        initializing = 1,  //!< entered after the enterInitializingMode call has returned
+        executing = 2,  //!< entered after the enterExectuationState call has returned
         finalize = 3,  //!< the federate has finished executing normally final values may be retrieved
```
kdheepak: Any reason for this change? To be consistent, maybe it should be `startup`, `initialize`, `execute`, `finalize` and `error`.

phlptp: to be consistent with the function calls `enterInitializingMode`, `enterExecutingMode`

phlptp:
looks like this PR had some bad interactions with the blocking_finalize Commit.  Will need to debug that

phlptp:
ok, I think the issue was resolved with the merge, seems to be good again

Issue #426, opened 2018-08-04
Default behavior for subscriptions


phlptp:
What should the default behavior be for subscriptions.  
In the 2.0 alpha release I am removing the distinct function calls for required and optional subscriptions in favor of a single call, and some function to set and retrieve options for the interface handles.  One of the properties that can be set is a required/optional option.  

The subscription call is essentially a wrapper around a nameless namedInput, along with an addSourceTarget call.    So namedInputs like publications will not have a required flag set.  But the question is  should a subscription target be required by default(this can of course be changed)  But should the default behavior be to
1),  create an Error if the requested publication is not present in the federation
2).  be optional.  

I have gone back and forth on this so I don't have a clear answer so I am looking for other opinions?

bpalmintier:
I think the default should be to raise an error, so that simple users know that no errors increases the odds the exchanges they expected actually happened.

Have a published value be optional globally could be an advanced option for folks, perhaps for very large federations, but even then it should be logged as a warning.

Perhaps, though, the best flexibility would be to have the federate be able flag individual subscriptions as optional, such that they would still get the checks for most uses and also have planned for when it is optional.

phlptp:
Are you thinking like a flag at the federate level that can set the default value of whether targets are optional or required?

PR #545, opened 2018-12-17
Flag tests and option updates

phlptp:

### Description

this pull request enables most of the flags for the interfaces (required, optional, single connection, strict type matching (on inputs).  as well as tests for them and corresponding C api functions.  

NOTES:  this will likely not pass all the test yet.  I would like it reviewed before updating all the API files since that will add 100's of files to the review which are autogenerated and likely won't help the review.   So be aware of that.


phlptp:
This also addresses #426

phlptp:
the buffer_data flag is not fully enabled as that really will only matter for dynamic subscriptions which are also not enabled yet.

src/helics/application_api/Endpoints.hpp
```diff
@@ -175,11 +175,19 @@ class Endpoint
     const std::string &getType () const { return fed->getEndpointType (*this); }
     /** get the actual endpoint id for the fed*/
     interface_handle getHandle () const { return handle; }
+    /** implicit conversion operator for extracting the handle*/
+    operator interface_handle () const { return handle; }
```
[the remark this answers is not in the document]
phlptp: the options were to have it be explicit or have several function calls be duplicated for each interface type.    setInfo,  setOption, getOption.    Otherwise you have an inconsistent interface, and then it gets really ugly with the overloads that are different from Federate, and the messageFederate and ValueFederate, and even more so for the combinationFederate.    We can debate it but I went with what I thought was the lesser of two evils   a implicit conversion operator  or a mess of functions and overloads that would be easy to break if anything changes.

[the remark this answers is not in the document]
eberleim: That seems fine then. My initial thought was that this should probably be `explicit operator interface_handle ()` and compel the implementer to do their own `static_cast<>()` but that's just me being pedantic.  This is unlikely to cause any real issues as long as it is used safely, so I'm not particularly worried about it, I just wanted to make sure we'd considered possible side effects.

src/helics/application_api/FederateInfo.cpp
```diff
@@ -99,7 +99,38 @@ static const std::set<std::string> validIntProperties{"max_iterations", "logleve
 static const std::set<std::string> validFlagOptions{
   "interruptible",         "uninterruptible",         "observer",        "source_only", "sourceonly",
   "only_update_on_change", "only_transmit_on_change", "forward_compute", "realtime",    "delayed_update",
-  "wait_for_current_time"};
+  "wait_for_current_time",
+};
+
+static const std::map<std::string, int> optionStringsTranslations{
```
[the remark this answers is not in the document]
phlptp: I will look into the missing option,  but yes these are for the options  
You are right interfaces are publication/inputs/endpoints/filters  

The setting of options for publications and inputs is in ValueFederate.cpp  for endpoints it is in MessageFederate.cpp     Federates.cpp would not know about anything but filters, the other codes handle the other interfaces.

src/helics/cpp98/Core.hpp
```diff
@@ -7,77 +7,61 @@ All rights reserved. See LICENSE file and DISCLAIMER for more details.
...
 #define HELICS_IGNORE_ERROR NULL
 namespace helicscpp
 {
-
 class Core
 {
   public:
     // Default constructor, not meant to be used
-    Core ():core(NULL) {};
+    Core () : core (NULL){};
```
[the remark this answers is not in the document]
phlptp: pretty sure this was a leftover comment from the C++API which has that comment as the object is abstract.  The same is not true for the C++98 and you would need the default constructor for putting it in Vectors or just making an object that was later assigned.

src/helics/application_api/MessageFederate.cpp
```diff
@@ -313,11 +313,6 @@ void MessageFederate::setMessageNotificationCallback (const Endpoint &ept,
 /** get a count of the number endpoints registered*/
 int MessageFederate::getEndpointCount () const { return mfManager->getEndpointCount (); }
 
-void MessageFederate::setEndpointOption (const Endpoint &ept, int32_t option, bool option_value)
```
afisher1: This is the only change I see in MessageFederate.cpp. What I don't see is how the flags are set for an Endpoint when in MessageFederate::registerInterfacesJson/Toml(const std::string &jsonString) it isn't checking for a "flags" key in that function nor calling a function similar to loadOptions found in federate.cpp.

src/helics/application_api/ValueFederate.cpp
```diff
@@ -153,12 +154,25 @@ static const std::string emptyStr;
 template <class Inp, class Obj>
 static void loadOptions (ValueFederate *fed, const Inp &data, Obj &objUpdate)
 {
-    // bool optional = getOrDefault (data, "optional", false);
+    addTargets (data, "flags", [&objUpdate](const std::string &target) {
```
afisher1: @phlptp I was expecting a function like this for MessageFederate.cpp for configuring flag options. Also I noticed that this function does not support all the possible flags listed in helics_handle_options enumeration. I understand if it doesn't make sense for some of the flags to apply to an Input or a Publication but I thought the strict type checking flag should be added as well.

src/helics/application_api/FederateInfo.cpp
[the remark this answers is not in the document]
afisher1: Yes I understand that. I added some more comments in MessageFederate.cpp and ValueFederate.cpp to try and help illustrate my confusion on flag handling for Endpoints, Publications, and Inputs.

src/helics/application_api/MessageFederate.cpp
phlptp: I think I skipped over that since the only valid flags on endpoints are required/optional  but I will add in the logic for checking the flags.

src/helics/application_api/ValueFederate.cpp
phlptp: I think all the flags are available in getOptionIndex now so can be set through the flags field

afisher1: excellent!

src/helics/application_api/MessageFederate.cpp
phlptp: there should be a loadOptions function in MessageFederate.cpp now that does the same as the others in ValueFederate.cpp now

PR #552, opened 2018-12-20
Alter FindZeroMQ module to make it work better with non-system installs

nightlark:

### Description

Changes the FindZeroMQ module to behave better when using it to find the ZeroMQ module included with a HELICS install prefix instead of a system copy that may be incompatible.

ZeroMQ seems to always expect to get installed in `lib` and never `lib64`. An alternative would be using the pkgconfig method of finding ZeroMQ, though the file generated by the ZMQ installer has an absolute install path, so it would be harder to get working as part of a HELICS installer.

### Proposed changes
- If a ZeroMQ install hasn't been specified when using the find_package in config mode, set ZeroMQ_LIBRARY_PATH to the folder ZeroMQ is located in (root directory doesn't work on systems that install to lib64). If set, the library path hint will take precedence over other locations so that it finds the right version -- if it has been set, the user should be pretty confident that it is the copy of ZeroMQ they want
- Disable searching for the zmq.h include folder if the option ZeroMQ_LIBRARY_ONLY is set


config/cmake/FindZeroMQ.cmake
```diff
@@ -26,61 +26,63 @@ else()
...
-	endforeach()
-	foreach(vs ${_VS_VERSIONS})
-		set(ZeroMQ_DEBUG_LIBRARY_NAME "libzmq-v${CMAKE_VS_PLATFORM_TOOLSET}-mt-gd-${ZeroMQ_NAME}")
-	endforeach()
+if (NOT ZeroMQ_LIBRARY_ONLY)
+  find_path(ZeroMQ_ROOT_DIR
+    NAMES
+      include/zmq.h
+    HINTS
+      ${ZeroMQ_REGISTRY_PATH}
```
phlptp: should ZeroMQ_LIBRARY_PATH be part of this section of HINTS as well?

nightlark: I don't think so, since the library path should just contain the so/a/dll -- and there is a ZeroMQ_INCLUDE_PATH that would be the equivalent to specify the location for the headers (maybe that hint should get moved up?).

Issue #551, opened 2018-12-19
getInputType() function returns empty string even though type is specified in the configuration file


Jacobhansens:
**Describe the bug**

getInputType() function returns empty string even though type is specified in the configuration file for that specific input.

**What is the expected behavior?**

getInputType() should return the type specified for the input in the configuration file

**To Reproduce**

Call getInputType() from federate that configures inputs from a configuration file.

**Environment (please complete the following information):**
Ubuntu 16.04
HELICS 2.0 Beta 2

**Additional context and information**

Jacobhansens:
I did something wrong in my setup. I can confirm this is not a bug and that the getInputType() does indeed work in HELICS 2.0 beta 2.

Jacobhansens:
Opening again as @phlptp let me know he has test cases where something is wrong.

phlptp:
For the C and language API's   there is helicsInputGetType  and helicsInputGetPublicationType(),   do we think this is clear enough?

Jacobhansens:
@phlptp, did you make a decision with respect to the C and language API's? I think my original issue is solved in this one but want to make sure you agree before closing.

phlptp:
I pretty much left it as is, since the C API is based on the interface objects functions, not purely on the Federate and Core Objects, so the change was limited to some internal changes.  And Filters in the C api don't support defined types so there is no function for them.    At this point I am inclined to leave it be.

Jacobhansens:
We will leave it as is then. closing out this issue

PR #553, opened 2018-12-20
Type function name update

phlptp:

### Description

This addresses issue #551  

### Proposed changes

- change the getInputType and getOutputType function to getInjectionType and getExtractionType(),  
- make all the interfaces consistent in the use of getName/ getKey functions, 
- add getInterfaceUnits function and remove separate calls for each interface type.

### DIFF
--- a/src/helics/application_api/Federate.cpp
+++ b/src/helics/application_api/Federate.cpp
@@ -1060,7 +1060,14 @@ std::string Federate::query (const std::string &target, const std::string &query
     }
     else
     {
-        res = coreObject->query (target, queryStr);
+        if (coreObject)
+        {
+            res = coreObject->query (target, queryStr);
+        }
+        else
+        {
+            return "#invalid";
+        }
     }
     return res;
 }
@@ -1151,7 +1158,7 @@ void Federate::addSourceTarget (const Filter &filt, const std::string &targetEnd
 {
     if (coreObject)
     {
-        coreObject->addSourceTarget(filt.getHandle(), targetEndpoint);
+        coreObject->addSourceTarget (filt.getHandle (), targetEndpoint);
     }
 }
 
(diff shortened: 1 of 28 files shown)


src/helics/application_api/Federate.cpp
```diff
@@ -1149,24 +1156,38 @@ CloningFilter &Federate::registerGlobalCloningFilter (const std::string &filterN
 
 void Federate::addSourceTarget (const Filter &filt, const std::string &targetEndpoint)
 {
-    coreObject->addSourceTarget (filt.getHandle (), targetEndpoint);
+    if (coreObject)
+    {
```
Jacobhansens: I think having an extra check here is good but should we not add an else statement with a warning or something if this is called on a non coreObject?  (This is present in more places)

src/helics/application_api/Federate.hpp
```diff
@@ -468,13 +458,32 @@ class Federate
...
     bool getInterfaceOption (interface_handle handle, int32_t option);
 
+    /** get the injection type for an interface,  this is the type for data coming into an interface
+    @details for filters this is the input type, for publications this is type used to transmit data, for endpoints
+    this is the specified type and for inputs this is the type of the transmitting publication
+    @param handle the interface handle to get the injection type for
+    @return a const ref to  std::string  */
+    const std::string &getInjectionType (interface_handle handle) const;
+
+    /** get the type for which data comes out of an interface,  this is the type for data coming into an interface
```
Jacobhansens: Should this sentence not be:

get the extraction type for an interface,  this is the type for data coming out of an interface

src/helics/application_api/Inputs.hpp
```diff
@@ -160,18 +160,24 @@ class Input
             }
         });
     }
-    /** get the Name for the subscription
+    /** get the Name/Key for the subscription
```
Jacobhansens: Should this not be input instead of subscription?

src/helics/application_api/Inputs.hpp
```diff
@@ -160,18 +160,24 @@ class Input
...
     @details the name is the local name if given, key is the full key name*/
     const std::string &getName () const { return actualName; }
-    /** get the type of the input*/
-    const std::string &getType () const
+    /** get the Name/Key for the subscription
```
Jacobhansens: Should this not be input instead of subscription?

src/helics/application_api/Federate.cpp
phlptp: other than throwing an error I am not sure what else to do.  The cases where that might be true are when an empty Federate somehow got created and used or after a call to finalize.  I could throw the invalid function call error?

Jacobhansens: I think that is a good idea. It is definitely an edge case but I think an error trow is appropriate.

phlptp: there are probably a few more cases where I need to check it but I will fix them as I see them.

Jacobhansens:
That addressed all my comment. I think you can go ahead and merge it in

phlptp:
can you go ahead and approve if you think it is ready.

Issue #550, opened 2018-12-19
Strict type checking not propagating properly from config files


Jacobhansens:
**Describe the bug**

I am trying to use the new flag for strict type matching. I set it in my JSON config file like this
```    
    ...
    "inputDelay": 0,
    "strict_input_type_checking": true,
    "publications": [
        {
    ...
```
But when looking at the function in NamedInputInfo->checkTypeMatch The boolean flag for type setting is still false. 

**What is the expected behavior?**

I expect that when this flag is set strict type matching will be enforced on all topics for this federate.

**To Reproduce**

Set "strict_input_type_checking" for any federate in the JSON config file while also ensuring a type mismatch is present and run the federate. Currently it will not complain about the mismatch.

**Environment (please complete the following information):**

Ubuntu 16.04
HELICS 2.0 Beta 2

Jacobhansens:
This issue is fixed, closing out this one

PR #554, opened 2018-12-20
propagate the flags to newly created interfaces and create a test cas…

phlptp:
…e to verify this.

### Description

Fixes issue #550

### Proposed changes

Fix issue #550 by adding a creation function inside of federate state that actually does the propagation.  And add a test case that verifies it.

### DIFF
--- a/src/helics/application_api/FederateInfo.cpp
+++ b/src/helics/application_api/FederateInfo.cpp
@@ -87,6 +87,10 @@ static const std::map<std::string, int> propStringsTranslations{
   {"realtime", helics_flag_realtime},
   {"ignore_time_mismatch", helics_flag_ignore_time_mismatch_warnings},
   {"delayed_update", helics_flag_wait_for_current_time_update},
+  {"strict_input_type_checking", helics_handle_option_strict_type_checking},
+  {"buffer_data", helics_handle_option_buffer_data},
+  {"required", helics_handle_option_connection_required},
+  {"optional", helics_handle_option_connection_optional},
   {"wait_for_current_time", helics_flag_wait_for_current_time_update}};
 
 static const std::set<std::string> validTimeProperties{"period",      "timedelta",    "time_delta",  "offset",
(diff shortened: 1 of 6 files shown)


Jacobhansens:
@phlptp, I am not able to get this working on my system. I think I am not fully understanding the correct way to specify this flag. I current do the following in my JSON config file:

```
{
    "name": "dsoBus7",
    "log_level": 2,
    "uninterruptible": true,
    "coreType": "zmq",
    "coreName": "",
    "coreInit": "1",
    "period": 10.0,
    "timeDelta": 10.0,
    "strict_input_type_checking": true,
    "publications": [
     ... (more here)
     "subscriptions": [
     ... (more here)
}
```

However this does not enable the strict type checking. Can you let me know the correct way to set this through a configuration file?

phlptp:
it might have worked if you had done "flags":"strict_input_type_checking"   but I made some updates that will allow what you had to work (I think).

Jacobhansens:
setting `strict_input_type_checking` now works. One question I have is that I see this in federateInfo.cpp:

```
{"strict_type_checking", helics_handle_option_strict_type_checking},
   {"strict_type_matching", helics_handle_option_strict_type_checking},
   {"strict_input_type_checking", helics_handle_option_strict_type_checking},
   {"strict_input_type_matching", helics_handle_option_strict_type_checking},
   {"strictinputtypechecking", helics_handle_option_strict_type_checking},
   {"strictinputtypematching", helics_handle_option_strict_type_checking},
   {"stricttypechecking", helics_handle_option_strict_type_checking},
   {"stricttypematching", helics_handle_option_strict_type_checking},
   {"strict", helics_handle_option_strict_type_checking}};
```

Which seems to suggest I have multiple options for spelling when setting this flag but that is not supported. Do we intend to support more ways of spelling?

Jacobhansens:
Okay that makes sense. I will approve now but in the future we should think about a more consistent way for checks on inputs like you mention

phlptp:
As far as validation goes,  I have done it in other programs but with HELICS you have 4 or 5 layers sometimes all pulling different things from the files so it is challenging to figure out where to do the validation.

Issue #499, opened 2018-10-29
Best Practices for cmake directories


phlptp:
Based on some review of best practices for CMAKE folder organization, we should consider having our /src, /test,  /interfaces folders be able to be used in a standalone context.    So essentially someone could pull out those folders and generate a project from them with CMAKE without the root folder.

nightlark:
Do you have links to some reading on cmake folder organization? I've thought some about how an interface could be separate from the main HELICS repository.

phlptp:
file:///C:/Users/top1/Downloads/git_cmake_conan_how_to_ship_and_reuse_our_cpp_projects__mateusz_pusz__cppcon_2018.pdf  starting around 18 for the cmake stuff.  But the whole presentation is mostly about conan, which we probably want to look at at some point in the not too distant future.

nightlark:
Another thing we should consider doing for the interfaces is make it possible to add additional language interfaces either by checking them out as a subfolder of interfaces, or otherwise create a new interface using the existing shared swig files as a base (I guess that then any tests for a language binding should live within the folder for that interface)

nightlark:
Similar separation for different core types could also be useful -- either to enable easier development of 3rd party core types, or to have better separation of dependencies for a particular core from the rest of the code.

nightlark:
The way it works with llvm is that the extras get checked out into a subfolder, and then you'd run cmake for the root llvm project and build as usual. The other way would be plugins/shared libraries like GridLab-D.

With either approach, each core is responsible for finding libraries it needs (the ZeroMQ core takes care of finding libzmq, other parts of HELICS don't need to know about libzmq).

phlptp:
I think we are getting progress toward this issue.  I think at some point we should make the src/helics folder capable of a standalone cmake build.  At that point I think we could close this issue.

PR #561, opened 2018-12-20
Update fmt library

phlptp:

### Description

Update the fmt library.  The recent updates allow a little cleaner (with some of our tweaks) cmake ,  One less check that shows up.


ThirdParty/fmt/support/cmake/cxx14.cmake
```diff
@@ -2,9 +2,19 @@
...
+	 set(SUPPORTS_ENUM_BASE ON)
+	 set(SUPPORTS_TYPE_TRAITS ON)
+	 set(SUPPORTS_USER_DEFINED_LITERALS ON)
+	return()
+endif()
+
 
 if (NOT CMAKE_CXX_STANDARD)
-  set(CMAKE_CXX_STANDARD 11)
+  set(CMAKE_CXX_STANDARD 14)
```
nightlark: Is this something that could be done right before the add_subdirectory for fmt, and if it sets the CMAKE_CXX_STANDARD then unset it afterward?

I don't remember if we had issues with it, but something like in the fmt cxx14.cmake file:
```
if (NOT CMAKE_CXX_STANDARD)
  set(CMAKE_CXX_STANDARD 14)
endif()
```
might be something to put in the HELICS CMakeLists.txt

phlptp: the CMAKE_CXX_STANDARD flag doesn't work correctly with all combinations of compilers and cmake versions so I haven't found it reliable.

phlptp: There were a few other things in the cmake for fmt that could be useful particularly for addressing #499.

PR #562, opened 2018-12-20
update the jsoncpp library,

phlptp:
change the namespace so it is a little easier to maintain in the future.

Now Json::  is a member of the helics::Json::  instead of Json_helics   this should make things simpler to update in the future and be more maintainable and a little more consistent with the original usage as long as you are in the helics namespace, but still prevent namespace conflicts.

### Description

Update Jsoncpp library with upstream changes and new namespace to make it a little easier to update in the future.


nightlark:
Instead of modifying the ThirdParty jsoncpp library to include the namespace, what about wrapping the #include for jsoncpp in `namespace helics { ... }`? Then we might not need a modified copy of the upstream library.

phlptp:
I tried that at one point but you run into all sorts of chaos with the headers that jsoncpp includes then also being wrapped in that same namespace and then the functions in jsoncpp not in the namespace failing.  and all sorts of other chaos in the program, and even if it did work ending up with an essentially duplicated std

nightlark:
So is the naming conflicts referring to other programs using helics that use their own copy of jsoncpp?

phlptp:
I ran into the issue with GridDyn and helics-fmi

phlptp:
I would also note that cereal does the same thing in wrapping rapidJson internally.

phlptp:
I don't know that I would trust an automated script  as some of the process involves adapting to upstream changes in the cmake and code,  they are all one time things and automating it would probably cause more problems than it solves.    Some documentation would probably be in order though.

Issue #418, opened 2018-08-01
ZMQ2 core type


phlptp:
Add a new communication core (ZMQ_2), that uses a single PUSH/PULL socket for cores and reduces use of REQ/REP sockets.  

This will alleviate the the possibility of running out of sockets if using a high number of federates on a single machine.    Basically copying the zmq core and adding in the necessary hooks so it is active.

shwethanidd:
For some reason, running with helics runner results in huge number of sockets. Investigating that issue.

PR #566, opened 2018-12-22
Zmq comms ss

shwethanidd:
Implementation of ZMQ Single Socket Core for high federate count systems. Since we are using single socket per core, it alleviates the problem of running out of sockets when having many federates (in the order of hundreds)  running in a single system.
It also contains test cases to test the implementation. 

This fixes issue #418.
-
-


src/helics/apps/MultiBroker.cpp
```diff
@@ -84,6 +85,11 @@ static std::unique_ptr<CommsInterface> generateComms(const std::string &type, co
 		case core_type::ZMQ:
 #if HELICS_HAVE_ZEROMQ!=0
 			comm = std::make_unique<zeromq::ZmqComms>();
+#endif
+			break;
+		case core_type::ZMQ_TEST:
```
phlptp: Do we want to leave this as ZMQ_TEST or is another short name more descriptive

shwethanidd: I would prefer to change it to "ZMQ_SS" similar to "TCP_SS"

src/helics/core/zmq/ZmqBroker.h
```diff
@@ -8,6 +8,8 @@ All rights reserved. See LICENSE file and DISCLAIMER for more details.
 namespace helics {
 namespace zeromq {
 class ZmqComms;
+class ZmqCommsTest;
```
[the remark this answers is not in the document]
shwethanidd: I'll change it to ZmqCommsSS

shwethanidd:
@phlptp Can you please check and approve PR

src/helics/core/BrokerFactory.cpp
```diff
@@ -72,15 +72,15 @@ std::shared_ptr<Broker> makeBroker (core_type type, const std::string &name)
...
-            broker = std::make_shared<zeromq::ZmqBrokerTest> ();
+            broker = std::make_shared<zeromq::ZmqBrokerSS> ();
         }
         else
         {
-            broker = std::make_shared<zeromq::ZmqBrokerTest> (name);
+            broker = std::make_shared<zeromq::ZmqBrokerSS> (name);
         }
 #else
         throw (HelicsException ("ZMQ broker type is not available"));
```
phlptp: Error string should be updated

src/helics/core/zmq/ZmqCommsSS.cpp
```diff
@@ -0,0 +1,543 @@
...
+//#include <csignal>
+#include <memory>
+#include <iostream>
+
+static const int DEFAULT_BROKER_PORT_NUMBER = 23414;  // Todo define a different port number
+static const int TX_RX_MSG_COUNT = 20;
+
+using namespace std::chrono;
+/** bind a zmq socket, with a timeout and timeout period*/
+static bool bindzmqSocket (zmq::socket_t &socket,
```
phlptp: Someday (probably not now) we should extract some of these functions that are common between the ZMQ implementations and put them in a separate file to reduce code duplication

src/helics/core/zmq/ZmqCommsSS.cpp
```diff
@@ -0,0 +1,543 @@
...
+    case connection_status::connected:
+    {
+    	ActionMessage cmd (CMD_PROTOCOL);
+        cmd.messageID = CLOSE_RECEIVER;
+        transmit (control_route, cmd);
+    }
+
+    break;
+    default:
+        if (!disconnecting)
```
phlptp: If the code gets here I would guess this is not going to work, since you are doing the same thing as if it were connected.  But in this case that probably doesn't matter since there is only one thread and the receiver and transmitter should have the same close

phlptp:
I added the core type to the lists to test for some of the api tests.  

Also one of the core_tests spits out a bunch of tick messages, and doesn't seem to have any checks so I am not sure what is going there.    And other than a few minor tweaks assuming it passes the newly enabled tests I think it is getting close.

src/helics/core/BrokerFactory.cpp
shwethanidd: Done

tests/helics/core/ZeromqSSCore-tests.cpp
```diff
@@ -0,0 +1,379 @@
...
+#include "helics/common/GuardedTypes.hpp"
+
+//#include "boost/process.hpp"
+#include <future>
+#include <iostream>
+
+namespace utf = boost::unit_test;
+using namespace std::literals::chrono_literals;
+
+BOOST_AUTO_TEST_SUITE (ZMQSSCore_tests, *utf::label("cic"))
```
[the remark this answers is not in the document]
shwethanidd: It's a typo, no idea how the extra 'c' got in there. I've corrected it.

src/helics/core/zmq/ZmqCommsSS.cpp
shwethanidd: Ok, leaving it as is for now.

src/helics/core/zmq/ZmqCommsSS.cpp
shwethanidd: Removed the check now

tests/helics/core/ZeromqSSCore-tests.cpp
```diff
@@ -0,0 +1,382 @@
...
+                    }
+            	}
+
+            }
+        }
+        vFed->finalize ();
+    }
+};
+
+BOOST_AUTO_TEST_CASE (zmqSSMultiCoreInitialization_test)
```
phlptp: I think my only concern is that this test case doesn't have any tests/checks in it.  It also uses a lot of sleeps which is a little concerning?    There should be something to indicate it has passed or failed.

phlptp:
> @phlptp: It's ready for you to take another look. Appveyor fails randomly, this time due to "error: in "messageTimer_tests/basic_test":"

This has been an ongoing occasional error in appveyor still haven't fully tracked it down yet.

PR #567, opened 2018-12-24
User guide

trevorhardy:
### Description

This is the "official" beta-worthy version of the User Guide. This branch has been merged with the latest `develop` branch prior to submitting this PR.

### Proposed changes
I'm assuming we'll want to release this with v2.0. It includes three examples that worked for the developer of said examples but I haven't had time to verify. A fourth example (using ns-3) is underway but won't be ready by the time of v2.0 release. Most of the supporting text for the example has been written but the full co-sim has not been executed. 

I'm not assuming what I've written is without flaws and it is incomplete in some places. Its probably 90% of what I hoped it would be and I'm going to call that good enough for v2.0 release. I plan on working on it after the v2.0 release, as well.


trevorhardy:
Just added several of you as PR reviewers; treat this as a suggestion rather than a hard request. Remove yourself if you can't perform the review or feel free to give a cursory review before approval.

trevorhardy:
And I've already noticed that I didn't complete the section on "Integrating a New Simulator" as well as not providing enough instruction on using helics_cli. I'll be working in on the former today; I assume it will not end up as a part of this PR (right?). The later I will need some help to understand better before I can write on it.

trevorhardy:
I've just realized I didn't finish the "Timing" section, either. I'm going to cancel this PR and put another one together by COB Thursday.

PR #568, opened 2018-12-27
Core type lists

### LINKED ISSUE #563: application_api_ci_test not running completely
**Describe the bug**
I don't think all the ci test are being run in travis.    I ran the tests locally and discovered some failures from a recent commit that should have been caught by the CI test execution specifically in the application_api_ci_tests  and it is running a bit faster than I would expect,  this needs to be checked somehow not sure how though.

phlptp:

### Description
This fixes #563 and the issues that fixing that exposed in the application_api_tests.
The test are passing reliably( ran them 4 times)  So we will see how the checks for the PR do as well.

### Proposed changes

- a few clang-tidy fixes are in place as well 
- a few internal enumeration renames
-


nightlark:
Are there still some test cases that occasionally fail on Travis?

phlptp:
There is one I know of.   I added a few things that would make it fail a little less often, but that particular test does not have a defined time pattern so It isn't guaranteed to meet the test by design.    I am removing a subscription by name, so it has to route through the root broker in some cases which depending on threading isn't guaranteed to though by the time the next data point comes.    At some point I suppose I should tweak the test with some conditional logic to just try again if it fails.  

I have seen one other failure mode but that hasn't replicated itself in a while, so I am not sure if that is actually fixed or not.

phlptp:
@afisher1  if you are around can you take a look at this.  I would like to merge the zmq stuff before this one since there may be some issues merging and I would prefer to do that in this branch than that one.

afisher1:
@photo I have viewed the changes but I'm not sure how I can verify that this has fixed #563 nor do I know what was the cause.

PR #570, opened 2018-12-29
(feature/569) adding a getDestination() function for Endpoints.

afisher1:

### Description

I'm adding a getter function for the targetDest property of the Endpoint class.


phlptp:
We probably should add a similar function to the C api.  And looking at that the C function should probably be helicsEndpointGetDefaultDestination, which makes me wonder if we should make the C++ function setDefaultDestination  and getDefaultDestination  to make everything consistent?  What do you think?  @nightlark  any opinions?

phlptp:
To be good we probably also need to add a statement to a test case somewhere along the line.

phlptp:
@afisher1  I added the functions to the C api and C++98 API and changed the names on the C++ api so it matches all the others.  I think I added everything take a look see what you think?  @nightlark  you will have to review this see if we are missing something.

afisher1:
@phlptp that looks correct to me.

Issue #560, opened 2018-12-20
Revise command line arguments


nightlark:
Go through the currently supported command line arguments and make sure they are consistent and don't conflict with each other. Support for -h (instead of?) -?

phlptp:
Whats wrong with -?    I use that all the time.    But we probably should go through and make sure the available options are consistent with conventions and internally consistent.

nightlark:
Mostly that -h seems to be more typical as a short form (for unix) help, and since most users will likely be using the unix form of arguments they should be handled well. Supporting both would probably be good, since it looks like some unix commands also support -?.

phlptp:
For Federate Info 
```
static const ArgDescriptors InfoArgs{
  {"broker,b"s, "address of the broker to connect"s},
  {"name,n"s, "name of the federate"s},
  {"corename"s, "the name of the core to create or find"s},
  {"core,c"s, "type of the core to connect to"s},
  {"type,t"s, "type of the core to connect to"s},
  {"coretype"s, "type of the core to connect to"s},
  {"offset"s, "the offset of the time steps"s},
  {"period"s, "the period of the federate"s},
  {"timedelta"s, "the time delta of the federate"s},
  {"rttolerance"s, "the time tolerance of the real time mode"s},
  {"rt_lag"s, "the amount of the time the federate is allowed to lag realtime before corrective action is taken"s},
  {"rt_lead"s,
   "the amount of the time the federate is allowed to lead realtime before corrective action is taken"s},
  {"coreinit,i"s, "the core initialization string"s},
  {"coreinitstring"s, "the core initialization string"s},
  {"maxiterations"s, ArgDescriptor::arg_type_t::int_type,
   "the maximum number of iterations a federate is allowed to take"s},
  {"log_level"s, ArgDescriptor::arg_type_t::int_type, "the logging level of a federate"s},
  {"separator"s, "separator character for local federates"s},
  {"inputdelay"s, "the input delay on incoming communication of the federate"s},
  {"outputdelay"s, "the output delay for outgoing communication of the federate"s},
  {"brokerport"s, ArgDescriptor::arg_type_t::int_type, "port number for the broker priority port"s},
  {"localport"s, "port number for the local receive port"s},
  {"autobroker"s, ArgDescriptor::arg_type_t::flag_type,
   "tell the core to automatically generate a broker if needed"s},
  {"port"s, ArgDescriptor::arg_type_t::int_type, "port number for the broker's port"s},
[...]

PR #574, opened 2019-01-02
Tweak a few of the command line options

phlptp:

### Description

Address most of issue #560.
remove a few duplicated options
move to --coretype, or -t as the preferred way to specify which core type to use on the command line
tweak a few other options

### DIFF
--- a/src/helics/core/NetworkBrokerData.cpp
+++ b/src/helics/core/NetworkBrokerData.cpp
@@ -21,8 +21,9 @@ namespace helics
 static const ArgDescriptors extraArgs{
   {"interface"s, "the local interface to use for the receive ports"s},
   {"local_interface"s, "the local interface to use for the receive ports"s},
-  {"broker"s, "identifier for the broker"s},
-  {"broker_address,b"s, "location of the broker i.e network address"s},
+  {"broker,b"s,
+   "identifier for the broker, this is either the name or network address use --broker_address or --brokername to explicitly set the network address or name the search for the broker is first by name"s},
+  {"broker_address"s, "location of the broker i.e network address"s},
   {"brokername"s, "the name of the broker"s},
   {"brokerinit"s, "the initialization string for the broker"s},
   {"max_size"s, ArgDescriptor::arg_type_t::int_type, "maximum message buffer size (16*1024)"s},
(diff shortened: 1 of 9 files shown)


src/helics/core/NetworkBrokerData.cpp
```diff
@@ -21,8 +21,8 @@ namespace helics
 static const ArgDescriptors extraArgs{
   {"interface"s, "the local interface to use for the receive ports"s},
   {"local_interface"s, "the local interface to use for the receive ports"s},
-  {"broker,b"s, "identifier for the broker"s},
-  {"broker_address"s, "location of the broker i.e network address"s},
+  {"broker"s, "identifier for the broker"s},
+  {"broker_address,b"s, "location of the broker i.e network address"s},
```
nightlark: Looking at the broker and broker_address options, it seems like the difference is that broker searches by name first, then tries to extract an address -- are there use cases where having the option to skip the broker search by name is required?

phlptp: Assuming I am understanding the question.  It is theoretically possible to have a broker in the local space but you want to connect with a broker on a different machine or process with the same name.    In which case you would potentially need both a name and address, which is why there are different options.    In most cases you need one or the other

nightlark: Okay, so that's a scenario where the broker address would need to be given to skip the local name search. I'm trying to find where the broker set using the option for a federate info structure is used -- I'm wondering if the search order used for that might be closer to the behavior of the broker option here rather than broker_address.

phlptp: for the federateInfo   everything in the broker option gets assigned to --broker when passed into the core.    It is equivalent to passing "--broker=..." in the coreinitstring

phlptp: so broker can be either the name or address

nightlark: Would it be more accurate to say that the --broker option in the broker is the broker's equivalent of the --broker option in federateinfo since both are looking for a name or address? In which case, changing the descriptions for the broker to say that --broker searches by name or address and that --broker_address sets the broker to search by name on (then -b could remain the short option for --broker).

phlptp: I updated the help line for the --broker argument and moved the -b back to it.

Issue #578, opened 2019-01-02
Broker connection time out hard coded in ZMQ core


Jacobhansens:
**Describe the bug**

Large federations can fail with the error `zmq broker connection timed out (2)` because the time out of the zmq polling is set to a fixed value and proper mechanism for retry is not in place. I have consistent failures at 1000 federates each with 100 publications on a single machine. 

With local source code changes to retry connection I am able to run the large federations without any issues.

**What is the expected behavior?**

I would suggest that the user is able to set this time out value to ensure the value is sensible for the size of the federation. We already have a timeout property of the core initialization and that might be a good option to use this value.  

**Environment (please complete the following information):**

Ubuntu 16.04
Helics 2.0 beta 2

phlptp:
Can you PR the changes you made for the retry?  I will take a look at the timeout

Jacobhansens:
I think the current knobs you describe are sufficient. I think they provide enough flexibility while not overloading the user with options to set. I would suggest as part of this we check all timeouts in each core to make sure they all map to one of these user specified options. I would furthermore suggest the following couple of things as improvements:

- Make sure the documentation that @trevorhardy is working on fully covers how each of these work and in what cases the user should be modifying them.  
- Either add more descriptive warning and error messages or create a portion of the documentation that includes more description for each and potential solutions to avoid them

phlptp:
leaving this open while we check the other comms and for documentation issues for the next week.

PR #580, opened 2019-01-03
update the timeout used in zmq comms

phlptp:

### Description

This partially addresses Issue #578

### Proposed changes

- Have the timeout on the ZMQ comms use the connectionTimeout variable from CommInterface instead of hard coded value
- use 0 timeout for some zmqComms_ss since 10ms was not particularly useful.

### DIFF
--- a/src/helics/core/NetworkBrokerData.cpp
+++ b/src/helics/core/NetworkBrokerData.cpp
@@ -24,6 +24,7 @@ static const ArgDescriptors extraArgs{
   {"broker,b"s,
    "identifier for the broker, this is either the name or network address use --broker_address or --brokername to explicitly set the network address or name the search for the broker is first by name"s},
   {"broker_address"s, "location of the broker i.e network address"s},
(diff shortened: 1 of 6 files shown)


phlptp:
@Jacobhansens  I would like to get any retry code you had in as well

Jacobhansens:
What I had was a very quick but not too good solution. I basically introduced the possibility of an infinite loop. I will get a better version together today and share that here for you to look at.

Jacobhansens:
@phlptp, please take a look at the retry logic I added. I tested it on my end and it works for me but I would appreciate your review on this as well.

Jacobhansens:
@Phil, if you want I can go ahead and approve this one, but wanted to check with you as you mentioned  this might go beyond the ZMQ core and I am not sure if you want that to be part of this PR.

phlptp:
I think we should revisit this at a later time but I am comfortable merging it.  I think this is part of a bigger issue on network resilience that isn't going to be resolved in the next week.  And will probably evolve as well expand the usage and scale.

phlptp:
@Jacobhansens  go ahead and approve unless you see something that should be fixed.

PR #594, opened 2019-01-15
Fix copyright encoding issues

kdheepak:


src/helics/common/cppzmq/zmq_addon.hpp
```diff
@@ -38,7 +38,7 @@ namespace zmq
 /*
     This class handles multipart messaging. It is the C++ equivalent of zmsg.h,
     which is part of CZMQ (the high-level C binding). Furthermore, it is a major
-    improvement compared to zmsg.hpp, which is part of the examples in the �MQ
+    improvement compared to zmsg.hpp, which is part of the examples in the ?MQ
```
phlptp: This looks screwed up now

kdheepak: Good catch. Fixed in 574bb12

Issue #610, opened 2019-01-19
Specify what is part of the public API (for non-breaking code changes)


nightlark:
For the note on future versions (https://github.com/GMLC-TDC/HELICS-src/blob/master/CHANGELOG.md), we should be more specific on what is parts of the code will be covered by the non-breaking code change guarantee.

I think non-"*_impl" headers in `src/helics/application_api`, `src/helics/shared_api_library`, and `src/helics/cpp98` should be covered since they are intended to be a main way of accessing things, along with headers in `src/helics`. `src/helics/common` should probably not be covered, since otherwise `cppzmq` is part of the public API, along with some other things that we want to change. There's probably some parts of `src/helics/core` that we might not want to hold to the non-breaking code guarantee.

PR #613, opened 2019-01-20
Update changelog and link to a new document with a more precise defin…

phlptp:
…ition of the helics Public API.

- [x] I agree to license my contributions under the same license as in this repository (BSD 3 clause)

### Description

Remedy Issue #610 by making a note in the Changelog and an additional documentation file with more details.


phlptp:
The load arguments call is made in all the loadArguments functions for each of the different apps.  

That is probably not the best design since loadArguments is not a virtual function but I can't make it private since it is needed in the derived apps and I don't really want to make it virtual since it is used in the constructors which would get all kinds of confusing.  This will definitely be redesigned when we remove program_options but I don't want to do anything right now.    But I will make a note of it in the changelog and linked document

PR #650, opened 2019-03-27
remove some line breaks in the license file 

### LINKED ISSUE #649: Github license detection
**Describe the bug**
The license is not being automatically detected by Github

Automatic detection of BSD-3-Clause license

Github should be able to automatically detect the license, and I am not sure why this isn't happening.  We need to investigate further to determine if anything can be done so the license automatically is detected.

phlptp:
so github detect the license.

### Description

If merged this pull request will change some line breaks in the license file so Github detects it properly

close #649

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,8 +1,6 @@
 BSD 3-Clause License
 
-Copyright (c) 2017-2019,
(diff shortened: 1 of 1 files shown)


phlptp:
I will merge this once the CI passes

Issue #605, opened 2019-01-18
Boost extraction


nightlark:
Issue for tracking the changes that would be needed to remove the transitive dependency on Boost currently required of projects using HELICS.

- [x] Replace `program_options` in `common/argParser.h` with an alternative for parsing arguments (ex: https://github.com/CLIUtils/CLI11)
- [x] Replace Boost `asio` with the upstream `asio` library (https://think-async.com/Asio/) -- removes dependency on Boost `system` without using any undocumented defines. `common/AsioServiceManagerFwd.hpp` should then be possible to remove
- [x] Remove use of `filesystem` from `application_api/FederateInfo.cpp`, and `common/argParser.h` to break the transitive dependency. Also used in `apps/helicsApp.cpp`, `apps/helicsConfigMain.cpp`, `apps/Player.cpp`, `apps/Recorder.cpp`, and `apps/Tracer.cpp` -- removal would allow people compiling HELICS that don't care about the tests to skip the build step for Boost
- [x] Extra: check for Boost headers included by public API headers -- removing would make so user projects don't need to find any parts of Boost (ValueConverter_impl.hpp uses boost::iostreams; this isn't part of the public API and involves serialization using cereal, which is C++ only and won't work with other languages. If a user chooses to use it for serialization, they need to include their own copy of boost includes)

phlptp:
does federateInfo use boost filesystem other than in argParser?  

If we wanted to remove it the applications,  what is a replacement for boost::filesystem?  short of C++17 filesystem I presume we don't need to remove it from the test system,  that uses the boost test and I have no desire to try to replace that.  

for std::filesystem we would need gcc 8.0 or clang 7.0 as a minimum version.

nightlark:
It looks like federateInfo doesn't use boost filesystem outside of the argParser.

Based on the way we're using it, the functions provided by https://github.com/wjakob/filesystem would be pretty close to a drop in replacement (the Mac situation with C++17 filesystem is a mess -- seems to be the inspiration for the headers). PlatformSTL seems to be another alternative.

Anyway, for most users getting a copy of Boost libraries shouldn't be a big deal. It might save some small hassle for people that need copies of HELICS built with different compilers.

phlptp:
For the test system the main thing is checking if files exist and some of removing files and paths, Also the getting the temp directory.

phlptp:
[...]

PR #653, opened 2019-04-03
Replace Boost.Asio with Asio

nightlark:
### Description

If merged this pull request will replace the use of Boost.Asio with Asio as part removing dependencies on compiled Boost libraries for issue #605. 

### Proposed changes
- Replaces uses of Boost.Asio with Asio
- Renames Asio Service Manager to Context Manager
- Forward declares AsioContextManager where needed instead of using a header file
- Removes use of boost::system::error_code
- Updates parts of Boost libraries used in CONTRIBUTORS.md 
- Adds Asio to libraries used in CONTRIBUTORS.md
- Adds Asio as a git submodule
- Enables git submodule support on AppVeyor


src/helics/common/AsioServiceManagerFwd.hpp
```diff
@@ -5,33 +5,10 @@ All rights reserved. See LICENSE file and DISCLAIMER for more details.
...
-
 class AsioServiceManager;
 
-#if (BOOST_VERSION / 100 >= 1066)
-namespace boost
-{
 namespace asio
 {
 class io_context;
 using io_service = io_context;
```
phlptp: Can we get rid of the alias now?

phlptp: Actually can we get rid of this entire file?

src/helics/common/AsioServiceManager.h
```diff
@@ -27,18 +27,18 @@ SPDX-License-Identifier: BSD-3-Clause
 #include <string>
 #include <vector>
 
-#include <boost/asio/io_service.hpp>
+#include <asio/io_context.hpp>
```
phlptp: We may want to rename this file AsioContextManager to better capture the update

phlptp:
I think this looks good assuming everything works correctly.  We may want to change asioServiceManager to AsioContextManager so the names don't get confusing at some point.

src/helics/common/AsioServiceManagerFwd.hpp
nightlark: It should be safe to get rid of the alias, maybe we can get rid of the file -- I think there's at least one file that includes it for the forward declaration.

phlptp: I think the forward declare might be small enough that it may make sense to just do that where it is needed.

nightlark:
Reminder: make sure docs are up to date for cloning git submodules

PR #658, opened 2019-04-19
move jsoncpp to a submodule

phlptp:
### Description

If merged this pull request will get rid of included modified jsoncpp amalgamated sources and replace them with a jsoncpp submodule

### Proposed changes

- move jsoncpp to a submodule
- remove jsoncpp sources from repo and stop using helics::Json namespace and just use jsoncpp library instead


phlptp:
This appears to be a bug in jsoncpp,  and there seems to be some fixes in PR or in recent.  I will merge those in with the fork I am pulling from.

PR #662, opened 2019-05-02
Cli11

phlptp:
### Description

If merged this pull request will replace the program_options command line parser with CLI11.  

### Proposed changes

- remove boost program options and some associated helper functions,  replace with CLI11.


phlptp:
the swig.i files are wrong now that I look at the changes,  I will need to fix those and update the included interface files again.

phlptp:
@nightlark @kdheepak I reversed the changes to the C api, so that has identical signatures.  I think this is ready to look at.  
Once this is merged the next step is to do some work on the cmake to actually pull out the boost libraries.  I think it is ready to take a look,

nightlark:
The formatting for help messages looks better than before, less duplicated lines, and nice grouping of arguments into categories (e.g. recorder cloning/capture_group).

Issues that I see:
* Using `helics_app <subcommand> --help` just prints `-h,-?,--help` as the options. It would be nice if this printed out the options similar to running `helics_<player|recorder> --help`
* Similar to the above, a `--help-all` option for printing help for all the subcommands (though verbose) might be nice for a user that just wants to see everything
* For `helics_broker --help` it looks like it just prints the config-file option, version, help, autorestart, and quiet (group) options. It would be nice to see some of the options like `log_level`, `coretype`, etc
* `helics_broker term` -- I'm guessing this is a work in progress? Right now, it feels like `helics_broker term --help` should print out the same arguments as for helics_broker. Running the subcommand, then typing in `--help`, `-h`, `-?` terminates after throwing an instance of 'CLI::CallForHelp'; other text entered throws 'CLI::ExtrasError'. `help` also throws 'CLI::ExtrasError', for when it is working, it would make sense for it to recognize `help` without any dashes as a valid terminal entry.

nightlark:
I think there might also be some unintended things going on with the log level argument to the broker; it doesn't seem to recognize `log-level`, and the text versions of the log levels (summary, warning, etc) aren't recognized.

phlptp:
@nightlark   try it again with the latest, I think I addressed most of the issues you had.
There was also one PR from CLI11 that I think I need to get in yet that dealt with some alignment issues.  I don't know that it needs to be in this PR or not

nightlark:
Those changes look good, just a few more (hopefully small) things.

A note on the output for `helics_app --help` saying that `--help` is supported for subcommands. I think this is the only change that is a "must-have" based on the reaction I got when showing it to a user.

This might be one of the alignment issues you mentioned -- it feels like there should be a blank line (or other separator) between the options output and the next "app" section.

So something like from this:
```
    --quiet
Federate Info Parsing
Usage: [OPTIONS]
```

To this:
```
    --quiet

Federate Info Parsing
Usage: [OPTIONS]
```

Also from talking to the aforementioned user: helics_broker help is really long and the short version is a better length (with the extra options in a man page, online, etc).
- My thoughts are that not having a way to get the other options might be hard for Windows users, maybe a verbose help option that shows all the extra broker options, or filtering the options listed based on the core type selected would be improvements. This might be one to revisit later based on changes in documentation.

phlptp:
> My thoughts are that not having a way to get the other options might be hard for Windows users, maybe a verbose help option that shows all the extra broker options, or filtering the options listed based on the core type selected would be improvements. This might be one to revisit later based on changes in documentation.

Can you expand on this a bit more, especially about windows users.  As far as I know the options are the same for windows users, except options can optionally have a '/' like windows standard options

nightlark:
> Can you expand on this a bit more, especially about windows users. As far as I know the options are the same for windows users, except options can optionally have a '/' like windows standard options

As far as I know, Windows doesn't have man page support like Linux does, so if we added a man page that describes all of the command line options (and switched the default help message to the previous short version), it would be useful to have some way for Windows users to print out a full list of the supported options (and Linux users that installed to a custom prefix, that isn't in the search path for man pages).

phlptp:
CLI11 has a distinction between help and help_all which we could tweak for our purposes though that will probably require some thought.  I think I want to eventually move to generating the help as a string so the calls work correctly instead of directly printing to cout in a callback for the help generation in CLI11.  But I need to think about how to do that some more.    I have never generated man pages so it sounds like a good idea.

phlptp:
lets make an issue about making and installing man pages for the app executables.

src/helics/application_api/FederateInfo.cpp
```diff
@@ -225,113 +216,136 @@ int getOptionIndex (std::string val)
...
+    app->add_option ("--name,-n", defName, "name of the federate");
+    app->add_option ("--corename", coreName, "the name of the core to create or find")->ignore_underscore ();
+    app->addTypeOption ();
+    app->add_option ("--coreinitstring,-i,--coreinit", coreInitString, "The initialization arguments for the core")
+      ->transform ([] (std::string arg) {
+          arg.insert (arg.begin (), ' ');
+          return arg;
+      })
+      ->ignore_underscore ();
+    app->add_option ("--broker,--broker_address", broker, "address or name of the broker to connect")
```
kdheepak: We should remove the underscore here, so that it is consistent with the other arguments.

src/helics/application_api/FederateInfo.cpp
```diff
@@ -225,113 +216,136 @@ int getOptionIndex (std::string val)
...
+      ->ignore_underscore ();
+    app->add_option ("--brokerport", brokerPort, "Port number of the Broker")
```
kdheepak: Or add an underscore here. I'd prefer removing all underscores from the command line arguments.

phlptp: with the ignore_underscore(); flag in CLI11 underscores are essentially ignored in the definition and on the command line.  so --brokerport, --broker_port --brokerport_  are all valid and go to the same option

phlptp: But it should probably be made consistent in the definitions regardless of what is valid or not

phlptp: Since this isn't critical I am going to merge this and address this in the subsequant PR which is already in progress to avoid having to do a rebase.

nightlark: I think one of the output messages from CLI11 used the variant without the underscore when I'd used the argument with an underscore, so using the non-underscore variant in the help messages might be more consistent with any error messages CLI11 might print.

PR #664, opened 2019-05-04
Remove boost filesystem

phlptp:
### Description

If merged this pull request will remove boost compiled libraries from the core helics library and remove the requirement from the installed static libraries and the linkages from the shared libraries.  

### Proposed changes

- change to using a standalone filesystem library.   
- remove references to boost filesystem in the static libs
- make the build system work with cmake 3.14
- change helics_config to not use boost filesystem.


phlptp:
This includes everything from the CLI11 PR so it will look a lot smaller once that is merged.

src/helics/core/BrokerBase.cpp
```diff
@@ -108,11 +108,14 @@ std::shared_ptr<helicsCLI11App> BrokerBase::generateBaseCLI ()
...
-    hApp->add_option ("--federates,-f,--min_federates,--minfed,-m", minFederateCount,
-                      "the minimum number of federates that will be connecting");
+    hApp
+      ->add_option ("--federates,-f,--minfederates,--minfed,-m", minFederateCount,
+                    "the minimum number of federates that will be connecting")
+      ->ignore_underscore ();
     hApp->add_option ("--name,-n,--identifier", identifier, "the name of the broker/core");
-    hApp->add_option ("--maxiter,--max_iterations", maxIterationCount, "the maximum number of iterations allowed",
-                      true);
+    hApp->add_option ("--maxiter,--max_iterations", maxIterationCount, "the maximum number of iterations allowed")
```
nightlark: Is `max_iterations` one of the arguments we want to remove the underscore from?

phlptp: probably, I will change it in the next commit

phlptp:
@nightlark  I think this is ready for review.  It seems to be working now.  The changes to filesystem got it to work in our cases.  though I am pretty sure they would not work in the authors intended case.  Seemed to be some issue with the wide strings and the windows function calls.

nightlark:
Did the original function seem to work when passed a `std::wstring` or a string prefixed with `L`?

phlptp:
I never really checked the conditions it worked under.    It was converting the path to a wstring then passing it into the `W` version of the function.   I suspect something in the wstring conversion was going wrong, but I didn't want to go down that rabbit hole when there was a simpler fix for our purposes.   I was going to make an issue with the filesystem library maintainer and see what they say about it.

gulrak:
Hi, I'm the maintainer of the filesystem library, and I just wanted to add some insights, as I was stumbling across this in a search result and was worried about a possible bug:

I reverted the GetFileAttributes-Change back to:

`::GetFileAttributesExW (p.wstring ().c_str (), GetFileExInfoStandard, &attr)`

 and ran the tests and they failed, but a quick debugging showed that there was a significant change in the filesystem.hpp besides reformatting, that broke the code:

`if (!(utf8_state = consumeUtf8Fragment(utf8_state, (uint8_t)*iter++, codepoint))) {`

was changed into:

` if ((utf8_state = consumeUtf8Fragment (utf8_state, (uint8_t)*iter++, codepoint)) != 0)`

and that is clearly not the same.

I couldn't find a single version in my repository with this bug (and it would break my tests and builds) so it must have happened after checkout/clone.

I would btw. suggest to use an updated version ([v1.1.4](https://github.com/gulrak/filesystem/commits/v1.1.4)) as there was some serious issue with the recursive_directory_iterator in earlier versions.

phlptp:
@gulrak You are right of course.  I think it was a matter of getting rid of a warning in a function we weren't currently using and getting it working on the functions we were using, then the whole thing got reformatted.  

[This](https://github.com/GMLC-TDC/HELICS-src/blob/clang_msys/ThirdParty/helics_includes/filesystem.hpp) is the version that will get merged soon,  which is based on (v1.1.4)  I was going to file a bug report in your library but then saw that you fixed most of the issues in the newest release.    The differences in the file should be much smaller,  probably a few things related to MSYS and the warning fix(hopefully correctly this time).  

Thanks for letting getting in touch!  It was a quite exciting find to come across your library.  We will probably be using it in a few other projects as well.

gulrak:
Actually that one looks like a v1.1.2 that is still missing a fix for POSIX systems that can lead to a crash in directory_iterators and that is the reason I made the v1.1.4. Im really sorry, that I had to do two releases in a short time, but I wanted to have a good feeling with the state of the latest release.

Thanks for using ghc::filesystem. I needed it myself, it was fun to code and I'm happy if it is of use for others as well. It always feels better to write software that is used. If you have some trouble when using it, I'm happy to help if I can.

Adding MingW-builds to my CI configs is still on my (non-public) roadmap, as I currently do run the tests for MingW manually from time to time, but it sure is better to have them automated too, and hopefully I get to that soon.

phlptp:
Thanks for the heads up I will make sure to update to the latest release before we merge that branch!

gulrak:
Btw, to make future versions easier to integrate, I added a fix for the ERROR_DIRECTORY_NOT_SUPPORTED issue on MingW on the trunk similar to yours, and the "critical" line in fromUtf8 is now:

`if ((utf8_state = consumeUtf8Fragment(utf8_state, (uint8_t)*iter++, codepoint)) == S_STRT)`

(The compare-to-zero is actually with the start state of the decoding state machine, so I choose the symbolic constant instead of the magic 0 and it is more declarative than the simple negation before was.)

Issue #449, opened 2018-09-07
remove xcode 6.4 support


phlptp:
[xcode 6.4 removal](https://travis-ci.org/GMLC-TDC/HELICS-src/jobs/425599734)
```
 Running builds with Xcode 6.4 in Travis CI is deprecated and will be removed in January 2019. If Xcode 6.4 is critical to your builds, please contact our support team at support@travis-ci.com to discuss options.
```

looks like travis will be removing support for Xcode 6.4.  
I think this means we remove the Xcode 6.4 travis build in the 2.1 release later this year.  Probably add a 9.X build for macs at the same time.   We are probably not changing the code level at that point, we will just drop the tested support, and anyone who uses it will have to report bugs if any compile errors show up.  

@nightlark  don't do anything with this until we are getting ready for a 2.1 release probably December.

nightlark:
There's a 9.x build that I forgot to remove from a PR that was merged a few weeks ago -- I've been thinking of removing that one since we are triggering a lot of builds on each commit when a PR is open.

nightlark:
It's around the time frame that xcode 6.4 support will be removed -- we should probably clean up the build matrix for XCode versions tested. Looking at the market share for the versions of macOS in the last 5 years and the age of hardware supported, it probably makes sense to target testing a XCode 9.x (macOS Sierra+) build for most commits, with testing of any older versions for PRs to master and/or develop.

phlptp:
I would be ok with that.  If you want to tweak the build matrix for develop I think you can do so now.  Leave master as is until after the 2.0 final comes out next week.    I think it reasonable to run 1 newer mac build for each build and 1 or 2 older ones on the nightly build or on develop/master PR's

phlptp:
@nightlark   what is the status of updating the build matrix to drop xcode 6.4.  It looks like some of the modules for HELICS 2.1 don't compile and I can't say I am all that inclined to fix them for a version unlikely to be used.

nightlark:
I have a branch with it removed from the builds, and attempted switch to Xenial for the Linux builds that needs to get removed (undefined Boost functions during linking, maybe the wrong libraries getting found). I think the switch to Xenial images on Travis might wait until we start using Docker for the builds.

PR #670, opened 2019-05-07
Build matrix changes

nightlark:
### Description

Removes the old XCode 6.4 build from Travis and replaces it with an XCode 8.3 build on macOS 10.12 (Sierra) for the new minimum. Adds an XCode 10.2 build on macOS 10.14 (Mojave).

### Proposed changes
- Remove XCode 6.4 (issue #449) and 8gm builds from Travis
- Add XCode 8.3 build on macOS 10.12 for PRs/commits to master and PRs/nightly to develop
- Add XCode 10.12 build on macOS 10.14 for all commits/PRs
- Update to GCC 6 for all commits/PRs and nightly
- Update to Clang 5 for all commits/PRs and nightly
- Switch Clang 3.6 to only build for PRs/nightly on develop and PRs/commits to master
- Switch GCC 4.9 to build for PRs and nightly on develop, and all commits to other branches
- Always use stx::optional
- Bump up test timeout to 600s (10min)
- Skip running tests on GCC 4.9 (Boost 1.61) and Clang 3.6 (Boost 1.58)


nightlark:
Thoughts on switching the GCC 4.9.3 (and/or Clang 3.6) builds to only run on PRs/commits to master and develop? We seemed to get a bit of a long queue earlier.

nightlark:
Updating to GCC 8.X and Clang looks like it will have to wait for now (probably until we move to a docker image). I've updated the builds to use the newest versions whitelisted on the Travis Trusty VMs, GCC 6 and Clang 5.

The gcc 4.9 and clang 3.6 (and XCode 8.3) builds are actually set to build on PRs/commits for master, and PRs/nightly for develop -- are there any cases where building for the PR wouldn't cover the same thing as push?

The current build matrix looks like:
GCC 6: All commits/PRs, nightly; nightly Valgrind+Coverage
Clang 5: All commits/PRs, nightly; nightly TSAN+UBSAN
Clang 3.6: Commit/PR master, PR/nightly develop
GCC 4.9: Commit/PR develop, All commits/PRs to other branches
XCode 8.3: Commit/PR master, PR/nightly develop
XCode 10.2: All commits/PRs, nightly

For builds using docker image(s), what compiler versions would we want?
- GCC 8
- Clang 6+ (8?)
- GCC 4.9
- Clang 3.6

phlptp:
I think for the docker builds it probably makes sense to have the oldest supported and the newest stable compilers.   Be that in docker or just in the build image

phlptp:
Is this ready for review?

scripts/run-ci-tests.sh
```diff
@@ -77,11 +77,11 @@ else
                 ;;
             *continuous*)
                 TEST_CONFIG="Continuous"
-		CTEST_OPTIONS+=" --timeout 360"
+		CTEST_OPTIONS+=" --timeout 480"
```
phlptp: do the new Xcode tests take longer?

nightlark: It seems like they might. The Application API tests took 470 seconds for one of the runs that passed. I noticed one of the Linux tests taking longer than the previous timeout.

nightlark:
It's almost ready to merge, there's some things that are weird about the PR build for macOS that I'm not sure about (maybe just cache), and I'm going to see if using brew to install some of the dependencies is faster now that the macOS versions are new enough that precompiled bottles are provided.

nightlark:
Using brew to install dependencies on macOS takes just under 30s, which is probably similar to the time it took to download the cache for dependencies (possibly shorter, because the Travis VMs have a copy of Boost already installed).

The application API tests are running significantly slower for all builds -- ~300s before to 440s now. Even the GCC 4.9 build and Clang 3.6 builds seem to be taking longer, the main change that comes to mind that would affect those is switching to always use stx::optional.

nightlark:
@phlptp Okay, so here is what I've found for the slower tests.

Using std::experimental::optional or std::optional, the application api tests run in around 295s (occasional timeouts); except it seems like the macOS tests on at least Mojave/10.14 are slow (440s/timeout).
Using stx::optional with the default build type, the application api tests take 430s+ to finish for all compilers.
Using stx::optional with a Release build (Coverage builds or sanitizer builds will most likely timeout or exceed the build time), the application api tests take ~295s (similar to std library) for most Linux builds, except GCC 4.9 which still takes 430s to run the tests. The macOS tests seem to still be slow on Mojave.

Without anything in the cache, Linux builds (compilation time for only HELICS) with the std library or unoptimized stx take ~400s. With release builds using stx, the no cache build time increases to 600-700s. macOS builds times are always slow and can have much worse build times (1700s is the highest I've seen).

nightlark:
For future reference, the macOS build failures (see commits trying several different versions of XCode) were not due to the macOS or XCode version -- they are most likely because the copy of ZeroMQ installed on Homebrew (likely the preferred way for many to install dependencies) is built using tweetnacl; see https://github.com/zeromq/libzmq/issues/2991 for additional information. tl;dr either disable CURVE (cryptography) when building ZMQ, or use libsodium instead (tweetnacl is not intended for use in production).

.ci/Brewfile.travis
```diff
@@ -0,0 +1,6 @@
+brew "pcre"
+brew "ccache"
+brew "cmake"
+brew "swig"
+brew "zeromq"
+brew "boost@1.69"
```
phlptp: Are we still testing with 1.61 and 1.58 on some of the builds?

scripts/run-ci-tests.sh
```diff
@@ -77,11 +77,11 @@ else
                 ;;
             *continuous*)
                 TEST_CONFIG="Continuous"
-		CTEST_OPTIONS+=" --timeout 360"
+		CTEST_OPTIONS+=" --timeout 600"
```
phlptp: We should try to get this down in the future

.ci/Brewfile.travis
nightlark: GCC 4.9 is using 1.61; it looks like 1.58 was used on the old XCode. Should we switch Clang 3.6 to Boost 1.58?

tests/helics/application_api/application-api-tests.cpp
```diff
@@ -23,6 +29,15 @@ struct globalTestConfig
...
     {
+        // macOS ZeroMQ can have issues shutting down, if not built from source without CURVE
+#ifdef ENABLE_ZMQ_CORE
+#ifdef __APPLE__
+        if (ZmqContextManager::setContextToLeakOnDelete ())
+        {
+            ZmqContextManager::getContext ().close ();
+        }
+#endif
+#endif
```
phlptp: Just a note that  I am pretty sure this has to do with the fact that Travis is two cores, and apple does some different things with library terminations, so If the context is terminated at the user level, then the library is deleted that never gets propagated to the zmq threads and they try to shut down after the master lock has been deleted resulting in some chaos during the termination sequence.

scripts/run-ci-tests.sh
nightlark: This also needs to be large enough for the -O1 sanitizer runs. It will likely depend on https://github.com/GMLC-TDC/HELICS-src/issues/583 -- figuring out what tests are redundant, maybe testing the functionality of individual cores instead of using all of them for complete system tests.

.ci/Brewfile.travis
phlptp: It might be good to have at least one of the builds (daily or otherwise) use it since we do claim to support it.

nightlark: I'm switching the Clang 3.6 build to use 1.58 -- I don't know how accurate it is, but I feel like usually if a system has Clang installed, it is more up to date than GCC.

src/helics/common/AsioContextManager.h
```diff
@@ -27,7 +27,11 @@ SPDX-License-Identifier: BSD-3-Clause
 #include <string>
 #include <vector>
 
+#define ASIO_STANDALONE 1
+#define ASIO_ERROR_CATEGORY_NOEXCEPT noexcept(true)
```
phlptp: do these only apply to the context manager, or all of ASIO?  If all of asio should we make the defines in CMAKE as part of the project, rather than defining them wherever we include ASIO?

nightlark: I’m guessing all of Asio. I’ll try with the noexcept one in CMake which already had ASIO_STANDALONE. I wasn’t sure it was actually getting defined correctly since for some reason it was trying to use BOOST_NOEXCEPT even though ASIO_STANDALONE disables using Boost.

nightlark: @phlptp It's just the context manager that needs it. I guess there's something off about the way io_context.hpp includes the ASIO header for setting up definitions correctly.

nightlark:
Alright, the Boost 1.58 is working. If we want to make more changes, we can do them as part of moving to using Docker and/or running some of the macOS builds on Azure.

Issue #671, opened 2019-05-08
Update `Hello World` documentation


nightlark:
**Describe the bug**
The [Hello World](https://helics.readthedocs.io/en/latest/introduction/hello-world.html) tutorial in the documentation needs updating to reflect the final 2.0 API -- including updating the links to example source files.

PR #673, opened 2019-05-12
Fix Hello World tutorial, broken Pub/Sub vs Endpoint table, and miscellaneous wording/abbreviation tweaks

nightlark:
### Description

This fixes the Hello World tutorial (issue #671), a broken [pub/sub vs endpoint table](https://helics.readthedocs.io/en/latest/introduction/pubsub-vs-message.html), and makes miscellaneous wording fixes and ensures more consistent use of  the more common written names for programming languages.

### Proposed changes
- Fix the Hello World tutorial source code links and function names to reflect the final HELICS 2.0 API
- Fix the pub/sub vs endpoint comparison table
- Misc word/sentence fixes


nightlark:
There is a RTD build with these changes here: https://helics.readthedocs.io/en/doc-fixes/

@kdheepak It looks like there are some shortcomings to using markdown for all of the pages -- in particular it seems like tables are unsupported, so it looks like pages with tables may need to use rst.

kdheepak:
We can use eval_rst code blocks to include rst. But I'm fine converting the
documentation to rst entirely if you think that would be better.

nightlark:
@kdheepak Ah thanks for the eval_rst tip -- I just switched it to use that and it looks right. I guess that's not a standard feature since GitHub parses rst/md files, but doesn't recognize eval_rst blocks in markdown.

phlptp:
Maybe fix the codacy issue, but otherwise looks good.

Issue #675, opened 2019-05-12
Auto download of submodules


phlptp:
**Describe the bug**
If a user doesn't do a recursive checkout or init the submodules, the compilation will fail.  

**What is the expected behavior?**
The git submodule checkout should be built into the cmake code to automatically fetch them if it isn't done already.

phlptp:
@nightlark  think we can close this now?

PR #678, opened 2019-05-13
Autoupdate git submodules

nightlark:
### Description
If merged this pull request will add an option to the CMake configure step to ensure that the git submodules are initialized and updated.

### Proposed changes
- Add an update git submodules script to the CMake build process to automatically download git submodules (issue #675)


nightlark:
Skipping submodules might only be possible with newer versions of git (one place mentioned 2.13 for an argument that specifies which ones to update). I'm not sure about the minimum version requirements for this method: https://stackoverflow.com/a/52185169

One of the ways suggested by GoogleTest is using ExternalProject_Add like we do with some of the autobuild options: https://github.com/google/googletest/tree/master/googletest#incorporating-into-an-existing-cmake-project

Issue #665, opened 2019-05-05
Convert some tests to Google Test


phlptp:
Wonder what peoples thoughts are on including google test in the project.  I think we would do it as a submodule.  And move a few tests to use GTEST,  possibly migrate all of them over a longer time period, some of our use of labels and global fixtures might not be easily amenable to google test framework,  but it would give some level of testing at all levels of boost, and give us the possibility of taking advantage of some of the things google test has the boost does not.

nightlark:
Including googletest seems like a good idea, and would make it possible to have a CI build running tests with our oldest supported version of Boost (partly as a result of some of the Boost.Test features we are using). With some changes to the test names and grouping into test suites, we could use filters to do the same thing as labels.

PR #680, opened 2019-05-14
Googletest add

phlptp:
### Description

If merged this pull request will add google test and convert the core-tests to use google test instead of boost tests.  

### Proposed changes

- integrate google test
- convert core-tests to use google test
- remove a few warning notices from the interface builds


phlptp:
Close issue #665

Issue #655, opened 2019-04-05
helicsInputIsUpdated is reset when sending as a double and receiving as a string


kdheepak:

phlptp:
@kdheepak  can you test this again with the develop branch.  I think I fixed it but I want to make sure it got fixed with what you doing.

kdheepak:
I will check and get back to you.

phlptp:
@kdheepak  any update?

phlptp:
@kdheepak  have you had a chance to test the fixes yet?

kdheepak:
I have not unfortunately, I'll get to it this afternoon.

phlptp:
@kdheepak have you tested this yet?

kdheepak:
I've tested this on the latest develop and this is resolved.

PR #683, opened 2019-05-17
Cmake version updates

phlptp:
### Description
Fix

If merged this pull request will fix Issue #655 
change the way some the version flags are detected and add helicsCLI11.hpp to the installed folder

### Proposed changes

- delete USE_CXX_17 in favor of a more standard CMAKE_CXX_STANDARD and match the detection to fmt named flags.  This was a bit better than trying to match version numbers of compilers and removes a duplicate check in fmt.


nightlark:
Is it issue #655 that this fixes?

phlptp:
yes I corrected it,  Thanks!

PR #687, opened 2019-05-23
Add sstream header to JsonProcessingFunctions

nightlark:
### Description

If merged this pull request will include the sstream header to JsonProcessingFunctions.cpp to fix a compile error encountered on MSVC.

### Proposed changes
- Add sstream header to JsonProcessingFunctions.cpp to fix compile error
- Remove JsonCpp include from a file that doesn't use it

### DIFF
--- a/src/helics/common/JsonProcessingFunctions.cpp
+++ b/src/helics/common/JsonProcessingFunctions.cpp
@@ -7,6 +7,7 @@ SPDX-License-Identifier: BSD-3-Clause
 
 #include "JsonProcessingFunctions.hpp"
 #include <fstream>
+#include <sstream>
 
 bool hasJsonExtension (const std::string &jsonString)
 {
--- a/src/helics/application_api/queryFunctions.cpp
+++ b/src/helics/application_api/queryFunctions.cpp
@@ -10,8 +10,6 @@ SPDX-License-Identifier: BSD-3-Clause
 #include "Federate.hpp"
 #include <thread>
 
-#include "json/json.h"
-
 #include <boost/algorithm/string.hpp>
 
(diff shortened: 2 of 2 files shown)


src/helics/common/JsonProcessingFunctions.cpp
```diff
@@ -7,6 +7,7 @@ SPDX-License-Identifier: BSD-3-Clause
 
 #include "JsonProcessingFunctions.hpp"
 #include <fstream>
+#include <sstream>
```
phlptp: IF this is needed how was it working before?

nightlark: I'd guess it was getting included somewhere by json/json.h, but I don't know. It might be a problem when using the jsoncpp target in a Debug build.

nightlark: It's possible this is just a bandaid, and that the cause of the problem is in the json library -- will need to try a few more things and see; fixing it might be something for this weekend.

phlptp: Ahh,  this came up when you got rid of the other header?  I see now.

phlptp:
Is the appveyor failure something important or just one of the random failures?

nightlark:
I've never seen that appveyor failure before.

phlptp:
@nightlark  can you restart it and see if it passes,  looks like it may have been cancelled by an appveyor issues or something like that

phlptp:
@nightlark  can this be merged?

nightlark:
I guess it is okay to merge for now. If I figure out what about the build environment or which flag was causing the error and come up with a better fix I'll open another PR later.

PR #695, opened 2019-06-01
Remove map traits

phlptp:
### Description

If merged this pull request will remove references to the MapTraits header from the public API.  and move the functionality that required it to a separate header that is included only when needed.  

### Proposed changes

- add an extra header with the map-traits linkages for internal use
- add is_easily_hashable trait for interface_handle objects


nightlark:
It looks like common/logger.h is also using something from the containers library -- do we just want projects using that header to provide their own copy of the required gmlc::containers library?

phlptp:
I will take a look at the logger and see what we want to do with that one.   I think it was put in there since it is useful for some other projects but isn't necessarily considered part of the API.    It is possible we don't want to include it any more and at some point in the future consider making it a separate repo like the containers or possibly include it with the containers repo since it uses it pretty directly.

phlptp:
@nightlark I split the logger into two pieces.  The logger which is more useful as a public header since federates could use it.  and the loggingCore which isn't public and contains the links to the containers library and the tripWire which also isn't public.    Take a look and make sure I didn't miss anything.

PR #696, opened 2019-06-01
Use gmlc prefixed container library headers

nightlark:
### Description

If merged this pull request will include gmlc/ prefixed gmlc::container library headers

### Proposed changes
- Add gmlc prefix to container library headers
- Update GMLC containers library git submodule


nightlark:
The update to the GMLC containers git submodule needs to wait until after https://github.com/GMLC-TDC/containers/pull/2 is merged.

PR #704, opened 2019-06-09
Message timer fixes

phlptp:
### Description

If merged this pull request will fix a recurring sporadic issue with the message Timer and some potential ZMQ issues if ZMQ happened to be installed on Windows without the headers.  

### Proposed changes

- reorder some operations in the message timer to eliminate a potential pathway for stack corruption
- deal with the case of an expired timer starting out.  
- add an additional test case to specifically test the expired timer pathway.  
- use the chrono literals to clean up code.

### DIFF
--- a/config/cmake/FindZeroMQ.cmake
+++ b/config/cmake/FindZeroMQ.cmake
@@ -164,13 +164,25 @@ find_library(
     PATHS /lib /usr/lib /usr/local/lib
 )
 
+if (ZeroMQ_REQUIRE_HEADERS)
+if(ZeroMQ_INCLUDE_DIR)
 if(
     (ZeroMQ_LIBRARY AND NOT ZeroMQ_LIBRARY-NOTFOUND)
     OR (ZeroMQ_STATIC_LIBRARY AND NOT ZeroMQ_STATIC_LIBRARY-NOTFOUND)
 )
     set(ZeroMQ_FOUND 1)
 endif()
+endif()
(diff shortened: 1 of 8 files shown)


phlptp:
This seems to be passing appveyor consistently now, though we may want to run a number more just to be sure.  I think i might have figured it out, though I can't be totally sure.  I think it had to do with the the timer and locks used.  If the OS switched contexts at the wrong time, the time potentially could have expired.  Given the setup this is almost impossible to occur if you have 4 or more cores, but with two it might occur every once in a while, actually a little more frequently than before due to the way google test uses threads.  Anyway if the timer had expired there was some possibility it would trigger the callback before everything was completely setup which was an actual bug that was fixed in the first couple rounds of fixes, but that left the message not being triggered if it had already expired so this required a new pathway to check for expired times.    So I am not fully confident it is fixed given the sporadic nature of it, but I at least have a story of why it was happening and why it could be fixed now.

phlptp:
@nightlark could you trigger another 10,  if those don't fail from the message time test we will call it good.

phlptp:
I am going to call this "better",   but the failure rate statistically seems lower 1/22 vs ~2/10.  And the segmentation fault/stack corruption that cropped up when the test failed has been resolved.  There may be further issues with the contextManager, but that will probably require some additional tests to even start to get a handle on and would not be specific to the messageTimers.  So Unless the current build fails or someone notices something in the review.   I will not be adding additional code.

PR #708, opened 2019-06-14
Trigger helics-ns3 repo builds with PR variables

nightlark:
### Description
If merged this pull request will pass variables needed for testing a HELICS PR to the helics-ns3 repo builds.

### Proposed changes
- Pass variables needed for testing a PR merge to the helics-ns3 Azure builds
- Trigger helics-ns3 builds for commit/PR to master

### DIFF
--- a/scripts/trigger-dependent-ci-builds.sh
+++ b/scripts/trigger-dependent-ci-builds.sh
@@ -19,23 +19,9 @@ curl -s -X POST \
     -d "$body" \
     https://api.travis-ci.org/repo/GMLC-TDC%2FHELICS-FMI/requests
 
-# Trigger helics-ns3 module build
-body='{
-"definition": {
-"id": 1
-},
-"reason": "individualCI"
-}'
-
-curl -s -X POST \
-    -H "Content-Type: application/json" \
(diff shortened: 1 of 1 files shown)


nightlark:
Most of the discussion I've found online for triggering builds in dependent repositories uses API requests, so we aren't the only ones doing things this way.

Travis definitely doesn't support anything related to triggering builds in other repositories, and everything I've found online uses API requests in some way. Same with GitLab.

Azure has some support for triggering pipelines under the same project, but I haven't found anything about passing custom variables from the triggering pipeline (only build artifacts) which we'd need for testing a PR correctly. Our mix of services (and currently separate projects for the different repositories) also means that we can't use that option. I think Atlassian Bamboo is similar, with a limited way of triggering build plans when another finishes.

Jenkins or BuildBot instances seem to be more common for larger projects, so they might have some features for triggering builds -- for Jenkins it looked like the complexity of triggering a reverse build is similar to using curl for a post request.

phlptp:
Ok,  lets just keep an eye on it for structural changes, since I imagine there will be a few more, like the demos that might need this too, and I want to make sure it is maintainable.

PR #709, opened 2019-06-14
Get rid of GCC 4.9.3 missing-field-initializers warning

nightlark:
### Description
If merged this pull request will remove the empty field initializer list. For some reason gcc 4.9.3 has issues with it and prints warnings.

### Proposed changes
- Remove the empty field initializer list when creating a helics_message object in the shared API library

### DIFF
--- a/src/helics/shared_api_library/MessageFederateExport.cpp
+++ b/src/helics/shared_api_library/MessageFederateExport.cpp
@@ -425,7 +425,7 @@ helics_message helicsEndpointGetMessage (helics_endpoint endpoint)
 
     if (endObj->lastMessage)
     {
-        helics_message mess{};
+        helics_message mess;
         mess.data = endObj->lastMessage->data.data ();
         mess.dest = endObj->lastMessage->dest.c_str ();
         mess.length = endObj->lastMessage->data.size ();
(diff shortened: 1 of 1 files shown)


nightlark:
With this change, the HELICS library (not including warnings in Boost code) compiles without warnings (using -Wall -Wextra -Wpedantic) using gcc 4.9.3. Also tested and there were no warnings using gcc 8.1.0 and clang 6.0.0 on LC.

Also tried building with Clang's -Weverything which gives many warnings, mostly about the use of non-C++98 compatible features and weak vtables.

src/helics/shared_api_library/MessageFederateExport.cpp
```diff
@@ -425,7 +425,7 @@ helics_message helicsEndpointGetMessage (helics_endpoint endpoint)
 
     if (endObj->lastMessage)
     {
-        helics_message mess{};
+        helics_message mess;
```
phlptp: I thought there was some reason this was an issue that the {} was required, but it may have been resolved in some of the HELICS 2.0 changes.    And I can't remember what it was right now.

nightlark: Would it be related to the default values for fields in the struct? The emptyMessage function above this one didn't use {}, but it seems to set all the fields -- would changing this to `helics_message mess = emptyMessage()` be something we would want to do?

phlptp: No, the intention was to not do anything with the fields that were just going to be filled in later.  There were times when the emptyMessage was needed, and other times when we just wanted space for the struct and the needed fields would be filled in later, so having a default initializer was redundent, and since the struct needed to work in a C context we couldn't make it behave like a class with the constructors and things.

PR #710, opened 2019-06-16
Add Drone Cloud config file for ARM and ARM64 CI builds

nightlark:
### Description
If merged this pull request will add a config file for ARM (32-bit) and ARM64 builds using [Drone Cloud](https://cloud.drone.io/).

### Proposed changes
- Add a .drone.yml file to the .ci folder for [Drone Cloud](https://cloud.drone.io/) on ARM and ARM64 platforms.


phlptp:
Does this result go into the PR checks?  And can a shield be added to the main readme? 
Why are only 2 of the PR checks showing up for this PR?

phlptp:
Ok I see the reason the Push check isn't there.    Do we not do appveyor checks on PR's?    Is that something that can be setup?  

Also when does the drone.ci get run/ on which branches?

nightlark:
It looks like the push checks aren't there because this PR is from a different repository. I don't know if it is - there's this 2014 article about triggering Appveyor builds for PRs that may not be relevant anymore, but I can't access that part of the settings: https://help.appveyor.com/discussions/questions/203-auto-run-tests-on-pull-requests#comment_33212404

 There are security implications if any secure env vars are used when building PRs originating from forks; I turned on the Appveyor setting to not use secure env vars in PRs originating from forks.

nightlark:
I think Drone will run for all branches -- someone else will need to go to the website and enable GMLC-TDC/HELICS-src builds. It looks like managing webhooks and services requires admin for a repository.

phlptp:
@nightlark can you post a link to one of the builds.    What do I need to do on the repo to enable it?

nightlark:
Alright, so results can appear in the list of checks passed. A shield could be added to the readme -- I'd probably want to make sure the ARM 32-bit result gets ignored or skips running tests for status reporting.

Enabling should be logging into Drone Cloud and activating it for GMLC-TDC/HELICS-src; from there I think I can change settings (Main->Configuration needs to be set to point to `.ci/.drone.yml`).

The most recent build log is at https://cloud.drone.io/nightlark/HELICS-src/45/2/1 which was run with the ghc::filesystem fix, and an older version of fmtlib (I think the upstream version also needs a fix). The Drone config file in this PR is updating to the latest upstream version of the git submodules; I don't know if we'd want it to do that (pro: catching breaking changes upstream early, con: "random" build failures).

phlptp:
I think I would prefer it use the current version of the git submodules,  those will be updated periodically but I don't want to have to debug any changes in upstream submodules when we are potentially trying to evaluate current changes in the main HELICS repo.  Those should be indepdendent of eachother.

phlptp:
I think I added it correctly.  You will have to double check once merged, then you can add the shields for master and develop like the others.

nightlark:
It looks like I can't change the "main" settings on Drone, only the secret env var section and below.

phlptp:
Did you need me to change something?

nightlark:
Main -> Configuration needs to be set to .ci/.drone.yml

PR #711, opened 2019-06-18
Update ghc::filesystem to include systemerror_r fix

nightlark:
### Description

If merged this pull request will update the included copy of ghc::filesystem to include a fix that better supports different compiler implementations of systemerror_r (GNU returns a char*, other compilers return int; preprocessor conditional directive didn't choose correctly for all compilers).

### Proposed changes
- Update ghc::filesystem to the latest version


phlptp:
Is this a version update on the filesystem header?  Something you ran into?

nightlark:
Yes, it's a version update -- building with musl instead of glibc was falling through to the `#else` block for systemerror_r return `char*`, but musl is POSIX compliant with systemerror_r returning `int`.

PR #713, opened 2019-06-20
Set dependent repo build branch to HELICS branch

nightlark:
### Description
If merged this pull request will use the HELICS branch that was just built as the source branch to use in the dependent repository build. This should help make the result reported for builds to master/develop a bit more meaningful (Examples master branch needs updating for the release) and help ensure the master branches in different repositories are compatible.

### Proposed changes
- Test the branch in the dependent repository whose name matches
- Refactor the curl invocation into a bash function

### DIFF
--- a/scripts/trigger-dependent-ci-builds.sh
+++ b/scripts/trigger-dependent-ci-builds.sh
@@ -3,21 +3,21 @@
 BUILD_MESSAGE="Test ${TRAVIS_REPO_SLUG} ${TRAVIS_COMMIT_RANGE}"
 
 # Trigger HELICS-FMI build
-body='{
-"request": {
-"message":"'
-body+="${BUILD_MESSAGE}"
-body+='",
-"branch":"master"
-}}'
+#body='{
+#"request": {
+#"message":"'
+#body+="${BUILD_MESSAGE}"
+#body+='",
+#"branch":"master"
+#}}'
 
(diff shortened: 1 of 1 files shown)


scripts/trigger-dependent-ci-builds.sh
```diff
@@ -3,21 +3,49 @@
...
 
 # Trigger HELICS-FMI build
-body='{
-"request": {
-"message":"'
-body+="${BUILD_MESSAGE}"
-body+='",
-"branch":"master"
-}}'
+#body='{
```
phlptp: are we not triggering the FMI build any more?

nightlark: It looks like at some point the API token stopped working. I thought I'd seen something about Travis API tokens expiring every few months, but it may also be GH permissions for the HELICS-bot account. I'm thinking of switching HELICS-FMI builds over to Azure so that most of the setup for the dependent repositories is the same -- and the Azure API rate limit for queuing builds is more generous (Travis API only allows 3 queued builds/hour).

phlptp: that sounds reasonable,  we probably should still have some of the builds of HELICS-FMI on travis though.  But the triggered builds seems like a good idea to make consistent across all the repos.

PR #717, opened 2019-06-22
Jsoncpp objlib

phlptp:
### Description

If merged this pull request will enable a jsoncpp cmake object library to be used instead of a standalone static library.  

### Proposed changes

- update the version of jsoncpp used
- create advanced options tor linking with the jsoncpp object library


nightlark:
I'm going to try a build+install with this branch to make sure headers aren't getting  installed outside the HELICS include folder.

CMakeLists.txt
```diff
@@ -543,13 +543,22 @@ include(addfmt)
 #  Create the target for jsoncpp
 # -----------------------------------------------------------
 include(addJsoncpp)
+
+if (JSONCPP_OBJLIB)
+target_include_directories(
+    helics_base_includes SYSTEM
+    INTERFACE
+    $<TARGET_PROPERTY:jsoncpp_headers,INTERFACE_INCLUDE_DIRECTORIES>
```
nightlark: I think we should have the headers be opt-in for projects using HELICS.
```suggestion
    $<BUILD_INTERFACE:$<TARGET_PROPERTY:jsoncpp_headers,INTERFACE_INCLUDE_DIRECTORIES>>
```

CMakeLists.txt
```diff
@@ -543,13 +543,22 @@ include(addfmt)
...
+)
+set_target_properties(jsoncppobj_lib PROPERTIES FOLDER Extern)
+else()
 target_include_directories(
     helics_base_includes SYSTEM
     INTERFACE
     $<TARGET_PROPERTY:jsoncpp_lib,INTERFACE_INCLUDE_DIRECTORIES>
```
```suggestion
    $<BUILD_INTERFACE:$<TARGET_PROPERTY:jsoncpp_lib,INTERFACE_INCLUDE_DIRECTORIES>>
```

CMakeLists.txt
phlptp: One issue with that is now we are using json more and more closer to the interface.  HELICS-FMI uses it, and a lot of the queries are defined in JSON.  I think it makes sense to have an included json interpreter when linking with HELICS in C++.    And I may swap GridDyn to use the HELICS version of JSON when linking with HELICS.  So I am not sure I want the headers not to be installed by default.  Without them being installed it would make it impossible to use the JSON components embedded in the HELICS library.

nightlark: They would still be installed by default, just not added to the include path unless the project links with HELICS::jsoncpp_headers. Having the headers tied to the helics-static target could be annoying with simulators that already use their own copy of jsoncpp.

phlptp: so this would only apply if jsoncpp was included via object library?  I would imagine if they were left separate as is default on MSVC they still get linked with the regular static library?    If so that makes sense then.

nightlark: It could be done so that it only applies if jsoncpp is included via object library. The way things are now for jsoncpp used as an object library should be a safe default that will avoid dependency conflicts or overwriting a system install of jsoncpp, and should work for adding to package repositories.

For the separate static jsoncpp library case (MSVC), linking to helics-static should link with the static jsoncpp library, but headers won't be included by default. I think there are two options:
1. line 558 could get reverted to `$<TARGET_PROPERTY:jsoncpp_lib,INTERFACE_INCLUDE_DIRECTORIES>` and then the json headers will always be included when linking to helics-static, but may require checking for HELICS::jsoncpp_lib (or in the requirements for HELICS-FMI say that it requires building HELICS with JSONCPP_OBJLIB set to OFF).
2. modify our custom copy of jsoncpp to always create and export a jsoncpp_headers interface library; then users can always guarantee that a HELICS::jsoncpp_headers target will exist

I was leaning towards 2, but 1 might be better because it sets a single configuration for simulators wanting to use the HELICS copy of jsoncpp (less options that need testing).

phlptp: for 2, would you think, just alias jsoncpp_headers to jsoncpp_lib?    The jsoncpp static lib target should contain the included headers.    They might not get propagated if you just link with the helicslib.   But if you link the jsoncpp target you should get the appropriate include folder.

nightlark: Aliasing jsoncpp_headers to jsoncpp_lib would probably work to get the right include folder.

PR #721, opened 2019-06-25
HELICS add_subproject test

nightlark:
### Description
If merged this pull request will add a test for using HELICS as a subproject.

### Proposed changes
- Adds a test for using HELICS with add_subproject
- Runs the test on Travis gcc 4.9 packaging build (~6 minute test time with no cache, overall build time is slightly less than other builds running tests; subsequent builds for PR/push to the same branch is much faster)

### DIFF
--- a/tests/helics/add_subdirectory_tests/main-apps.cpp
+++ b/tests/helics/add_subdirectory_tests/main-apps.cpp
@@ -1,0 +1,14 @@
+/*
+Copyright © 2019,
+Battelle Memorial Institute; Lawrence Livermore National Security, LLC; Alliance for Sustainable Energy, LLC.  See
+the top-level NOTICE for additional details. All rights reserved.
+SPDX-License-Identifier: BSD-3-Clause
+*/
+
+#include "helics/helics_apps.hpp"
+
+int main (int argc, char *argv[])
+{
+    volatile helics::apps::BrokerApp broker("--help");
+    return 0;
+}
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -790,9 +790,9 @@ if(ENABLE_PACKAGE_BUILD)
         cereal
     )
     if(WIN32)
-        set(CPACK_RESOURCE_FILE_LICENSE "${PROJECT_SOURCE_DIR}\\\\LICENSE")
+        set(CPACK_RESOURCE_FILE_LICENSE "${HELICS_SOURCE_DIR}\\\\LICENSE")
     else(WIN32)
-        set(CPACK_RESOURCE_FILE_LICENSE "${PROJECT_SOURCE_DIR}/LICENSE")
+        set(CPACK_RESOURCE_FILE_LICENSE "${HELICS_SOURCE_DIR}/LICENSE")
     endif(WIN32)
 
     set(CPACK_COMPONENT_APPLICATIONS_DISPLAY_NAME "Application")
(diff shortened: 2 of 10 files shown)


nightlark:
The appveyor builds will only run the find_package test; the build with add_subdirectory on appveyor is very slow, and there's no cache to help it out on subsequent runs.

phlptp:
why are the mappedVectorTests getting added back in again?  There seem to be a few other old test files that shouldn't be there?

nightlark:
> why are the mappedVectorTests getting added back in again? There seem to be a few other old test files that shouldn't be there?

Good catch -- it looks like when I merged with develop `git add` from resolving a conflict added them back in.

nightlark:
@phlptp I think I got it rebased with files in the correct state now.

tests/helics/add_subdirectory_tests/main-apps.cpp
```diff
@@ -0,0 +1,14 @@
+/*
```
phlptp: Is there any way we can have these files be the same as in the other tests?  I was working on adding the package_config_tests as well which would use the same files again, just wondering if we could use the same files over again instead of adding more copies,  or are they a little different?

nightlark: The CMakeLists.txt files are different, but the source files are the same.

phlptp: Lets work on unifying them in the next version when we merge in the package_config tests which will also be similar.  For now this looks fine.  Though I am wondering if these tests belong in here or in a different repository.    Is there a way we can make sure they don't get run if someone does make test?

nightlark: The way I know of would be only enabling them with an option that defaults to off, maybe the cpack enable option. I need to see if I can find a list of ctest properties for one that disables a test by default.

I think the find_package tests should be here as a quick test to catch some possible issues with HELICSConfig and the export targets to reduce the chance of triggering builds in a lot of dependent repositories that will all fail.

nightlark:
Reference for properties that can be set on tests: https://cmake.org/cmake/help/v3.15/manual/cmake-properties.7.html#test-properties

tests/helics/add_subdirectory_tests/main-apps.cpp
nightlark: I set the first slower add_subdirectory test to disabled, and the find_package test to enabled since it only takes a few seconds to run.

phlptp: Is the add_subdirectory test still running on travis?

nightlark: Yes, it still runs on Travis

nightlark: But the reason it still runs is because the CMake version used for that Travis build doesn't support the DISABLED property. :neutral_face:

phlptp: that is unfortunate, sounds like that might be the best option to add a specific advanced flag to enable that test

nightlark: Yea. The DISABLED property doesn’t make much sense either — I haven’t found any arguments to ctest that will run a disabled test. The way I can see to use it is setting the disabled flag if eg the python interface is disabled, so that the tests not run still show up in the summary after running ctest.

Issue #389, opened 2018-07-20
Set version of shared library on Mac


kdheepak:
Version number of the `libhelicsSharedLib.dylib` is set to 0.0.0 by default. We can set it to the corresponding version of HELICS.

```
~/local/helics-develop/lib $ otool -L libhelicsSharedLib.dylib                                                                                                                                             
libhelicsSharedLib.dylib:
        @rpath/libhelicsSharedLib.dylib (compatibility version 0.0.0, current version 0.0.0)
        /usr/local/opt/boost/lib/libboost_program_options.dylib (compatibility version 0.0.0, current version 0.0.0)
        /usr/local/opt/boost/lib/libboost_filesystem.dylib (compatibility version 0.0.0, current version 0.0.0)
        /usr/local/opt/boost/lib/libboost_system.dylib (compatibility version 0.0.0, current version 0.0.0)
        /usr/local/opt/boost/lib/libboost_date_time.dylib (compatibility version 0.0.0, current version 0.0.0)
        /usr/local/opt/boost/lib/libboost_timer.dylib (compatibility version 0.0.0, current version 0.0.0)
        /usr/local/opt/boost/lib/libboost_chrono.dylib (compatibility version 0.0.0, current version 0.0.0)
        /usr/local/opt/zeromq/lib/libzmq.5.dylib (compatibility version 7.0.0, current version 7.5.0)
        /usr/lib/libc++.1.dylib (compatibility version 1.0.0, current version 307.5.0)
        /usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1238.60.2)
```

PR #723, opened 2019-06-26
Set SO version for HELICS shared libraries

nightlark:
### Description
If merged this pull request will set the SO version for the HELICS shared libraries.

### Proposed changes
- Set SO version for the HELICS shared libraries.

### DIFF
--- a/src/helics/CMakeLists.txt
+++ b/src/helics/CMakeLists.txt
@@ -44,6 +44,10 @@ if(BUILD_SHARED_LIBS)
       $<INSTALL_INTERFACE:$<INSTALL_PREFIX>/include>
   )
 
+  set_target_properties(helics-shared PROPERTIES
+    VERSION ${HELICS_VERSION}
+    SOVERSION ${HELICS_VERSION_MAJOR}
+  )
 
   if(WIN32)
     set_target_properties(helics-shared PROPERTIES WINDOWS_EXPORT_ALL_SYMBOLS TRUE)
--- a/src/helics/shared_api_library/CMakeLists.txt
+++ b/src/helics/shared_api_library/CMakeLists.txt
@@ -42,6 +42,11 @@ target_include_directories(
         $<INSTALL_INTERFACE:$<INSTALL_PREFIX>/include>
 )
 
+set_target_properties(helicsSharedLib PROPERTIES
+    VERSION ${HELICS_VERSION}
+    SOVERSION ${HELICS_VERSION_MAJOR}
+)
+
 set_target_properties(
   helicsSharedLib
   PROPERTIES


kdheepak:
This is great. I was also going to suggest something similar after reading chapters regarding shared libraries in "The Linux Programming Interface" (Chapter 41 and 42). 

The couple of other things that they recommended were:

1) Always use -fPIC
2) Use $ORIGIN in the rpath to bind to dependencies. This way you can create applications that can be moved after install without any issues.

nightlark:
For the rpath, the packaging guidelines for different distros seem to not always recommend the same thing.

Another tricky bit is ABI compatibility of the helics-shared and helics-static libraries for different compilers — I’m not sure how well including those in a package will work; at least not a universal Linux package.

nightlark:
I’m also trying to work out how interfaces for other languages fit into packages — it seems like most install to the system python dist-packages subfolder.

nightlark:
I don't have a mac to test on, but I think this might also address issue #389

src/helics/CMakeLists.txt
```diff
@@ -50,6 +50,11 @@ if(BUILD_SHARED_LIBS)
       $<INSTALL_INTERFACE:$<INSTALL_PREFIX>/include>
   )
 
+  set_target_properties(helics-shared PROPERTIES
+    VERSION ${HELICS_VERSION}
+    SOVERSION ${HELICS_VERSION_MAJOR}
```
phlptp: does the SOVERSION not need to change for minor version updates?  Or is this actually about backwards compatibility

nightlark: It’s backwards compatibility. The SONAME is set in all of the shared libraries to sharedLib.so.<SOVERSION> and is what a program actually links to, so a new major version could be installed side-by-side the old one without breaking programs that are linked to the older version.

nightlark:
@kdheepak can you check that the file names look okay on macOS? I think I'm going to merge this soon, since the tests pass on the Travis mac build there shouldn't be any breaking changes.

kdheepak:
I think if we install from source we should set rpath. When installing from source in a package manager's formula, we should follow guidelines.

kdheepak:
That appears to have solved the version in the shared library.

```
╭─ ~/GitRepos/HELICS-src/build-osx (soversion)
╰─ $ otool -L src/helics/shared_api_library/libhelicsSharedLib.dylib
src/helics/shared_api_library/libhelicsSharedLib.dylib:
        @rpath/libhelicsSharedLib.2.dylib (compatibility version 2.0.0, current version 2.1.0)
        /usr/local/opt/zeromq/lib/libzmq.5.dylib (compatibility version 8.0.0, current version 8.1.0)
        /usr/lib/libc++.1.dylib (compatibility version 1.0.0, current version 307.5.0)
        /usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1238.60.2)
```

nightlark:
Is setting the rpath needed to fix something that broke as a result of this change, or is it a separate issue?

If it's not something that was getting done before, then I'd rather lump setting a $ORIGIN rpath in with a change that fixes an issue with installing HELICS libraries in the bin folder on Linux, since that looks like it could involve additional rpath changes.

kdheepak:
I agree, it's not required in this PR.

PR #728, opened 2019-06-27
Remove HELICS shared library install to bin

nightlark:
### Description
If merged this pull request will stop HELICS from installing a shared library to the bin directory on systems other than Windows.

### Proposed changes
- Remove HELICS shared library install to bin


nightlark:
Closing this for now -- there are some other platform differences that need to get worked out. Easiest solution for now is to just delete the shared library from the bin folder during packaging.

PR #733, opened 2019-07-06
Test issue fixes

phlptp:
### Summary

If merged this pull request will fix a series of issues from undefined behavior sanitizer, thread sanitizer,  valgrind, and klockwork, and some occasionally failing test cases.  

### Proposed changes

- fix a vptr race with the ZMQSSComms
- fix a race condition with transfer of interfaceInfo to the CommonCore
- fix undefined behavior with the actionMessage Serialization
- change the threadSanitizer to run on the system tests
- make a blacklist of the zmq unaligned access since that is not part of HELICS
- fix memory leak in json writer usage

### DIFF
--- a/.ci/undef_blacklist.txt
+++ b/.ci/undef_blacklist.txt
@@ -1,1 +1,1 @@
-src:yqueue.hpp
+src:*/yqueue.hpp
--- a/.ci/undef_blacklist.txt
+++ b/.ci/undef_blacklist.txt
@@ -1,1 +1,1 @@
-src:*/yqueue.hpp
+src:*/libzmq/*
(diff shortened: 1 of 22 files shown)


nightlark:
This fixes the invalid memory alignment errors with the 32-bit ARM builds -- looks like it was the actionMessage serialization changes.

phlptp:
I wondered if that was the case with that issue on older ARMs  My understanding is they have slightly different instructions if there is a possible unaligned access which memcpy would use, but not necessarily the reinterpret cast.    Though the only way I can think of that being unaligned if using the TCP cores and messages getting merged.  Then they might actually get off alignment.

Issue #740, opened 2019-07-12
helics_max_time inconsistency


phlptp:
**Describe the bug**
what is the current behavior
the helics_max_time value used in the c/python/... libraries is inconsistent with the maxTime used in the C++ library.  For the common use cases of indicators this works fine, but when used for checking a return value it can lead to inconsistencies.  

**What is the expected behavior?**
 What is the motivation / use case for changing the behavior?
The two values should be consistent with eachother across the interface boundary.

See [Gitter chat](https://gitter.im/GMLC-TDC/HELICS?at=5d28ef6dd14e297762be4cb7)

PR #741, opened 2019-07-13
set the maxTime in C++ and C to be consistent with eachother,   meani…

phlptp:
### Summary

If merged this pull request will fix an inconsistency in the helics_time_maxtime value in the C api

### Proposed changes

- change the conditions in which a time is treated as maxTime,
- change the value of helics_time_maxtime
- add some conditions on the outputs of the C api functions that check max time and make it equal to the helics_time_maxval

fix Issue #740

### DIFF
--- a/src/helics/common/timeRepresentation.hpp
+++ b/src/helics/common/timeRepresentation.hpp
@@ -105,7 +105,8 @@ class integer_time
         auto divBase = static_cast<base> (div);
         double frac = div - static_cast<double> (divBase);
         baseType nseconds = (divBase << N) + static_cast<base> (frac * multiplier);
-        return (t < -1e12) ? ((t > 1e12) ? nseconds : maxVal ()) : minVal ();
+        return (t < toDouble (minVal () + 10000)) ? ((t > toDouble (maxVal () - 10000)) ? nseconds : maxVal ()) :
+                                                    minVal ();
     }
     static CHRONO_CONSTEXPR baseType convert (std::chrono::nanoseconds nsTime) noexcept
     {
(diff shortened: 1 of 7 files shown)


phlptp:
@nightlark  once you take a look at this one, I will merge it then I need to update the interface files for the python/matlab/java included files and will generate it for the json_aggregate output PR, so I only have to do it once.

src/helics/common/timeRepresentation.hpp
```diff
@@ -105,7 +105,8 @@ class integer_time
         auto divBase = static_cast<base> (div);
         double frac = div - static_cast<double> (divBase);
         baseType nseconds = (divBase << N) + static_cast<base> (frac * multiplier);
-        return (t < -1e12) ? ((t > 1e12) ? nseconds : maxVal ()) : minVal ();
+        return (t < toDouble (minVal () + 10000)) ? ((t > toDouble (maxVal () - 10000)) ? nseconds : maxVal ()) :
```
nightlark: Why was 10000 chosen?

src/helics/shared_api_library/api-data.h
```diff
@@ -45,8 +45,8 @@ extern "C"
     const helics_time helics_time_zero = 0.0; /*!< definition of time zero-the beginning of simulation*/
     const helics_time helics_time_epsilon = 1.0e-9; /*!< definition of the minimum time resolution*/
     const helics_time helics_time_invalid = -1.785e39; /*!< definition of an invalid time that has no meaning*/
```
nightlark: Is any adjustment to the helics invalid time also needed?

phlptp: I will look into that

src/helics/common/timeRepresentation.hpp
phlptp: The maxTime in C++ has integral representation of 9223372036854775807,  the double representation of this gets converted back to 9223372036854774784.  This means a difference of 1024  is the lowest that number could be used to make this consistent.  10000 is 10 microseconds at the default time resolution which seems close enough to not matter that much and give a little more wiggle room.  I think it has to be done at the underlying representation level to be consistent as well.  

The other option was the float representation of Time::maxTIme()  but that would have been ~26 seconds worth of time shift,  that seemed to be kind of a waste since I don't think that would come into play that often if ever.

src/helics/shared_api_library/api-data.h
phlptp: I don't believe so, there is no equivalent to that in the C++ API since it would use exceptions where the invalid time would have been used.

src/helics/common/timeRepresentation.hpp
phlptp: I should put a note in the code about why it is there though.

PR #742, opened 2019-07-15
Update helicsCLI11App enum

nightlark:
### Summary
If merged this pull request will update a few instances of the parse_return enum to parse_output that weren't updated -- not sure why this wasn't causing builds to fail.

### Proposed changes
- parse_return => parse_output in appMain.cpp and helics-broker.cpp files

### DIFF
--- a/src/helics/apps/appMain.cpp
+++ b/src/helics/apps/appMain.cpp
@@ -108,10 +108,10 @@ int main (int argc, char *argv[])
 
     switch (ret)
     {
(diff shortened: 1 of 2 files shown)


phlptp:
I made these same modifications in the just merged json_aggregate_output branch so I am going to close this PR.  

My guess is the builds never got updated.  I think we can turn on a setting that all PR's need to be updated to the latest head but that caused a number of issues too so we turned it off.

PR #750, opened 2019-07-18
Broker key

phlptp:
### Summary

If merged this pull request will add a broker capability to HELICS.  This means if a key is specified at a broker,  cores must have that same key to connect and if a core tries to use a key and a broker doesn't have one it also errors.  

### Proposed changes

- add broker key capability

- clear a couple TODO's related core errors

- add an errorCode and errorstring storage in the broker base so it accessible later


phlptp:
@nightlark I may be adding a few more commits later,   but any feedback on what exists currently would be welcome.

nightlark:
It would be useful to have the key argument show up in the help message(s).

PR #753, opened 2019-07-22
Utilities lib

phlptp:
### Summary

If merged this pull request will split out the utilities library from HELICS to be used as a submodule.  This contains some stringOperations merges with some other operations from GridDyn, as well as the timeRepresentation.  

### Proposed changes

- use utilities submodule
- remove utilities code from helics::common
- fix all the related code changes


src/helics/common/CMakeLists.txt
```diff
@@ -7,18 +7,13 @@
...
-  timeRepresentation.hpp
   JsonProcessingFunctions.hpp
   JsonBuilder.hpp
   TomlProcessingFunctions.hpp
   AsioContextManager.h
   logger.h
   loggerCore.hpp
-  stringOps.h
   GuardedTypes.hpp
-  generic_string_ops.hpp
```
nightlark: It looks like there is still a copy of this file in src/helics/common

CONTRIBUTORS.md
```diff
@@ -67,6 +67,9 @@ Several containers developed for HELICS and other applications were branched int
 ### [gmlc/concurrency](https://github.com/GMLC-TDC/concurrency)  
 Several concurrency related structures and containers were developed for HELICS and other applications and were branched into a separate repository as a header only library and also includes the modified [libGuarded](https://github.com/copperspice/libguarded). It is included in HELICS as a submodule and is released under a [BSD-3 clause](https://github.com/GMLC-TDC/concurrency/blob/master/LICENSE) license.
 
+### [gmlc/utilities](https://github.com/GMLC-TDC/utilities)  
+Several generic operations and utilities from HELICS and GridDyn are encapsuled in a separate repository. mostly dealing with String Operations but including a few other features as well.  It is included in HELICS as a submodule and is released under a [BSD-3 clause](https://github.com/GMLC-TDC/utilities/blob/master/LICENSE) license.
```
nightlark:
```suggestion
Several generic operations and utilities from HELICS and GridDyn are encapsuled in a separate repository, mostly dealing with String Operations but including a few other features as well.  It is included in HELICS as a submodule and is released under a [BSD-3 clause](https://github.com/GMLC-TDC/utilities/blob/master/LICENSE) license.
```

nightlark:
Should the gmlc headers from the utilities library be installed in `helics/gmlc` or `helics/external/gmlc`, similar to some of the 3rd party libraries whose headers we install?

phlptp:
I thought about it, but we need the helics_time header to work properly,   There are a few options of how to do that I suppose.  But for simplicity for now I just put timeRepresentation on a path that the header would find it.    open to other ideas on how to handle that, but that header is the key driver.

phlptp:
It would appear the location timeRepresentation got placed doesn't work either based on the results of the examples build

PR #756, opened 2019-07-27
Include licenses for 3rd party libraries with HELICS installations

nightlark:
### Summary

If merged this pull request will include a combined license file with HELICS installations to ensure compliance with the licenses used by 3rd party libraries.

### Proposed changes

- Generate a combined license file for 3rd party libraries using CMake
- Include 3rd party licenses file with HELICS installations


CMakeLists.txt
```diff
@@ -797,11 +797,27 @@ install(
...
+	"cereal" "${PROJECT_SOURCE_DIR}/ThirdParty/cereal/LICENSE"
+	"GMLC Concurrency" "${PROJECT_SOURCE_DIR}/ThirdParty/concurrency/LICENSE"
+	"GMLC Containers" "${PROJECT_SOURCE_DIR}/ThirdParty/containers/LICENSE"
+	"cppzmq" "${PROJECT_SOURCE_DIR}/ThirdParty/cppzmq/LICENSE"
+	"fmt" "${PROJECT_SOURCE_DIR}/ThirdParty/fmtlib/LICENSE.rst"
+	"Google Test" "${PROJECT_SOURCE_DIR}/ThirdParty/googletest/LICENSE"
+	"JsonCpp" "${PROJECT_SOURCE_DIR}/ThirdParty/jsoncpp/LICENSE"
+	"tinytoml" "${PROJECT_SOURCE_DIR}/ThirdParty/toml/LICENSE"
+	"GMLC Utilities" "${PROJECT_SOURCE_DIR}/ThirdParty/utilities/LICENSE"
+)
```
phlptp: as we add more subproject we will just add them to this list?

phlptp: are we already copying over the libzmq license?  
also would it make sense to make this a variable that can be appended conditionally in some cases?

We probably want the libguarded license added as well concurrency\gmlc\libguarded\LICENSE

nightlark: Yes, we will just add them to this list -- parsing .gitmodules or doing a search for possible license files is an option, but I feel like maintaining a list is easier.

For libzmq I'm not doing anything with the license (if we're using it as a shared library I don't think we need to deal with installing the license?). That might be one where appending conditionally is needed, based on if autobuild is on (otherwise finding a system copy of the license file may be hard).

PR #757, opened 2019-07-27
Improve external network interface selection reliability

nightlark:
### Summary

If merged this pull request will improve the reliability of the external interface selection when using the resolver could return an incorrect result, such as in emulated networks.

### Proposed changes

- Uses OS syscalls to get all network interface addresses
- Makes sure the resolved network interface address exists on the system
- Prioritizes picking addresses that both the OS syscall and resolver returned
- Falls back to picking a non-loopback address if no interface with an address matching the resolved address exists
- Uses resolved addresses that aren't assigned to an interface last

### DIFF
--- a/ThirdParty/netif/gmlc/netif/NetIF.hpp
+++ b/ThirdParty/netif/gmlc/netif/NetIF.hpp
@@ -32,9 +32,15 @@ namespace gmlc
 namespace netif
 {
   #if defined(_WIN32)
-    typedef IF_ADDRS PIP_ADAPTER_ADDRESSES
+    using IF_ADDRS = PIP_ADAPTER_ADDRESSES;
   #else
-    typedef IF_ADDRS struct ifaddrs*
+    using IF_ADDRS = struct ifaddrs*;
+  #endif
+
+  #if defined(_WIN32)
+    using IF_ADDRS_UNICAST = PIP_ADAPTER_UNICAST_ADDRESS;
+  #else
+    using IF_ADDRS_UNICAST = struct ifaddrs *;
   #endif
 
   inline std::string addressToString(struct sockaddr *addr)
@@ -98,7 +104,7 @@ namespace netif
     #endif
   }
 
-  inline auto getSockAddr(IF_ADDRS addr)
+  inline auto getSockAddr(IF_ADDRS_UNICAST addr)
   {
     #if defined(_WIN32)
       return addr->Address.lpSockaddr;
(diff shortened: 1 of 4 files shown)


phlptp:
would your intention be to move this to an eventual networking repo when we get around to that?

ThirdParty/netif/gmlc/netif/NetIF.hpp
```diff
@@ -0,0 +1,193 @@
...
+#else
+  #include <arpa/inet.h>
+  #include <ifaddrs.h>
+#endif
+
+namespace gmlc
+{
+namespace netif
+{
+  inline std::string addressToString(struct sockaddr *addr)
```
phlptp: we probably should add a doxygen comment on these functions stating what they do

ThirdParty/netif/gmlc/netif/NetIF.hpp
```diff
@@ -0,0 +1,193 @@
...
+        src_addr = &((struct sockaddr_in6 *)addr)->sin6_addr;
+        break;
+      default: // Invalid address type for conversion to text
+        return std::string();
+    }
+    inet_ntop(family, src_addr, addr_str, INET6_ADDRSTRLEN);
+    return std::string(addr_str);
+  }
+	
+  template <class A>
```
phlptp: And the requirements on types for A?

nightlark:
> would your intention be to move this to an eventual networking repo when we get around to that?
Eventually I want it  in a repo with CI set up that can build/test it on an assortment of OSes (Windows, Linux, macOS; MinGW/MSYS, BSD?).

ThirdParty/netif/gmlc/netif/NetIF.hpp
nightlark: On Windows it is `PIP_ADAPTER_ADDRESSES` and on other platforms (macOS, Linux) it is `struct ifaddrs *`; I'm not sure if other platforms use something else for their interface address structure.

I'm not sure if using templates+decltype has any benefits or is just overcomplicating things -- do you think it would make sense to switch to using a typedef based on platform instead?

phlptp: That might be a bit clearer than a template with no restrictions?  I am sure it would error, but as a template I would have no idea how to use it.  with the typedefs ( or just completely separate functions) I could probably figure out by the arguments.

nightlark:
There are some unit tests that would be nice to have, but I think they would need to be run in a VM or emulated network to avoid having the results vary on different systems.

Issue #606, opened 2019-01-18
Logging API to get user messages into the logging system


phlptp:
**Is your feature request related to a problem? Please describe.**
The need for more logging 

**Describe the solution you'd like**
a clear function call for getting user messages into the log

This is a feature request to expose the Core API logging calls to the application API and C library.

trevorhardy:
To add some details, when using Python's `logging` package, message sent to `logging.info()` make there way into the HELICS log file but appear to be flushed at times that are inconsistent with the HELICS log messages. This creates confusion in the log file and makes it impossible to understand the order of events as reported by the federate. A native HELICS logging function would allow federate authors to write messages into the HELICS log file that would ensure synchronization with the other HELICS log messages.

```
EV_Controller (131073) (0)::Granted Time=21600
EV_Controller (131073) (0)::exec=21600 allow=21600, value=21301, message=9.22337e+09, minDe=21301 minminDe=21301
EV_Controller (131073) (0)::exec=21600 allow=21600, value=21301, message=9.22337e+09, minDe=21301 minminDe=21301
EV_Controller (131073) (0)::processing cmd time_grant:From (1) Granted Time(21600) to (131073)
EV_Controller (131073) (0)::processing cmd time_grant:From (131074) Granted Time(21600) to (131073)
EV_Controller (131073) (0)::processing cmd time_grant:From (131073) Granted Time(21600) to (131073)
EV_Controller (131073) (0)::processing cmd pub:From (131074) handle(0) size 17 at 21600 to 131073
EV_Controller (131073) (0)::exec=21600 allow=21600, value=21301, message=9.22337e+09, minDe=21301 minminDe=21301
EV_Controller (131073) (0)::receive publication pub:From (131074) handle(0) size 0 at 21600 to 131073
EV_Controller (131073) (0)::processing cmd pub:From (131074) handle(1) size 17 at 21600 to 131073
EV_Controller (131073) (0)::exec=21600 allow=21600, value=21301, message=9.22337e+09, minDe=21301 minminDe=21301
EV_Controller (131073) (0)::receive publication pub:From (131074) handle(1) size 0 at 21600 to 131073
EV_Controller (131073) (0)::processing cmd pub:From (131074) hanEV Controller grantedtime = 21900.0
Total Feeder Load is 2881032.742045034 + 933250.2484611124 j
EV Controller grantedtime = 22200.0
Total Feeder Load is 2892596.679470324 + 936856.727896499 j
[...]

PR #758, opened 2019-07-30
Logging api enhancement

eberleim:
### Summary

If merged this pull request will expose user-accessible log writing interfaces to the Federate API.

### Proposed changes

-See #606 
-Add additional test cases and API components for external libraries (@10tranr)


src/helics/application_api/Federate.cpp
```diff
@@ -9,6 +9,7 @@ SPDX-License-Identifier: BSD-3-Clause
 #include "../core/CoreFactory.hpp"
 #include "../core/core-exceptions.hpp"
 #include "../core/helics_definitions.hpp"
+#include "../core/CommonCore.hpp"
```
phlptp: The federate code should not need to be aware of CommonCore to maintain the layers properly.   All federate interaction should be with the interface in the core api (Core.hpp), otherwise we will run into a lot of trouble in the future making changes.

src/helics/core/federate_id.hpp
```diff
@@ -15,6 +15,51 @@ namespace helics
 /** base date type for id types*/
 using identififier_base_type = int32_t;
 
+/** class defining a local_federate_id
```
phlptp: Is there a reason this was shifted in the file?

src/helics/application_api/Federate.hpp
```diff
@@ -73,7 +73,9 @@ class Federate
       asyncCallInfo;  //!< pointer to a class defining the async call information
     std::unique_ptr<FilterFederateManager> fManager;  //!< class for managing filter operations
     std::string name;  //!< the name of the federate
-  public:
+    std::function<void(int, const std::string &, const std::string &)> loggerFunction; //!< Logging function
```
phlptp: This should probably use the setLoggingCallback method on the core instead of storing it separately.  That way any log messages from the internal federate would get logged as well and make sure things are synchronized properly.

src/helics/application_api/Federate.hpp
```diff
@@ -483,6 +485,27 @@ class Federate
...
+    /** log a simplified message to the federate Logger
+    @param message the message to log
+    */
+    virtual void logMessage (const std::string &message) const;
+
+    /** set the logging function
+    @details function must have signature void(int level, const std::string &sourceName, const std::string
+    &message)
+    */
+    virtual void setLogger (std::function<void(int, const std::string &, const std::string &)> logFunction)
```
phlptp: this would seem a duplicate function to setLoggingCallback,  Are you thinking there is some distinction between this logger and the callback used in setLoggingCallback which goes to the core?

phlptp:
I like the addition of logMessage functions in the federate.    I think I would like to better understand the advantages of separate log function for the federate vs.  the one that can also be called by the federate representation in the core.

src/helics/application_api/Federate.cpp
eberleim: This was accidentally left in. It is not necessary, and will be removed.

src/helics/core/federate_id.hpp
eberleim: This was moved so base_type and global_broker_id_shift would both be available for the globalIndex method added to local_federate_id (inverse of global_federate_id's localIndex). 
I simply moved all the constexpr together as a group, to avoid scattering them throughout the file.

src/helics/application_api/Federate.hpp
eberleim: This was also my first inclination, however doing so resulted in a series of behavior problems that I was ultimately unable to resolve without changing components in the Core and CommonCore which I was not familiar enough with to change (possibly big breaking components). 
The trace using the setLoggingCallback and passing messages through the core ultimate reaches a point of failure at CommonCore.cpp:1788 when that path of logic fails to identify as a federate, gets a nullptr, and returns false, never logging the message. 
This may itself be a bug, rather than a configuration issue, however I was unable to resolve it in a way that I found satisfactory.

phlptp: I think we should revisit that and if there is some bugs we should fix them.  The way it is setup in this PR is that only the messages that get sent from the new federate API calls will end up in that log, and I don't think that is really what we intend, we really want an ability to merge the core generated message with user generate ones, to get a good sequence.

phlptp:
@eberleim I updated this branch on top of the latest develop and removed a few of the functions you had put in.  global_federate_id should not be accessible in the core API and the callLogger operation should not be accessible in the core so those functions had to be removed otherwise they would break the layer encapsulation in HELICS,  I update the header files so global_federate_id is not part of the public API an is purely used in the core and not accessible elsewhere.  

I want to get a series of tests in the system tests that actually exercise the logging functions, so I was trying to figure out what your use case was.  Clearly we were missing a function in the Federate API to get log messages into the main logging system so that is clearly needed but I was wondering what you were trying to accomplish with some of the other changes, so we can make some test cases that reflect the desired use cases.

eberleim:
@phlptp It looks like you successfully addressed the issues most of my tweaks were intended to address (I unfortunately did not understand the original intent of some of the code, and was concerned with moving or changing it). 
I added a few additional test cases, but overall it appears to be working as would be desired.

phlptp:
I think there might be a few issues with the file operations on the logging yet that I came across when looking at this, so there are a few more tests I want to add, but I think it mostly is working now.  We also need to add the new functions to the C api, and a few tests of that interface.

phlptp:
There is already an [enumeration](
https://github.com/GMLC-TDC/HELICS/blob/ee7300bbbb171bea9e159d9db283ece430c5452f/src/helics/helics_enums.h#L111)
of the logging levels.    Though that comment is probably out of date now since there are more levels.

Changing the numbers right now would probably be considered a breaking change, but one we probably should consider for HELICS 3.0.  

Are you suggesting in the C api we have a `logDebugMessage`,  `logErrorMessage`, `logWarningMessage`, and `logSummaryMessage` functions predefined with the appropriate levels instead of a generic logMessage,  I think we would still need to keep the `LogLevelMessage`

nightlark:
Ah well that's good. It looks like some of the comments have numbers that don't match the enum.

Yea, some functions predefined with appropriate levels that call `logLevelMessage`, which users could still call directly if they wanted to.

phlptp:
@nightlark  can you take a look at the recent updates with some of your recommendations
@eberleim what do you think of @nightlark suggestions and the changes?

eberleim:
@phlptp I think the feedback and the resulting changes make sense. The generic logMessage function arbitrarily being assigned a level was a good prototype, but the more specifically named functions will be easier for the users. 
@nightlark 's comment on the use of log-level 8 in the tests is a good one, I had chosen that simply to be above/below the threshold to verify the filtering was working, however, it might be worth adding a test for arbitrarily high log level (500), and making sure we have some reasonable error output when the level exceeds our defined levels (1-9 as far as I'm aware).
I'll try to put a few more tests together this weekend that clearly demonstrate that behavior if I can, and if we don't currently have a warning/error for that, we'll at least have a good demonstration.

phlptp:
@nightlark  brought up the possibility of making a much bigger gap between the different levels.  Internally I think we mostly use the enumeration levels, but would changing the enumeration levels be considered a breaking change to allow more space between them for customized logging levels for users.   I think this is a good idea that should make it into HELICS 3.0 but I am not entirely clear if this would be a breaking change now and something we shouldn't do yet.  It would seem to fall in a bit of a gray area.  

As far as limited levels,  I am not sure I want to do that yet given the possibility of custom levels.

phlptp:
@eberleim Do you have an ETA on any additional tests?

phlptp:
I think I am going to merge this now and we can work on some additional tests and error messages if desired at a later point.  I have a feeling working out additional details on the logging will be an ongoing effort.

PR #762, opened 2019-08-02
New message api

phlptp:
### Summary

If merged this pull request will add a new message object API for better access to messages in the C and language API's, and fixes an inability to access raw data in the language API's for messages.

### Proposed changes

- add messageObject api to the C shared library
- add a few test cases using the API

The old API is still there and functional but will likely be deprecated in HELICS 2.3
This new api actually could be faster in C, since it allows access to the messages with no copying.  
and a much cleaner way to potentially do custom filters in the language API's

### DIFF
--- a/src/helics/shared_api_library/MessageFederate.h
+++ b/src/helics/shared_api_library/MessageFederate.h
@@ -112,6 +112,12 @@ extern "C"
     @param[in,out] err a pointer to an error object for catching errors
     */
     HELICS_EXPORT void helicsEndpointSendMessage (helics_endpoint endpoint, helics_message *message, helics_error *err);
+    /** send a message object from a specific endpoint
+    @param endpoint the endpoint to send the data from
+    @param message the actual message to send
+    @param[in,out] err a pointer to an error object for catching errors
+    */
+    HELICS_EXPORT void helicsEndpointSendMessageObject (helics_endpoint endpoint, helics_message_object message, helics_error *err);
 
     /** subscribe an endpoint to a publication
     @param endpoint the endpoint to use
@@ -144,6 +150,11 @@ extern "C"
     @return a message object*/
     HELICS_EXPORT helics_message helicsEndpointGetMessage (helics_endpoint endpoint);
 
+    /** receive a packet from a particular endpoint
+    @param[in] endpoint the identifier for the endpoint
+    @return a message object*/
+    HELICS_EXPORT helics_message_object helicsEndpointGetMessageObject (helics_endpoint endpoint);
+
     /** receive a communication message for any endpoint in the federate
     @details the return order will be in order of endpoint creation then order of arrival
     all messages for the first endpoint, then all for the second, and so on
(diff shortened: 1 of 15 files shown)


src/helics/shared_api_library/MessageFederate.h
```diff
@@ -144,13 +150,42 @@ extern "C"
...
     /** receive a communication message for any endpoint in the federate
     @details the return order will be in order of endpoint creation then order of arrival
     all messages for the first endpoint, then all for the second, and so on
     within a single endpoint the messages are ordered by time, then source_id, then order of arrival
     @return a unique_ptr to a Message object containing the message data*/
     HELICS_EXPORT helics_message helicsFederateGetMessage (helics_federate fed);
 
+    /** receive a communication message for any endpoint in the federate
+    @details the return order will be in order of endpoint creation then order of arrival
+    all messages for the first endpoint, then all for the second, and so on
```
kdheepak: "order of arrival *of* all messages"

src/helics/shared_api_library/MessageFederate.h
```diff
@@ -190,6 +225,148 @@ extern "C"
...
+    @param[out] actualSize  the actual length of data copied to data
+    @param[in,out] err a pointer to an error object for catching errors
+    */
+    HELICS_EXPORT void helicsMessageGetRawData (helics_message_object message, void *data, int maxlen, int *actualSize, helics_error *err);
+
+    /** get a pointer to the raw data of a message
+    @param message a message object to get the data for
+    @return a pointer to the raw data in memory
+    */
+    HELICS_EXPORT void *helicsMessageGetRawDataPointer (helics_message_object message);
```
kdheepak: Should we add a `#define` for this `void *` return type?

src/helics/shared_api_library/MessageFederateExport.cpp
```diff
@@ -448,23 +497,111 @@ helics_message helicsFederateGetMessage (helics_federate fed)
...
+        return nullptr;
+    }
+    if (!endObj->messages.empty ())
+    {
+        helics_message_object mess = endObj->messages.back ().get ();
+        return mess;
+    }
+    return nullptr;
+}
+
```
kdheepak: `Last` message is ambiguous. Does this refer to oldest message? Users are most likely going to want to use the current time as a reference, so last and first in this context is not clear.

kdheepak:
I don't know if I'm a fan of using `MessageObject`. Can we deprecate `helicsFederateGetMessage` in one version and change the meaning of `helicsFederateGetMessage` in the next version?

src/helics/shared_api_library/MessageFederate.h
phlptp: this function should not be available in the swig generated functions, if it is we need to fix something.    It is really only useful in a C context so we can have access to the underlying data without copying the data.  So in C, I would prefer it remain a void * vs char * or some other name to match C conventions for raw data.  If it were a char * it might be interpreted as a string and that could cause issues.  void * can be cast easily.    Like I said this function doesn't make sense in the language bindings since they wouldn't have direct access to raw data safely.

src/helics/shared_api_library/MessageFederateExport.cpp
phlptp: hmmm,   how about `getRecentMessage` or `getMostRecentMessage`   I could also just drop this function it isn't required.  It is there for convenience since it would be easy to go back and get a message again now.

phlptp:
So I think the plan for versioning would be to label the messageObject as experimental for 2.2,  stabilize it in 2.3 and deprecate the previous getMessage function call.  and in 3.0 remove the deprecated functions and update them to the new name, so there wouldn't be a getMessageObject in 3.0 any longer.

src/helics/shared_api_library/MessageFederate.h
phlptp: I will clarify this on the next commit

phlptp:
I removed the *getLastMessage functions.  The intention was to enable a way to get object to previous messages but, as I look at it again, that function may be unclear as to purpose and functions, so I am removing it for now.  If we need it again later we can come back to it and revisit in a later version.

PR #768, opened 2019-08-08
Fix a few typos

mattirish:
TIL that dissemble means "to conceal one's true motives, feelings, or beliefs"
Who knew!?


phlptp:
I suppose that's why it passed the spell check, interesting

PR #771, opened 2019-08-10
Federate clone

phlptp:
### Summary

If merged this pull request will add a clone app to the helics_app for cloning a federate

### Proposed changes

- add Clone app
- add clone.md for documentation
- add a few tests of the clone functionality


phlptp:
I was starting to do some of that with the json files,  then reading the json and doing some value comparison.  That prevents the need for precise file matching

PR #774, opened 2019-08-13
Git all archive implementation

dnadeau4:
### Summary

add git-all-archive.py to create a tarball with archive source and all submodules source file

### Proposed changes

python scripts/git-all-archive.py will archive current repository and create a tarball with the branch name.  HELIC-<branch>.tar.gz


phlptp:
The code itself looks good, and as far as I can tell the output looks good.  
A couple things we might want to add.
Somewhere in the comments of the python code or a small script add the install requirements and commands
`pip install pygithub` and the additional requirements.    the first time I ran I had a number of errors and had to figure out what I needed.  A separate short script that does the install might be good as we will likely run this in a CI job, that will make installing requirements easier.

Second I am wondering if some cleanup might be useful, such that once the zip file is constructed the intermediate files are deleted,  maybe as an additional option of some kind.

phlptp:
@nightlark  what do you think?

kdheepak:
If you want to not have a Python dependency for this, there's a few command line tools ([1, 2, 3]) with no dependencies that exist for this very purpose: uploading files to a github tags.

[1] https://github.com/tfausak/github-release/releases/tag/1.2.4
[2] https://github.com/aktau/github-release/releases/tag/v0.7.2
[3] https://github.com/kdheepak/github-release/releases/tag/v0.1.4

^ I wrote this one ([3]), and it is used to upload binaries of itself to github releases from Travis. It's pre-compiled command line tool executable with no dependencies so it should work everywhere. I can also address issues if you want to use that instead :)

scripts/git-all-archive.py
```diff
@@ -0,0 +1,254 @@
...
+    stdoutdata, stderrdata = runcmd(cmd)
+    print("> archiving all submodules")
+    cmd = 'git submodule foreach --recursive \'git archive --verbose --prefix=repo/$path/ --format tar master --output ' +\
+        currentdir + '/repo-output-sub-$sha1.tar\''
+    stdoutdata, stderrdata = runcmd(cmd)
+
+    # Delete pre-existing work
+    #
+    if os.path.exists("/tmp/repo"):
+        shutil.rmtree("/tmp/repo")
```
nightlark: I think it would be better to create a temporary directory using the python function instead of using /tmp, for systems that use different locations for the temp directory (e.g. on one of the clusters TMPDIR is set to /var/tmp/<username> which avoids multiple users trying to use files with the same name): https://docs.python.org/3.7/library/tempfile.html#tempfile.TemporaryDirectory

nightlark:
> If you want to not have a Python dependency for this, there's a few command line tools ([1, 2, 3]) with no dependencies that exist for this very purpose: uploading files to a github tags.
> [1] https://github.com/tfausak/github-release/releases/tag/1.2.4
> [2] https://github.com/aktau/github-release/releases/tag/v0.7.2
> [3] https://github.com/kdheepak/github-release/releases/tag/v0.1.4
> ^ I wrote this one ([3]), and it is used to upload binaries of itself to github releases from Travis. It's a pre-compiled command line tool executable with no dependencies so it should work everywhere. I can also address issues if you want to use that instead :)

Most of the CI services also seem to provide tasks to deploy to GitHub Releases -- if those have the functionality needed, maybe we can skip installing dependencies/binaries for the deploy step.

https://docs.travis-ci.com/user/deployment/releases/
https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/utility/github-release?view=azure-devops

dnadeau4:
I did not have time to test the archive file.
 
I used a TMPDIR now and I cleaned all the tar file.
I go the docker working on circle ci with OCTAVE
    https://circleci.com/gh/GMLC-TDC/HELICS/tree/git-all-archive

dnadeau4:
https://circleci.com/gh/GMLC-TDC/HELICS/8?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link

Dockerfile-Octave
```diff
@@ -0,0 +1,19 @@
+FROM ubuntu:18.04 as builder
```
phlptp: We should probably store these somewhere more organized since I expect we will end up with a few of them them but I am not sure what is best yet so this is fine for now

PR #780, opened 2019-08-21
add clang memory sanitizer and docker

dnadeau4:
### Summary

Add docker using clang libcxx and libcxxabi with memory sanitizer flag enable.
```
 ENV MSAN_CFLAGS="-fsanitize=memory -stdlib=libc++ -L/root/develop/libcxx_msan/lib -lc++abi -I/root/develop/libcxx_msan/include -I/root/develop/libcxx_msan/include/c++/v1 -Wno-unused-command-line-argument -fno-omit-frame-pointer -g -O1 -Wl,-rpath=/root/develop/libcxx_msan/lib"
```
### Proposed changes

New docker file found in config/docker directory
-


.circleci/config.yml
```diff
@@ -20,24 +20,70 @@ aliases:
...
   helics_test:
     jobs:
       - helicsoctave
-#           requires:
-#              - previous_test_passed
+      - helicsMSan
+  nightly:
+    triggers:
+      - schedule:
+          cron: "25 * * * *"
```
nightlark: Is this scheduled to run at :25 past every hour?

dnadeau4: I was trying to trigger at 2:25pm.  It did work.  I just changed it to `"0 0 * * *"` and branch `develop`

dnadeau4: BTW checking out the `developer` branch should pass since I did not change the code.

.circleci/config.yml
```diff
@@ -20,24 +20,70 @@ aliases:
        echo "*** Octave test result: "${RESULT}
        exit $RESULT
 
+  - &run_helics_tests
+    name: run_helics_tests
+    command: |
+       SUMRESULT=0
```
phlptp: would it be possible to put this in a script? it is likely in the future there will be a few additional tests that need to be added,  It would simplify future maintenance if it were in a script vs purely in the .yml.

Also might be useful to have this be in a for each loop and then a list of the tests?

dnadeau4: allright, but I will delete the intermediate messages such as
```
       echo "*** Helics core-test result: "${RESULT}
```
and maybe just 
```
       echo "*** last results: ${RESULT}"   
```

I'll try to echo the command line so that you know which one past/failed.

dnadeau4:
@phlptp you can merge when it's all green, the new script is in there and worked.   I did not put the license on top of the file, and realized that most of them don't have one.  😏

phlptp:
I should try to stick a memory bug in at some point just to see it fail and make sure that is working.

phlptp:
For the licensing I will be making a cleanup commit shortly in the next few days that can add it.  Or if you make another PR with the thread sanitizer, you can add it then.

PR #781, opened 2019-08-21
MinGW build without PowerShell

nightlark:
### Summary
If merged this pull request will get use the standard script section on Travis for the MinGW build.

### Proposed changes
- Use the same script step in Travis for the MinGW build as for the other builds
- Remove the use of powershell for running MinGW build (not needed)
- Echo some of the build commands with variables expanded

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,7 @@ before_cache:
 
 env:
   global:
-    - MAX_BUILD_PARALLEL=2
+    - MAKEFLAGS="-j 2"
 
 # Create aliases for some of shared build configuration
 _basic_env:
@@ -262,7 +262,7 @@ install:
 script:
   - mkdir -p build && cd build
   - cmake ${JOB_OPTION_FLAGS} ${HELICS_DEPENDENCY_FLAGS} ${HELICS_OPTION_FLAGS} ${CMAKE_COMPILER_LAUNCHER} ..
-  - cmake --build . --parallel ${MAX_BUILD_PARALLEL}
+  - cmake --build . -- ${MAKEFLAGS}
 
   # Setup counters for coverage
   - if [[ "$BUILD_TYPE" == "Coverage" ]]; then pushd .. && scripts/lcov-helper.sh setup-counters && popd ; fi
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -13,6 +13,7 @@ before_cache:
 env:
   global:
     - MAKEFLAGS="-j 2"
+    - CMAKE_GENERATOR="Unix Makefiles"
```
phlptp: Is this just adding a default generator?

nightlark: Yea.

.travis.yml
```diff
@@ -261,11 +259,12 @@ before_install:
 install:
   - source scripts/install-ci-dependencies.sh
   - source scripts/setup-helics-ci-options.sh
+  - export CMAKE_COMPILER_LAUNCHER="-DCMAKE_C_COMPILER_LAUNCHER=ccache -DCMAKE_CXX_COMPILER_LAUNCHER=ccache"
   
 script:
   - mkdir -p build && cd build
-  - cmake .. ${JOB_OPTION_FLAGS} ${HELICS_DEPENDENCY_FLAGS} ${HELICS_OPTION_FLAGS} -DCMAKE_C_COMPILER_LAUNCHER=ccache -DCMAKE_CXX_COMPILER_LAUNCHER=ccache
-  - make ${MAKEFLAGS}
+  - cmake -G "${CMAKE_GENERATOR}" ${JOB_OPTION_FLAGS} ${HELICS_DEPENDENCY_FLAGS} ${HELICS_OPTION_FLAGS} ${CMAKE_COMPILER_LAUNCHER} ..
```
phlptp: Denis had a request to echo the full expanded cmake command in the build log so we could easily see which options were being used for cmake on the different tests.  The variable are useful but they make it difficult to see what we are doing sometimes.

phlptp:
looks like this broke something with the clang 3.6 build on travis.  Looks like something with boost.

nightlark:
Looks like it was accidentally removing the line that disabled the Boost tests.

PR #782, opened 2019-08-23
Broker server

phlptp:
### Summary

If merged this pull request will add a broker server for ZMQ brokers

### Proposed changes

- add a broker server and associated messages
- add some new files for putting all the network defaults in a single place


phlptp:
@nightlark there are a few things to add yet and I am working on more test cases but I though you might like to take a look and see if you have any suggestions.

src/helics/core/networkDefaults.hpp
```diff
@@ -0,0 +1,25 @@
...
+*/
+#pragma once
+
+// For ZMQ Comms
+constexpr int DEFAULT_ZMQ_BROKER_PORT_NUMBER = 23404;
+
+constexpr int BACKUP_ZMQ_BROKER_PORT_NUMBER = 23406;
+
+constexpr int DEFAULT_ZMQSS_BROKER_PORT_NUMBER =
+  23414;  // Todo define a different port number (change in HELICS 3.0)
```
nightlark: What's the problem with this port number?

tests/helics/core/TcpSSCore-tests.cpp
```diff
@@ -27,7 +28,6 @@ using namespace std::literals::chrono_literals;
 using asio::ip::tcp;
 using helics::Core;
 
-#define TCP_BROKER_PORT 33133
 #define TCP_BROKER_PORT_STRING "33133"
```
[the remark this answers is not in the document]
phlptp: The alternate port is just used for the testing, I don't think it has much use outside of that.  A string version ?  maybe, but the only place it is actively used right now is in the test code just so we don't have a bunch of extra to_string() calls.

src/helics/core/networkDefaults.hpp
phlptp: It is right now in the range for the auto allocation of ZMQ brokers.  I would like a little more separation between them, but I don't want to change it until 3.0 as that could cause some issues between the versions.

PR #788, opened 2019-08-28
Add option to set number of threads when building Docker image

nightlark:
Adds an argument that can be used when running docker build to change the number of threads make uses in several of the dockerfiles used by HELICS (default 2; use `--build-arg MAKE_PARALLEL=4` to build using 4 threads).

Another option would be getting rid of `-j` and changing the argument to be for passing in extra flags/options to make.


phlptp:
might want to update the docker.md as part of this change with the new args

PR #789, opened 2019-08-29
Develop

phlptp:
### Summary

If merged this pull request will update master to v2.2.0

### Proposed changes

-


kdheepak:
Documentation for a bunch of MATLAB functions seems to have been removed. Do you know why?

kdheepak:
Should we add the [link to the PESGM tutorial](https://github.com/GMLC-TDC/pesgm-2019-helics-tutorial) in the documentation here?

phlptp:
For docs I am not opposed to a direct update to master that is outside the tagged version, so I am not sure that is needed right now.   But we should add that at some point.

phlptp:
My guess is it has something to do with having to turn off the output.i in the swig.  That had a bunch of old references in it that was starting to cause some trouble.  I think we need to spend some time figuring out the swig doc system to clean that up a little.

phlptp:
should probably make an issue about the code help and swig though,  it is something we need to deal with at some point.

PR #792, opened 2019-08-29
Use a multi-stage Docker build for the HELICS apps container

nightlark:
### Summary
If merged this pull request will build the HELICS apps docker container using a multistage build to reduce the size, along with doing a Release build of the apps.

Some numbers for the effect of the changes on sizes-- the image size shown by `docker images` goes from 1.01GB -> 265MB from the multistage build, setting the CMake build type to Release decreased the size from 265MB to 132MB. On DockerHub, the size listed next to the tag went from 300MB to 58MB with both changes.

### Proposed changes
- Change the CMAKE_BUILD_TYPE to Release
- Use a multi-stage build to create the HELICS apps Docker image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ Minor release with some updates to the networking portion of HELICS and some API
 -  CI tests using docker for clang memory sanitizer and the octave interface.
 -  Scripts for generating a single zip file with all the code including submodules.  This will be generated for each new release.  
 -  A broker server that generate multiple brokers on a single system and handles the port allocation intelligently. (Only ZMQ currently supported, this is not backwards compatible, though regular 2.2 brokers should work with 2.1 federates if needed.)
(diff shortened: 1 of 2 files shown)


phlptp:
have you verified that the helics_broker and other apps are on the path in the new image?

phlptp:
Also can you add a line in the change log then I think we can get this in.  I want to give @kdheepak  a couple hours to review anyway.

nightlark:
Using `docker run helics-test helics_broker --version` worked, and `which helics_broker` from an interactive shell in the docker container gave the path as `/usr/bin/helics_broker`.

I'll also check with the tagged DockerHub image.

nightlark:
Here's the default PATH in the Ubuntu:18.04 image we're using.
```bash
root@5876b9199d15:/# echo $PATH
/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
```

Running the broker and a couple echo apps in the docker container (pulled from Docker Hub) worked -- I'm not sure how communication with brokers/federates outside of the container works, but that would be good to document once we figure out what arguments/addresses are needed.

PR #794, opened 2019-08-31
try to get the circle-ci asan operational

phlptp:
### Summary

If merged this pull request will update the circle_ci tests specifically to get the asan test passing

### Proposed changes

- update the build files slightly
- add a blacklist for the ASAN configuration


phlptp:
@nightlark  I think I got this ironed out now.  The asan test should be passing unless something new comes up, and I got the image almost cut in half for helics/clang-asan

PR #795, opened 2019-09-01
Use DockerHub source code checkout

nightlark:
### Summary
If merged this pull request will use the copy of HELICS checked out by DockerHub instead of cloning from GitHub as a step. Docker build needs to be run from the root folder of the HELICS repository.

Unfortunately, turning caching on still messes up the image -- I think the way that we can cut down on the time will be having some docker images set up for building with all dependencies installed that we use as a base for the first stage so we can skip running apt-get.

### Proposed changes
- Copy HELICS checked out by DockerHub (current directory of host) to the builder Docker image


phlptp:
The issue I have with this is that I think we then need a separate dockerfile so that we can have a dockerfile that can be run directly and not depend on something in the environment that is running it.  I like the idea  but I think it should be dedicated dockerfile for building dockerhub images and we change the build config to match.  Then have a separate dockerfile that can do it completely in an isolated context.

phlptp:
also a little confused why the Circle-CI asan is failing.  Seems to not be picking up the suppression file from the merge this morning.

nightlark:
> The issue I have with this is that I think we then need a separate dockerfile so that we can have a dockerfile that can be run directly and not depend on something in the environment that is running it. I like the idea but I think it should be dedicated dockerfile for building dockerhub images and we change the build config to match. Then have a separate dockerfile that can do it completely in an isolated context.

The user needs to get the Dockerfile somehow, which will probably be from cloning the HELICS repository (or downloading the all submodule archive). Then in the main folder for their copy of the HELICS repository they can run `docker build -t helics-test -f config/Docker/Dockerfile-HELICS .` to build the image.

I'm not sure that a completely standalone Dockerfile just for building the Docker image of the HELICS apps will get any use, since if they're a developer they will have cloned the Git repository, and if they are a user then they'll just pull a copy of an already built image from the Docker Hub registry.

nightlark:
I added an argument (settable using the docker build  `--build-arg` option) that will clone a copy from GitHub instead of using the copied directory. Setting `ENABLE_GITHUB=true` will turn it on, and the branch can be changed to a branch (or tag) other than develop by setting the `GIT_BRANCH` argument. It's probably best to run it from the `config/Docker` subfolder (or a folder with just the Dockerfile), to minimize the number of things copied into the build container since Dockerfiles don't support conditional COPY commands.

With the options in the same file we will be regular testing that the build commands still work, whereas I can easily imagine a separate Dockerfile for builds with the GitHub clone breaking without anyone noticing while the one run on Docker Hub gets fixed.

phlptp:
I was picturing it more as a building block that people would edit and modify to build there own images.  In which case I like your solution but maybe add some comments and a commented out line that does the clone in the docker file, and some more instructions in the docker file itself.   Also the docker.md page should be edited to reflect the changes and additional options.

nightlark:
What types of images do you have in mind? Would it be for someone who makes their own app like PiReceiver, NonlingsFed, or AnotherPowerSim, or is there a different target set of users?

phlptp:
Lets say we had an application that wanted to use HELICS and was containerized.  That would be built in a separate repository.  So I would start that from a HELICS dockerfile, then tweak it to add any of the other requirements, then have HELICS built and the application.  

The other base containers I could see  HELICS with the python interface prebuilt so someone could dump python code using HELICS into a container and have it ready to go,

Maybe something similar with Octave, julia, or Java 
At some level it is more about have a good base image which builds HELICS, for a container that can run the apps, but also for a base dockerfile that can be tweaks and modified as needs arise and as a good example for people who might do this in the future.

nightlark:
It sounds like what we want is another Docker image that gets built for releases/develop that has the shared libraries installed along with files required for interface bindings (or at least the swig files required to build a language interface).

For most cases I don't think the starting point should be a HELICS dockerfile that requires modification, and should instead be an already built image to use as a base by doing something like `FROM helics/helics-base:v2.2.0 as builder` in their own dockerfile.

If interfaces are needed for octave and python a `RUN apt-get install -y octave python` line could be added to install those programs and use the interface files already provided with the helics-base image (or run swig to generate language bindings using the copy of the swig interface files installed in the image). Maybe apt-get a version of gcc/clang that's compatible with the one used to build HELICS in the base image if they want to link with one of the C++ HELICS libraries. If we wanted to have images with the python interpreter, julia, java, or octave pre-installed, those could all use the helics-base image as a starting point and act as an example for users who want to build off of the base image we provide.

Then if an unusual case happens to crop up where someone absolutely needs to modify the way HELICS is compiled (special flags for running the clang sanitizer or odd compiler requirements) they could copy+paste the contents of the HELICS dockerfile into theirs as an additional stage and give it a different name (similar to the `as builder` part of what we have now).

phlptp:
Ok, I think the additional docker images should probably wait until more discussion occurs.  I suspect the views on needs and usefulness will change somewhat over the next few weeks.  For this PR lets add a few more comments in the existing docker file, and some additional instructions on the docker.md or other md file.  Then we can get this PR merged, and continue work on make the overall process better and figuring out what more we need.

PR #797, opened 2019-09-01
Warnings and 32 bit

phlptp:
### Summary

If merged this pull request will fix the 32 bit issue on drone and a several other static code analysis warnings from clang-tidy and others

### Proposed changes

- fix 32 bit issue with drone and other 32 bit compilation
- fix some warnings and possibly the undefined behavior warning we have been getting.

### DIFF
--- a/src/helics/application_api/ValueConverter_impl.hpp
+++ b/src/helics/application_api/ValueConverter_impl.hpp
@@ -205,7 +205,7 @@ void ValueConverter<X>::convert (const X *vals, size_t size, data_block &store)
 {
     detail::ostringbufstream s;
     archiver oa (s);
-    oa (cereal::make_size_tag (size));  // number of elements
+    oa (cereal::make_size_tag (static_cast<cereal::size_type>(size)));  // number of elements
     for (size_t ii = 0; ii < size; ++ii)
     {
         oa (vals[ii]);
--- a/src/helics/common/logger.h
+++ b/src/helics/common/logger.h
@@ -29,6 +29,9 @@ namespace helics
 {
 class LoggingCore;
 
+constexpr int always_log = -100000;  //!< level that will always log
+constexpr int log_everything = 100;  //!< level that will log everything
+
 /** class implementing a thread safe Logger
 @details the Logger uses a queuing mechanism and condition variable to store messages to a queue and print/display
 them in a single thread allowing for asynchronous logging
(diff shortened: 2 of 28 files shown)


src/helics/common/logger.h
```diff
@@ -29,6 +29,9 @@ namespace helics
 {
 class LoggingCore;
 
+constexpr int always_log = -100000;  //!< level that will always log
```
nightlark: One of the guidelines that appears some collections of hpc/scientific software library "collections" (RADIUSS at the lab, another one focused on what seemed to be scientific computation math libraries) was providing a way to silence output.

src/helics/core/ActionMessage.hpp
```diff
@@ -157,11 +157,11 @@ class ActionMessage
 
     int serializedByteCount () const
     {
-        int size = 45;
+        int size = 45;  // 45 is the base size of the message
```
nightlark: Is there a struct or class that this number can be computed from, or is it arbitrary?

src/helics/common/logger.h
phlptp: I am working with @eberleim on #758.  I want to revisit some of the logging mechanics and assumptions in that PR,  I think this discussion might belong in that PR about what policy we want to allow.  This was mainly about refactoring and cleaning up some clang-tidy warnings.   In general I agree there should be a way to shut off all output, but there is a natural tension between the different layers over user generated log messages.

src/helics/core/ActionMessage.hpp
phlptp: It is a partial subset of the ActionMessage that gets serialized every time, so it isn't arbitrary.  It probably should be made clearer as to the origins.

src/helics/application_api/ValueConverter_impl.hpp
```diff
@@ -205,7 +205,7 @@ void ValueConverter<X>::convert (const X *vals, size_t size, data_block &store)
 {
     detail::ostringbufstream s;
     archiver oa (s);
-    oa (cereal::make_size_tag (size));  // number of elements
+    oa (cereal::make_size_tag (static_cast<cereal::size_type>(size)));  // number of elements
```
phlptp: this is what actually fixed the 32 bit issues on drone

src/helics/core/ActionMessage.cpp
```diff
@@ -227,6 +227,32 @@ int ActionMessage::toByteArray (char *data, int buffer_size) const
...
+    {
+        size += static_cast<int> (3 * sizeof (int64_t));
+    }
+    // add additional string data
+    if (!stringData.empty ())
+    {
+        for (auto &str : stringData)
+        {
+            // 4(to store the length)+length of the string
+            size += static_cast<int> (str.size ()) + 4;
```
nightlark: It might be slightly clearer if the +4 was changed to `sizeof (uint32_t)` so it matches the type used in the serialization function, and flipping the order of things in the comment or the line below so that they match..

PR #801, opened 2019-09-05
Apps test to gtest

phlptp:
### Summary

If merged this pull request will convert the apps test to use gtest

### Proposed changes

- convert helics_apps test to use gtest instead of boost
- add some debugging output to an occasional failing test in tcpcore tests
- add apps_tests to circle ci build tests


phlptp:
I was wondering if the apps test would generate any new sanitizer issues.  Looks like it picked up a race condition.  I am not going to fix that in this PR, since it looks to be somewhere in the query mechanisms.

PR #803, opened 2019-09-09
Add documentation test to Travis

kdheepak:
### Summary

If merged this pull request will add documentation generation to travis builds as a separate job.


nightlark:
RTD added a feature for testing PRs that reports the result of doc builds with a link to the generated docs (see checks on https://github.com/readthedocs/readthedocs.org/pull/6160 for an example). I've contacted support to get the feature enabled for the helics RTD project - if it works I think it would be better than an additional job in the queue on other CI systems.

nightlark:
I think so, a doc build on Travis is mostly redundant given the RTD check.

PR #807, opened 2019-09-11
Binary only install

phlptp:
### Summary

If merged this pull request will add an option to only install the binary file (shared objects and executables)  no headers or static libs

### Proposed changes

- add binary only install option
- update a couple of the submodules to support this option
- update the docker file to use this option and add an apps  docker file to eventually replace the existing one so it is more clear what it does


CMakeLists.txt
```diff
@@ -294,7 +297,7 @@ if(GENERATE_DOXYGEN_DOC)
             ${HELICS_CURRENT_BINARY_DIR}/Doxyfile @ONLY
         )
         add_custom_target(
-            doc ${DOXYGEN_EXECUTABLE} ${HELICS_CURRENT_BINARY_DIR}/Doxyfile
+            helics_doc ${DOXYGEN_EXECUTABLE} ${HELICS_CURRENT_BINARY_DIR}/Doxyfile
```
nightlark: Maybe this target could be renamed to `helics_doxygen` (or something similar) to make it clear that it is just running doxygen rather than building all of the docs.

phlptp: I think I renamed it since it was conflicting with something else.   I am working on a more substantial doc update PR,  I will rename it in that PR to helics_doxygen, there were some other issues with this project that I am also resolving in that PR.

phlptp:
Do you think a header only install option would be useful as well or is that not useful in the context of dev packages?

nightlark:
Realistically, it would be easy to build everything and then make a package with just the include folder, since the bin/lib folders would be needed for other packages.

PR #812, opened 2019-09-17
Create and upload source archive with all submodules for releases

nightlark:
### Summary

If merged this pull request will add a GitHub Action that runs when a release is published (release + prerelease) to create and upload a source archive with the code for all submodules included.

Example release: https://github.com/nightlark/HELICS/releases/tag/v2.2.0.4

### Proposed changes
- Create separate bash-only and python versions of the git all archive script
- Create archive for the tagged git ref and upload to the published release

### DIFF
--- a/.github/workflows/release-all-submodules-archive.yml
+++ b/.github/workflows/release-all-submodules-archive.yml
@@ -16,7 +16,7 @@ jobs:
         echo $(jq --raw-output '.action' $GITHUB_EVENT_PATH)
         echo $(jq --raw-output '.release.upload_url' $GITHUB_EVENT_PATH)
     - name: Create archive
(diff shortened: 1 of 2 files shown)


phlptp:
will both of these versions run or is there a reason for both?

nightlark:
I'm only using the bash version, because I don't want to have to install any packages in the VM.

nightlark:
The other is the original version that runs a pip install and calls the python script Denis wrote.

phlptp:
will each workflow have a separate file in the workflow directory?  
Can the workflows use a docker image?

nightlark:
Originally I thought each would have a separate file, but it might end up being based on when the workflow runs (e.g. single file for releases). Steps can be run in the context of a docker image.

Issue #810, opened 2019-09-14
Add sha256 file to the single zip file


phlptp:
**Is your feature request related to a problem? Please describe.**
We need to generate a sha256 file associated with the single file zip so it can be verified by automated package generators.

**Describe the solution you'd like**
this should be done immediately on creation so it needs to be part of the script that generates the original file

nightlark:
Is the sha256 file a separate download? Do you have a link to an example of the file format that the automated package generator is looking for?

phlptp:
It is needed for a lot of package builders for [Example](https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-cython/PKGBUILD)

I think the Cmake format is fine.  The main purpose is as a reference to use when creating the build packages in the future and if anyone wants to verify the download.  And to make it legitimate I think the SHA code has to be generated at the point of creation and before upload.  

Is this something that can be added to #812 ?

nightlark:
I'll add it in a separate PR, I want to see if I can work out a way to get the hash for other uploaded assets to also be included in the file.

Issue #648, opened 2019-03-13
Automatic Installer generation


phlptp:
I think one next step is to have the windows installers (possibly others) automatically generated for the release candidates.  I think we are a good chunk of the way there, but I think we need a few configurations and some testing.  Possibly something we can do with Azure as appveyor seems a little slow.

kdheepak:
I'm already doing it on Travis and Appveyor. I can just add an extra step that builds the installer and we can push the final products to the releases page. 

https://travis-ci.com/GMLC-TDC/helics-conda

https://ci.appveyor.com/project/kdheepak/helics-conda-0bk88

phlptp:
I think it is mostly a matter of making sure we have the correct configurations generated and making a decision on what those configurations should be, then we can make sure those configurations are used on the CI builds and uploaded as appropriate.

For Windows
1. MSVC 2017 x64 Release with C++ Shared Lib
2. MSVC 2017 x86 Release
3. MSVC 2017 x64 Debug

4. MSVC 2015 x64 Release

5. Some build with static system libraries included and just the C shared library and headers

phlptp:
#759 is a step in this direction

nightlark:
Additional tasks related to this for the next minor release:
- [ ] macOS installer
- [ ] macOS, Linux, and Windows shared library archives for x86 and x64

nightlark:
Possibly set up future nightly builds.

PR #815, opened 2019-09-18
Build Windows installers and a SHA-256 hash file for releases

nightlark:
### Summary

If merged this pull request will add several Windows installer builds to releases using GitHub actions, and generate a SHA-256.txt file with hashes for the Windows installers and all submodules source code archive to address issue #810. Also related to #648, automatic installer generation.

VMs for Windows builds can take some time to start, and compiling takes ~10 minutes. Overall time after publishing a release for the files to appear is probably 15-30 minutes. Test releases: https://github.com/nightlark/HELICS/releases/tag/v2.2.0.9 and https://github.com/nightlark/HELICS/releases/tag/v2.2.0.11

### Proposed changes
- Build HELICS Windows installers (32-bit and 64-bit variants) using Visual Studio 2017 and 2019, with Python 3.7 interface
- Generate a SHA-256.txt file for HELICS releases with hashes for the Windows installers and the source code archive containing all submodules


nightlark:
Currently assets are uploaded to the release as soon as they are ready, and generating the SHA-256 file happens at the end with the artifacts from the previous steps. This could be changed to upload all of the assets to the release at the end.

There aren't VM images with VS2015, so it's just 2017 and 2019 installers built with the C and C++ shared libraries, static libraries, apps, and the Python 3.7 interface. I'm not sure how to deal with multiple Python versions without a large increase in the number of assets that need generating -- separate installers for different Python versions in 32/64-bit variants would be an approach.

If someone needs a debug build of the libraries, I think that might be a point at which they should build HELICS from source, or eventually with a package manager that allows switching compiler options.

phlptp:
what is the default python installed with VS2017 Release?  with VS2019?  and is the python included in these images the one with Visual studio or a separate installation.

What I had done before is VS2017 Release, VS2017 Debug,  VS2017 Release (static std lib) with exe and zip file.

I think we also need to answer questions of how we want to publish Matlab interface and Java interface potentially in isolation from the other libraries like the installers here.  

So as far as I can tell the PR looks good but a little discussion on what exactly we want as far as installers go might be warranted

nightlark:
I don't think there is any special Visual Studio version of Python. All of the versions are installed in C:/hostedtoolcache/windows/, the default one added to the PATH is 3.6; the other versions that can be added to PATH instead are 2.7, 3.4, 3.5, 3.6, and 3.7.

phlptp:
then it is probably worth it to build against 3.6 so it is most likely compatible with 3.6 and 3.7+

nightlark:
Set of files for releases based on discussion today:
- gzip file with all code
- VS2017 and VS2019 64-bit zip files with debug + release variants
- Generic HELICS installer (exe+zip) (64-bit) for executables and C shared library (static std lib) with Python 3.6 and Java 1.8 interfaces
- Zip with C shared library with C/C++98 headers (and swig interface files)
- Matlab interface

phlptp:
@kdheepak  What are you thoughts on the above list?

phlptp:
@nightlark are you planning to update this more, or do we want to merge this and work on the rest of the list in the 2.3 release?

kdheepak:
I would like a tar with C shared library with C/C++98 headers and no swig interfaces.

nightlark:
@kdheepak would the one with the swig interface files work? It just has the `helics.i` and `output.i` files in the `share/helics/swig` subfolder, none of the language specific interfaces or files.

phlptp:
@kdheepak  what platform?

kdheepak:
> would the one with the swig interface files work? It just has the helics.i and output.i files in the share/helics/swig subfolder, none of the language specific interfaces or files.

Okay, this works! I thought it had the compiled files. 

I'd like this for all platforms, Windows, Mac and Linux; with x86 and x64.

nightlark:
If we aren't releasing 2.2.1 today, then I'll change the set of uploaded installers from the current msvc ones to these:
- VS2017 and VS2019 64-bit zip files with debug + release variants
- Generic HELICS installer (exe+zip) (64-bit) for executables and C shared library (static std lib) with Python 3.6 and Java 1.8 interfaces
- Zip(/tar) with C shared library with C/C++98 headers (and swig interface files); Windows x86/x64

Archives with the shared library for other platforms will likely be for the release after.

phlptp:
I want to get the broker timeout in yet, and that will hopefully get finished over the weekend, so the most realistic target would be next Tuesday for a release.

PR #817, opened 2019-09-18
fix for the libzmq on mingw

phlptp:
### Summary

If merged this pull request will fix the failed build on mingw (and MSYS)

### Proposed changes

- revert the zmq library for MINGW and MSYS to 4.3.1,  This seems to be a failure with libzmq, though still investigating.


phlptp:
There is a fix in the libzmq master for this but I don't want to move libzmq to master so for now MINGW builds will hold the 4.3.1 release until a new release.

PR #818, opened 2019-09-20
Find python better

phlptp:
### Summary

If merged this pull request will remove the use of FindPython as it could not guarantee consistent results with the old method and using them together produced numerous problems.   In the future it might be practical to switch completely to the newer find but not until the minimum cmake version is upgraded. 

### Proposed changes

- switch methods for building finding python libraries


CMakeLists.txt
```diff
@@ -519,7 +519,13 @@ if("${isSystemDir}" STREQUAL "-1")
 endif()
 
 # add the local directory to the rpath
-list(APPEND CMAKE_INSTALL_RPATH "./")
+
+	if(APPLE)
+		list(APPEND CMAKE_INSTALL_RPATH "@loader_path/.")
+		list(APPEND CMAKE_INSTALL_RPATH "@executable_pat/.")
```
```suggestion
		list(APPEND CMAKE_INSTALL_RPATH "@executable_path/.")
```

interfaces/python/CMakeLists.txt
```diff
@@ -10,10 +10,91 @@
...
+	if(${CMAKE_MAJOR_VERSION}.${CMAKE_MINOR_VERSION} VERSION_GREATER 3.12)
+		option(HELICS_USE_NEW_PYTHON_FIND "Use the FindPython Routine only available from cmake 3.12 and later" OFF)
+		mark_as_advanced(HELICS_USE_NEW_PYTHON_FIND)
+	else()
+		set(HELICS_USE_NEW_PYTHON_FIND OFF CACHE BOOL "" FORCE)
+		hide_variable(HELICS_USE_NEW_PYTHON_FIND)
+	endif()
+ 
+ if (HELICS_USE_NEW_PYTHON_FIND)
+	Find_package(Python3 COMPONENTS Interpreter Development)
```
nightlark:
```suggestion
	find_package(Python3 COMPONENTS Interpreter Development)
```

PR #819, opened 2019-09-20
Disable static lib install

nightlark:
### Summary

If merged this pull request will add an option to disable installing the static libraries (helics-static and helics_app), and only installs C++ headers if at least one of the C++ libraries (shared/static) is installed.

### Proposed changes
- Add an option to disable installing static libraries, mostly for a non-MSVC Windows installer
- Only install C++ headers if either the shared or static libraries will be installed


phlptp:
the ASAN failure is interesting but it looks like a ~1/1000 test failure caused an assert which terminated the test causing the callback variable to go out of scope before it was finished being used  by the cores.  I just fixed it in #818, but I am guessing it wouldn't show up if ran again.

CMakeLists.txt
```diff
@@ -124,6 +124,7 @@ mark_as_advanced(HELICS_WITH_CMAKE_PACKAGE)
 option(HELICS_BINARY_ONLY_INSTALL "only install the helics binary executables" OFF)
 mark_as_advanced(HELICS_BINARY_ONLY_INSTALL)
 
+option(DISABLE_STATIC_LIB_INSTALL "Disable installing of static libraries" OFF)
```
phlptp:
```suggestion
option(DISABLE_STATIC_LIB_INSTALL "Disable installing of static libraries" OFF)
mark_as_advanced(DISABLE_STATIC_LIB_INSTALL)
```

CMakeLists.txt
```diff
@@ -805,28 +811,30 @@ option(HELICS_INSTALL_CEREAL "Install a copy of the cereal headers used for comp
...
-)
-add_library(cereal INTERFACE)
-add_library(HELICS::cereal ALIAS cereal)
-target_include_directories(cereal
-    INTERFACE
-        $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/ThirdParty>
-        $<INSTALL_INTERFACE:$<INSTALL_PREFIX>/include/helics/external/optional>
-)
-install(TARGETS cereal ${HELICS_EXPORT_COMMAND} COMPONENT cereal)
+if (BUILD_SHARED_LIBS OR NOT DISABLE_STATIC_LIB_INSTALL)
```
phlptp: I am not sure BUILD_SHARED_LIBS is the right variable to use here,  I thought there was some cases where I turned it off after enabling another variable to prevent it from being used in the submodules since some of them use it as well.

nightlark: I don't think there is an alternative that would work at the moment. How about a variable like `HELICS_INSTALL_CPP14_HEADERS` that gets set when `BUILD_SHARED_LIBS` is used to enable the CXX shared library in the main CMakeLists.txt?

phlptp: I think it is OK as is,  I remember running into some trouble with the on the jsoncpp library but I explicitly disabled inside the cmake for adding jsoncpp.

So I think you are using the appropriate variable for checking if the C++ shared library is enabled.

phlptp:
you are having bad luck with the sanitizer tests on this PR

nightlark:
I added a check earlier on that sets HELICS_INSTALL_CPP14_HEADERS based on the result of the longer condition, and replaced the places that were installing headers with a check for this new variable instead (conveys the intent a bit more clearly).

PR #823, opened 2019-09-24
Broker timeout

phlptp:
### Summary

If merged this pull request will add a timeout on the broker if comms are lost with a core.  This will either assume disconnection or generate a federation error.  

### Proposed changes

- add lost comms detection on the broker and the ability to ping broker connections.


tests/helics/system_tests/brokerTimeoutTests.cpp
```diff
@@ -0,0 +1,85 @@
...
+
+    auto cr = Fed2->getCorePointer ();
+    auto tcr = std::dynamic_pointer_cast<helics::testcore::TestCore> (cr);
+    EXPECT_TRUE (tcr);
+
+    auto cms = tcr->getCommsObjectPointer ();
+    cms->haltComms ();  // this will terminate communications abruptly
+    tcr.reset ();
+
+    bool val = brk->waitForDisconnect (std::chrono::milliseconds (2000));
```
nightlark: Does the time it takes to disconnect after the timeout much larger than 200ms, or does it vary significantly?

phlptp: It seems to vary quite a bit.  The tick only gets passed if no data has been sent, and that only occurs on the tick timeouts which rely on an independent thread operating.  Then it has to send the pings and wait for a sufficient number of ticks to pass, so the real timeout time is quite a bit longer than the timeout specified.  There are probably things we can do to refine it but it also is limited to number of threads on a CI system.  For it to function as a tick it really has to be independently operating and I don't want it so frequent that it accounts for a decent amount of processing.    I think we will refine it as we get more experience in it.  But I think this PR handles some of the most common scenarios with timeouts.

PR #827, opened 2019-09-26
Shared lib test to gtest

phlptp:
### Summary

If merged this pull request will convert the shared library test to use gtest and add the tests to circle ci.


phlptp:
My intention is to tackle the application_api tests soon after this is merged, then we can remove another chunk of the boost configuration

PR #829, opened 2019-09-29
App api to gtest

phlptp:
### Summary

If merged this pull request will convert the application_api_test to use google test

### Proposed changes

- convert application api tests to use google test and unify all uses of ci_skip in the different tests.


phlptp:
These tests exposed a race condition see Issue #830.

That race condition will be fixed in a separate PR since it was not introduced by this PR merely exposed.

phlptp:
There is one more test that gets built using boost test,  There is another longer running branch that converted it but there were some other issues preventing that from being merged.  So that will need to be investigated.  But we are getting close to eliminating boost tests and all the associated machinery

PR #831, opened 2019-09-30
Add benchmark tests

phlptp:
### Summary

If merged this pull request will add google benchmarks as a package and 1 benchmark test
and move google test to be a git clone instead of a submodule.  

### Proposed changes

- add google benchmark as git clone or fetch content
- change google test to use fetch Content or git clone
- add a benchmark about conversion of the ActionMessage

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -907,26 +907,32 @@ set(LICENSE_LIST
 	"GMLC Utilities" "${PROJECT_SOURCE_DIR}/ThirdParty/utilities/LICENSE"
 )
 
-if(BUILD_HELICS_TESTS)
 if(NOT CMAKE_VERSION VERSION_LESS 3.11)
+
+if(BUILD_HELICS_TESTS)
 	list(APPEND LICENSE_LIST "Google Test" "${PROJECT_BINARY_DIR}/_deps/googletest-src/LICENSE")
-	else()
-	list(APPEND LICENSE_LIST "Google Test" "${PROJECT_BINARY_DIR}/_deps/googletest/LICENSE")
-	endif()
 endif()
 
 if(HELICS_BUILD_BENCHMARKS)
-if(NOT CMAKE_VERSION VERSION_LESS 3.11)
 	list(APPEND LICENSE_LIST "Google Benchmark" "${PROJECT_BINARY_DIR}/_deps/gbenchmark-src/LICENSE")
-	else()
-	list(APPEND LICENSE_LIST "Google Benchmark" "${PROJECT_BINARY_DIR}/_deps/gbenchmark/LICENSE")
-	endif()
 endif()
 
 if (EXISTS ${PROJECT_BINARY_DIR}/_deps/libzmq-src)
 	list(APPEND LICENSE_LIST "ZeroMQ" "${PROJECT_BINARY_DIR}/_deps/libzmq-src/COPYING")
-elseif (EXISTS ${PROJECT_BINARY_DIR}/_deps/libzmq)
+endif()
+
+else()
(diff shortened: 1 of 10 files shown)


phlptp:
This is a next step to completely remove boost tests.   The step after this is to move some of the perf tests to use google benchmarks

CMakeLists.txt
```diff
@@ -683,6 +683,16 @@ if(BUILD_HELICS_TESTS AND BUILD_TESTING AND ENABLE_TEST_CORE)
     add_subdirectory(tests)
 endif()
 
+option(
+  HELICS_BUILD_BENCHMARKS
```
nightlark: This option seems similar to the one for building tests -- could we rename it to `BUILD_HELICS_BENCHMARKS` so CMake groups the options together?

README.md
```diff
@@ -137,6 +137,12 @@ A list of optional component that are not included in HELICS but are optionally
 #### [zmq](http://www.zeromq.org)
   ZeroMQ message passing is used in the ZMQ core networking.  The automatic download currently uses version 4.3.1. No ZMQ library code is included in the HELICS source.  HELICS installers include ZeroMQ binary libraries. ZeroMQ is licensed under [LGPL](https://github.com/zeromq/libzmq/blob/master/COPYING.LESSER) with a modification to allow for linking and in various forms and distribution of the binary under different terms if the library was not modified.  No modification of the ZMQ library or any of the ZeroMQ source files is included in the HELICS source code.  It is an optional component and can be excluded by setting  `ENABLE_ZMQ_CORE` to `OFF`
 
+### [Google Test](https://github.com/google/googletest)  
+  Helics test are written to use the Google Test and mock frameworks. Google Test is included in the source tarbal but is downloaded as an optional component. Google Test is released with a [BSD-3 clause](https://github.com/google/googletest/blob/master/LICENSE) license.
```
nightlark:
```suggestion
  HELICS tests are written to use the Google Test and mock frameworks. Google Test is included in the source tarball but is downloaded as an optional component. Google Test is released with a [BSD-3 clause](https://github.com/google/googletest/blob/master/LICENSE) license.
```

README.md
```diff
@@ -137,6 +137,12 @@ A list of optional component that are not included in HELICS but are optionally
...
+
+### [Google Benchmark](https://github.com/google/benchmark)  
+ Some timing benchmarks with HELICS are written to use the Google Benchmark library. Benchmarks is an optional component and is not included in the main source tarball and must be downloaded separately. Google Benchmark is released with a [Apache 2.0](https://github.com/google/benchmark/blob/v1.5.0/LICENSE) license.
```
nightlark:
```suggestion
 Some timing benchmarks with HELICS are written to use the Google Benchmark library. Benchmarks is an optional component and is not included in the main source tarball and must be downloaded separately. Google Benchmark is released with an [Apache 2.0](https://github.com/google/benchmark/blob/v1.5.0/LICENSE) license.
```

CMakeLists.txt
phlptp: Eventually I want to get all the helics cmake options to look like HELICS_XXXX so they are "namespaced"  I don't think we will be able to do that completely until HELICS 3.0 but any new option I have been trying to put into that form.  I suppose I could change it the BUILD_HELICS form again.    But not sure how that would impact things later on or if it doesn't matter much.

phlptp: I am going to merge this now, and I have another branch working with some of the cmake variables that should come in the near future.  if we decide we want to change the variable to BUILD_HELICS_BENCHMARKS we can do it then before the next release.

PR #834, opened 2019-10-02
Interface gen docker

phlptp:
### Summary

If merged this pull request will include docker files for interface generation and a single sanitizer image with all the sanitizers information contained in it.  

### Proposed changes

- single docker image for all sanitizers
- tell circle ci to use the new image for the tests
- create a dockerFile for for generating interface files.  
- remove existing files for when overwriting the interface generation files


phlptp:
I will try to fix some of these issues in another PR, but that will likely not be very quick to fix the issues exposed by the updated Thread sanitizer.

phlptp:
I am not going to merge this yet,  I would like @nightlark  to take a look at the interface gen Docker image and script and see if there are any changes required or to be added

nightlark:
I tried running a job in a docker container, based on how running steps seems to work I think there are some steps that would make sense to shift from the dockerfile to a workflow step -- I'll take a closer look tomorrow morning and comment with some changes.

phlptp:
I think I want to wait a few days to merge this one until after some of the others, since I want to check if the previous TSAN errors were fixed first before updating to the new one, which seemed to add a bunch more detections of potential errors.

phlptp:
@nightlark  can you take a look at this again before I merge it?

PR #837, opened 2019-10-04
Klockwork and sanitizer fixes

phlptp:
### Summary

If merged this pull request will resolve a number issues (many probably not that important) from klokwork code analyzer, and fix a race condition from [thread Sanitizer](https://circleci.com/gh/GMLC-TDC/HELICS/1415)

### Proposed changes

- klockwork fixes
- race condition fix by moving the detection of subscribers into the actual call vs the core loop, then generating all the messages in the setValue call instead of in the core loop.  Doing it in the core loop had potential race conditions and if we try to lock there to prevent it there was lock potential and confusing data paths that didn't quite work.  The race condition itself was pretty benign but could have led to an incorrect federate receiving some data or missing the first data publication.  
- clean up the locking call on FederateState to use more lock_guard operations
- make the lock() function call the sleeplock which spins for a little while before sleeping now


phlptp:
This does fix the Thread Sanitizer issue,  and PR #836 fixes the octave issue so hopefully on one of these PR's we can see the Circle CI tests all pass before we break them again.

src/helics/apps/helics-broker-server.cpp
```diff
@@ -277,6 +276,11 @@ void terminalFunction (std::vector<std::string> args)
     {
         std::string cmdin;
         std::cout << "\nhelics_broker_server>>";
+        if (cmdin == "quit()")
```
nightlark: Do any of the other commands use parenthesis?

phlptp: no, this was mainly a way to make klockwork happy,  and just in case the app screwed up and you wanted a way to abort.    Really just needed a short circuit path that bypassed the CLI app and quit immediately.      Open to other suggestions.

nightlark: I think for consistency with other commands it should just be "quit".

phlptp: I changed it to "exit" || "q"  and actually made it after the line that gets the command from the user.

PR #844, opened 2019-10-09
Add additional benchmarks

phlptp:
### Summary

If merged this pull request will add additional benchmark code

### Proposed changes

- update the echo benchmark and ring benchmark to use one thread in the main benchmark loop
- add conversion benchmarks for converting different types
- add messageLookup benchmark for message broadcasts
- add different ring size benchmarks and simplify the test code
- move the ping for keep-alive to a priority channel

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -104,6 +104,7 @@ jobs:
         - MATRIX_EVAL="COMPILER=clang && CC='clang-5.0' && CXX='clang++-5.0'"
         - CCACHE_CPP2=yes
         - USE_SWIG=true
+        - BUILD_BENCHMARKS=true
         - USE_CMAKE_VERSION=3.14.5
         - CXX_STANDARD=17
 
--- a/benchmarks/helics/ActionMessageBenchmarks.cpp
+++ b/benchmarks/helics/ActionMessageBenchmarks.cpp
@@ -42,3 +42,104 @@ static void BM_AM_FromString (benchmark::State &state)
 }
 // Register the function as a benchmark
 BENCHMARK (BM_AM_FromString);
+
+static void BM_AM_toString_time (benchmark::State &state)
+{
+    ActionMessage obj (CMD_TIME_REQUEST);
+    std::string load;
+    load.reserve (500);
+    for (auto _ : state)
+    {
+        obj.to_string (load);
+    }
+}
+// Register the function as a benchmark
+BENCHMARK (BM_AM_toString_time);
+
+static void BM_AM_FromString_time (benchmark::State &state)
+{
+    ActionMessage obj (CMD_TIME_REQUEST);
+    std::string load;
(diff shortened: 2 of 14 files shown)


phlptp:
One goal is to have a moderate set of included benchmarks ready for 2.3 that can be used for a baseline.

So @nightlark @trevorhardy @jcfuller1 @bpalmintier @kdheepak  and anyone else,  What benchmarks should we have.    Keep in mind these should be ones that can be run on a single machine.  We are planning multimachine benchmarks, but that will require a different mechanism.  This is about benchmarks that can be built and run on any machine in a local fashion.  So this can be system  tests or tests of a particular function or aspect of HELICS.  

I am planning on adding a few more not necessarily to this PR but in another in the next couple weeks.

trevorhardy:
@phlptp, can you clarify the purpose of the benchmarks, particularly with respect to some of the characterization test we've been working on [over in this related repository](https://github.com/GMLC-TDC/HELICS-Characterization-Tests)? I'm assuming you're after something different with the benchmark testing.

phlptp:
These tests are solely ones that can run on a single system in a single process,  
So what I have in now are some microbenchmarks on internal message serialization, and data conversions.
Also an echo test, which is probably similar to the characterization tests,
a ring test which is basically a data token going around and around a ring of federates a bunch of times.
and a random message test which generates some number of endpoints in some number of federates and then a certain amount of random source and destinations which send the messages.

So these benchmarks are single process, and can be built and run wherever you build HELICS.  Meant for testing/benchmarking different aspects or small test cases.    My expectation is that it would be a complement to the characterization library just something much easier to setup and run.

phlptp:
I am going to merge this but if anyone has any suggestions for additional local benchmarks let me know.

trevorhardy:
Other benchmark ideas:

1. Running an additional test based on your message one that implements a few basic filters (delay, drops) with random parameters.
2. Messages of increasing size (I've done orders of magnitude in the past) up to some limit.
3. Something to try to gum up the buffer size on the network port? (May or may not be a good idea but gives us some visibility about how the OS's handle this stuff).
4. A large number of messages (may be redundant with above).

Doing these things across all the core types?

bpalmintier:
Where is the right place to put in something basic around total data exchange time overhead? That is how long does it take to pass data around the federation. Seems this could be a function of the number federates, the comm backend, the number and types of subscriptions, number of messages, etc. Perhaps this is just adding timing to some of the current benchmark ideas, but might also be a somewhat different set of scenarios.

phlptp:
I think to some extent many of the benchmarks are trying to get at this concept.  There are a large number of different ways data can be communicated so it is not a simple single number to get at that timing.    The ring benchmark is probably the closest.  That is passing a data token around a federation in different configurations a certain amount of times and that is pretty much all it is doing.

Issue #498, opened 2018-10-24
Include commit hash in HELICS version string


nightlark:
**Is your feature request related to a problem? Please describe.**
The HELICS version number and date strings rely on manually updating the CMakeLists.txt file for new releases. For build artifacts from CI, distinguishing which commit it originated from (ex: in a log file) is impossible for later reproducing the build.

**Describe the solution you'd like**
Add an option to include a commit hash as part of the version string (to the HELICS 2.0 branch).

phlptp:
This is the same as #376.    The real question is how to do we do this but only have the regular release number show up for any users?

kdheepak:
This is what I've used in the past and it reads well I think.

```
    const version_string = compile_time_exec("git describe --tags HEAD", DEFAULT_VERSION_NUMBER)
```

nightlark:
The compile_time_exec solves the problem with a cmake-only solution of ensuring that the version string is up to date if new commits get pulled without re-running cmake.

PR #845, opened 2019-10-09
Cmake update

phlptp:
### Summary

If merged this pull request will do some refactoring of the cmake 

### Proposed changes

- add a git hash based description for builds not of one of the tagged releases
- reformat some of the cmake code
- update cmake with some more modern best practices.
- test are off by default
- examples are off by default
- all references to HELICS_BOOST_TESTS have been removed since they are no longer present
- change a few more variables and options in CMAKE to have HELICS_ prefix


phlptp:
Things to do yet 
- [ ] make the <HELICS>/src directory able to generate as the root directory in cmake
- [ ] reformat all the cmake file ( I don't want to do this until the PR is reviewed since it will introduce a lot of changes that probably won't be meaningful
- [ ] add ability to generate both helics2 and helics3 in the same build.  (Is this useful?)

nightlark:
Does "generate both helics2 and helics3 in the same build" mean HELICS 2.0 and the future 3.0 versions in the same build folder?

CMakeLists.txt
```diff
@@ -193,89 +197,93 @@ if("${PROJECT_SOURCE_DIR}" STREQUAL "${PROJECT_BINARY_DIR}")
...
-    option(EMBEDDED_DEBUG_INFO "embed debug info into obj files" OFF)
-    mark_as_advanced(EMBEDDED_DEBUG_INFO)
-    if(EMBEDDED_DEBUG_INFO)
-        ucm_set_embedded_debug(EMBEDDED)
-    else()
-        ucm_set_embedded_debug(EXTERNAL)
-    endif()
-endif(MSVC)
-endif (HELICS_ROOT_PROJ)
+mark_as_advanced(EMBEDDED_DEBUG_INFO)
```
```suggestion
mark_as_advanced(HELICS_EMBEDDED_DEBUG_INFO)
```

config/cmake/version_describe.cmake
```diff
@@ -0,0 +1,45 @@
...
+		string(FIND ${TAG_DESCRIPTION2} "-dirty" dirty_loc)
+		if (dirty_loc GREATER 0)
+			string(SUBSTRING ${TAG_DESCRIPTION2} 0 ${dirty_loc} tag_desc)
+			string(APPEND tag_desc ${hash_string} "-dirty")
+		else()
+			set(tag_desc ${TAG_DESCRIPTION2})
+			string(APPEND tag_desc ${hash_string})
+		endif()
+		
+		string(FIND ${tag_desc} "heads/" hloc)
```
nightlark: Should this get done before appending "-dirty" to the tag description?

interfaces/CMakeLists.txt
```diff
@@ -7,21 +7,21 @@
 # SPDX-License-Identifier: BSD-3-Clause
 #
 
-if (NOT DISABLE_C_SHARED_INTERFACE)
```
nightlark: Where is DISABLE_C_SHARED_INTERFACE defined? In the main CMakeLists.txt file I was only finding HELICS_DISABLE_C_SHARED_LIB.

config/cmake/version_describe.cmake
phlptp: I don't think it matters as heads would be at the beginning, and I can't say I am particularly concerned with the performance of the string processing algorithms in cmake

interfaces/CMakeLists.txt
phlptp: this should be fixed, I will fix it in the next update

kdheepak:
> generate both helics2 and helics3 

I think @phlptp means Python2 and Python3 in the same build.

kdheepak:
Generating Python2 and Python3 libraries in the same build is only useful for creating .exe files or .dmg files. From conda's perspective we will need only one at a time, and even if we generate both, only one should be copied to the output. I think it's likely that that is going to cause more pain than add value, since it'll need Python2 and Python3 libraries installed in the same environment. I'd avoid it unless cmake has very good support for it already.

config/cmake/version_describe.cmake
```diff
@@ -0,0 +1,52 @@
...
+			string(SUBSTRING ${tag_desc} 6 -1 tag_desc)
+	    endif()
+	else()
+	   set(tag_desc "")
+	endif()
+
+
+	set(${result} ${tag_desc} PARENT_SCOPE)
+  else(GIT_FOUND)
+	set(${result} "" PARENT_SCOPE)
```
kdheepak: Use `"noversion"` or `"source"` instead of empty string?

nightlark: Empty string is equivalent to a false value in CMake. `NOTFOUND` or something suffixed with `-NOTFOUND` like `GITVERSION-NOTFOUND` should make it work the same for existing comparisons.

kdheepak: I was suggesting setting the result as `tarball` or `built-from-source` to indicate that it was built from the source tarball.

nightlark: If it was built from the source tarball then that implies they got it from a tagged release and the version number should be accurate. I think most versions built by package managers (spack, vcpkg, conda-forge) would then end up with a version string of "2.0.0-noversion" or "2.0.0-source" since they'll likely be pulling the all submodules tarball for building.

phlptp: The intention was for there to be no additional string if the build was a tagged version of HELICS, so I was thinking empty, I suppose NOTFOUND would be equivalent, but I am kind of inclined to leave it empty.  Over time if the cmake function proves useful we might think about adding more options to it and making a little more flexible.

phlptp:
@nightlark  can you take a look at the package tests on travis gcc 4.9 I am not entirely sure what the problem is currently or where to look to fix it.

phlptp:
I am not going to add ability for generating Python2 and Python3 simultaneously.

phlptp:
It seems the changes to CMAKE and renaming a few variables "fixed" a few of the tests that should have been active but were not.  This in turn exposed a few bugs that have likely been present for quite a while from tests that were intended to be run but weren't being run.  Mainly in the sub_project_tests.  So I am working through the process to try to figure out what is happening and how to fix them, so they all pass. 

Also the changes activated a number of warnings in visual studio that were thought to be on, but due to unknown paths in CMAKE they weren't actually being enabled.  Cleaning up the CMAKE file, let them activate and now we got a bunch of warnings,  most were innocuous but a couple were actual bugs in "unusual paths that have probably never been used"  but were bugs nonetheless.  

So this is probably a good thing, but it is causing some headaches in the short term with this branch.

phlptp:
@nightlark @kdheepak  I think this is pretty much ready now.  I switched nearly all the options Except BUILD_XXX_INTERFACE,  ENABLE_XXX_CORE.  Those will wait to change over until python 3 since those are possibly higher use.  I think I changed most of the documentation, but there may be some gaps yet.    On Monday I will run all the CMAKE through a formatter to clean it up, before merging, so if you want to take another look and see if there is anything I missed before tomorrow afternoon and a I make a whole bunch of formatting changes, and merge soon after.

phlptp:
[HERE](https://external-builds.readthedocs.io/html/helics/845/installation/helics_cmake_options.html) is a new page on all the CMAKE options

phlptp:
Assuming everything passes, I am going to merge this tomorrow afternoon,  @kdheepak, @nightlark  take another look.  I think it cleans up and organizing things a lot for the CMake code.

phlptp:
This will close issues #498 and #499

Issue #848, opened 2019-10-17
Benchmark types


phlptp:
Based on conversation from PR #844 

- [x] Running an additional test based on your message one that implements a few basic filters (delay, drops) with random parameters.(#880)
- [x] Messages of increasing size (I've done orders of magnitude in the past) up to some limit. (@nightlark )
- [-] Something to try to gum up the buffer size on the network port? (May or may not be a good idea but gives us some visibility about how the OS's handle this stuff). -- I don't think this will be done in 2.3,  it may be more valuable once some of the network testing is accomplished.  
- [x] A large number of messages (may be redundant with above). (@nightlark -small variant on 2nd item)
- [x] Echo structure like the echo one for pub/sub.

phlptp:
If people think of any more lets add them to the list.  The goal is to have these ready to go by HELICS 2.3 release.

phlptp:
Added some additional benchmarks in #892 for HELICS 2.4

PR #849, opened 2019-10-17
add echo Message Benchmark tests

phlptp:
### Summary

If merged this pull request will add an echoMessage Benchmark

### Proposed changes

- add echo message Benchmark
- add overload of send in Endpoint Object that uses unique_ptr<Message>


phlptp:
One part of Issue #848

Issue #853, opened 2019-10-23
Got disconnect from TestA Federate(1879048192)


kdheepak:
**Describe the bug**

When a federate is performing some domain task for too long, the HELICS federate sends a disconnect to the broker.

**What is the expected behavior?**

The timeout for performing some domain task must be made larger or the error message must be made clearer. It is not obvious to the user why the federate has been disconnected from the broker.

**To Reproduce**

```julia
# test1.jl
using HELICS
const h = HELICS

broker = h.helicsCreateBroker("zmq", "", "-f 1 --loglevel=7")

while h.helicsBrokerIsConnected(broker)
    sleep(5)
end

h.helicsBrokerFree(broker)
h.helicsCloseLibrary()
```

```julia
# test2.jl
import HELICS
const h = HELICS

fedinfo = h.helicsCreateFederateInfo()
h.helicsFederateInfoSetCoreName(fedinfo, "TestA Federate")
h.helicsFederateInfoSetCoreTypeFromString(fedinfo, "zmq")
h.helicsFederateInfoSetCoreInitString(fedinfo, "--federates=1")
h.helicsFederateInfoSetTimeProperty(fedinfo, h.HELICS_PROPERTY_TIME_DELTA, 1)
vfed = h.helicsCreateValueFederate("TestA Federate", fedinfo)
pub = h.helicsFederateRegisterGlobalTypePublication(vfed, "testA", "double", "")
sub = h.helicsFederateRegisterSubscription(vfed, "testA")
h.helicsFederateEnterExecutingMode(vfed)

while true
    sleep(5)
end
```

**Environment (please complete the following information):**
 - Operating System: [all] ( tested on a Mac )
 - Language Extension: [all] ( tested in Julia )
 - HELICS version: 

```
$ ./deps/usr/bin/helics_app --version
2.2.1 (09-27-19)
```

**Additional context and information**

```
66424-OsBwF-idWka-ek7og-JA7QY (0)::connecting
66424-OsBwF-idWka-ek7og-JA7QY (0)::||connected on tcp://127.0.0.1:23404
66424-OsBwF-idWka-ek7og-JA7QY (1)::|| priority_cmd:unknown -1 from 0
66424-OsBwF-idWka-ek7og-JA7QY (1)::|| cmd:protocol:From 0 from 0 to 0
66424-OsBwF-idWka-ek7og-JA7QY (1)::|| cmd:tick:From 0 from 0 to 0
[...]

PR #856, opened 2019-10-24
Add timeout test

kdheepak:
### Summary

If merged this pull request will add a new timeout test in Python for a value federate. It initializes a federate, enters execution and waits 60 seconds, and then checks if the broker is still connected.

This test can be run explicitly by using the following:

```
pytest -sv tests/python/test_value_federate.py -k test_value_federate_runFederateTimeoutTest
```

Related to #853

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,8 +10,8 @@ A note on future revisions.
 Bug fix release
 
 ### Fixed
--  links in the README changed with an automated move to travis-ci.com
--  Issue #853,  which was causing core connections to timeout if no direct communication was observed for a period of time.  This bug fix release fixes that issue where the pings were not being correctly accounted for in the timeout detection. 
+-  Links in the README changed with an automated move to travis-ci.com
+-  Fix issue #853, which was causing core connections to timeout if no direct communication was observed for a period of time.  This bug fix release fixes that issue where the pings were not being correctly accounted for in the timeout detection. 
 - Fix Ctrl-C issue when using HELICS in some language api's
 
 ## \[2.2.1\] ~ 2019-09-27
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,13 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 A note on future revisions.  
   Everything within a major version number should be code compatible (with the exception of experimental interfaces).  The most notable example of an experimental interface is the support for multiple source inputs.  The APIs to deal with this will change in future minor releases.  Everything within a single minor release should be network compatible with other federates on the same minor release number.  Compatibility across minor release numbers may be possible in some situations but we are not going to guarantee this as those components are subject to performance improvements and may need to be modified at some point.  Patch releases will be limited to bug fixes and other improvements not impacting the public API or network compatibility.  Check the [Public API](./docs/Public_API.md) for details on what is included and excluded from the public API and version stability.
+## \[2.2.2\] ~ 2019-10-27
+Bug fix release
+
+### Fixed
+-  links in the README changed with an automated move to travis-ci.com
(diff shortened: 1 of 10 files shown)


nightlark:
Does pytest have a way to control which tests are run similar to GoogleTest filters or Boost.Test labels?

kdheepak:
Yes, using command line arguments. The example I shared above runs just one test. If you use `pytest -sv tests/python/test_value_federate.py` instead, it runs just the tests in that file. Is that what you are asking?

nightlark:
I was thinking for the opposite case, to run everything except the listed test.

kdheepak:
There's a way to mark specific tests as ones that will be skipped or ones that will fail.

https://docs.pytest.org/en/latest/skipping.html

We have to do it in the code though. It's a one line addition to can mark this as one that should fail.

nightlark:
I think it's one that should be marked as skipped, then in the cmake a new add_test can be added that runs this specific test with some label that can separate it from faster tests.

phlptp:
I can easily see a number of this type of tests being needed in the future.  There should be a way to disable them easily.  Or have them be disabled by default an only run in some of the CI builds.

kdheepak:
If this `HELICS_PYTHON_ADDITIONAL_TESTS` is set to any value except an empty string, this test will run. `export HELICS_PYTHON_ADDITIONAL_TESTS=""` is the default and will result this test not being run.

In the future we can add additional conditions, e.g. if `export HELICS_PYTHON_ADDITIONAL_TESTS="value,message"`, run the additional value and message federate tests but not the additional filter tests.

kdheepak:
I'm getting this error when I try to build this branch from source:

```
[ 43%] Linking CXX static library libhelics-static.a
[ 43%] Built target helics-static
tests/helics/performance_tests/CMakeFiles/perf-tests.dir/build.make:115: *** target pattern contains no `%'.  Stop.
make[1]: *** [tests/helics/performance_tests/CMakeFiles/perf-tests.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
tests/helics/application_api/CMakeFiles/application-api-tests.dir/build.make:325: *** target pattern contains no `%'.  Stop.
make[1]: *** [tests/helics/application_api/CMakeFiles/application-api-tests.dir/all] Error 2
```

I used `cmake -DBUILD_PYTHON_INTERFACE=ON -DCMAKE_INSTALL_PREFIX="/Users/$USER/local/helics-develop/" ../`. 

I added `-DBUILD_HELICS_TESTS=OFF` to make the above error go away.

And I was able to run the test and verify that it is in fact now resolved on this branch!

```
pytest -sv tests/python/test_value_federate.py -k test_value_federate_runFederateTimeoutTest
============================================================================== test session starts ===============================================================================
platform darwin -- Python 3.7.3, pytest-5.2.1, py-1.8.0, pluggy-0.13.0 -- /Users/USER/miniconda3/bin/python
cachedir: .pytest_cache
rootdir: /Users/USER/GitRepos/HELICS
collected 14 items / 13 deselected / 1 selected

tests/python/test_value_federate.py::test_value_federate_runFederateTimeoutTest PASSED

================================================================== 1 passed, 13 deselected in 60.32s (0:01:00) ===================================================================
```

The fix works!

phlptp:
I think we resolved the error you were getting in 2.3,  might have had to do with finishing converting everything to google tests

phlptp:
do we want to tag a bug fix release with this or just wait two weeks for 2.3?

kdheepak:
I would like a bug fix release.

phlptp:
I updated the release documents with the version number and date and the changelog and also fixed the readme with the correct links for travis-ci.

phlptp:
@nightlark  can you review?

phlptp:
@kdheepak do another review then once the required CI builds have completed you can go ahead and merge

kdheepak:
Looks good to me!

PR #857, opened 2019-10-24
Add Windows builds on Azure

nightlark:
### Summary
If merged this pull request will add several Windows CI builds running on Azure Pipelines with a similar feature set enabled as the current AppVeyor build.

### Proposed changes
- Adds 64-bit VS 2015, 2017, and 2019 builds
- Adds a 32-bit VS 2017 build
- Java interface enabled for the 64-bit builds
- All builds run the build tree find_package test


phlptp:
might want to add the Azure build badge to the Readme.

which branches/PR's does this run for?

nightlark:
It runs on everything.

phlptp:
running in parallel is nice, but the build time is not much faster or seemingly slower than for appveyor.  Is there some caching or other things we can do to make it faster?

nightlark:
I think the VS 2017 and 2019 builds are faster than the Appveyor builds -- 25-26 minutes for the recent builds compared to 28-30 minutes on Appveyor. The VS2015 build is close to the Appveyor build time, but is also installing Boost (disabling boost and avoiding the install would cut the time down for that job). Appveyor does seem to have a lot more variety in build times, yesterday there were some builds builds that took 24 minutes, while a build today took 31 minutes.

PR #859, opened 2019-10-26
Add SWIG interface file update GitHub Action

nightlark:
### Summary
If merged this pull request will add a GitHub Actions workflow to create pull requests to develop for updating the generated SWIG interface files.

Example PR created: https://github.com/nightlark/HELICS/pull/6

### Proposed changes
- Add a swig interface file update PR GitHub Actions workflow
- Change the title of the release build workflow to better describe it than "Release all submodules archive"


phlptp:
So from what I can tell,  this generates the new interface files using the docker container for that. 

Copies them over into a different docker that does the comparison and then generates  a PR?

Is that pretty much the gist of it?

nightlark:
Something like that, but no copying is used -- it mounts a workspace as a shared volume for the containers.

nightlark:
I tried to make the part that pushes a branch and creates a PR with changed files an action that can be reused in other workflows.

phlptp:
Do we want to wait to merge this until after @kdheepak commit to develop or before?  If before I think you can merge

nightlark:
I'll merge now (expecting a PR that updates swig version numbers), then when Dheepak opens a PR with the fix to develop we can make sure it works with the entire cycle of a PR (e.g. runs after it gets merged).

PR #862, opened 2019-10-27
New clang-format

kdheepak:
I took the existing `clang-format` file and uploaded it to https://zed0.co.uk/clang-format-configurator/. I then chose values for every field that I had even a minor opinion on, i.e. this clang-format file is highly opinionated. Even if I didn't have a preference, I chose an option other than "default" so that it'd appear in the file. I prioritized for what I thought is better readability.

You can download the file and upload it to the website above, and tweak settings on the left and see how the code changes on the right.


.clang-format
```diff
@@ -1,56 +1,88 @@
...
+AllowShortIfStatementsOnASingleLine: Never
+AllowShortLambdasOnASingleLine: Empty
+AllowShortLoopsOnASingleLine: 'false'
+AlwaysBreakAfterDefinitionReturnType: None
+AlwaysBreakAfterReturnType: None
+AlwaysBreakBeforeMultilineStrings: 'false'
+AlwaysBreakTemplateDeclarations: 'Yes'
+BinPackArguments: 'false'
+BinPackParameters: 'false'
+BreakBeforeBraces: Attach
```
kdheepak: I think it's easier to read `Attach`:

```cpp
try {
  foo();
} catch () {
}

void foo();

void foo() {
  int aaaa = 12;
  int b = 23;
  int ccc = 23;
}

class foo {};

if (foo()) {
} else {
}

enum X : int { A, B };
```

than `Allman`:

```cpp
try
{
  foo();
}
catch ()
{
}

void foo();

void foo()
{
  int aaaa = 12;
  int b = 23;
  int ccc = 23;
}

class foo
{
};

if (foo())
{
}
else
{
}

enum X : int
{
  A,
  B
};
```

since in `Attach`, the blocks created by `{}` are "attached" to the context as opposed to in `Allman` where it is separated from the context by a newline break. I think `Attach` helps in readability.

.clang-format
```diff
@@ -1,56 +1,88 @@
...
+SortIncludes: 'true'
+SortUsingDeclarations: 'true'
+SpaceAfterCStyleCast: 'false'
+SpaceAfterLogicalNot: 'false'
+SpaceAfterTemplateKeyword: 'false'
+SpaceBeforeAssignmentOperators: 'true'
+SpaceBeforeCpp11BracedList: 'false'
+SpaceBeforeCtorInitializerColon: 'true'
+SpaceBeforeInheritanceColon: 'true'
+SpaceBeforeParens: ControlStatements
```
kdheepak: I think it's easier to read when `SpaceBeforeParens` is `ControlStatements`:

```cpp
void f() {
  if (true) {
    f();
  }
}
```

versus when `SpaceBeforeParens` is `Always`:

```cpp
void f () {
  if (true) {
    f ();
  }
}
```

versus when `SpaceBeforeParens` is `Never`:

```cpp
void f() {
  if(true) {
    f();
  }
}
```

With `SpaceBeforeParens` being `ControlStatements` it is clear what a function call site or function definition is, versus some expression that needs parens to evaluate in an order the developer intends. With `SpaceBeforeParens` set to `ControlStatements`, it is more obvious that there is an expression in the `if` clause has to evaluated and differentiates it from other call sites. `ControlStatements` is the most easiest to read in my opinion.

.clang-format
phlptp: I actually prefer Allman as it gives more whitespace in the file which I think helps with readability,  In some situations I can see it working better like the above example but in general I prefer the extra whitespace,  I have seen code that uses it and it can get really cluttered unless you manually add whitespace lines    I wouldn't want to make this change now though regardless of what we ultimately decide as this one would cause huge changes in the code base, so if we were to make it I don't think we should do that until the 3.0 release which would be a disconnect anyway.

.clang-format
phlptp: This one I agree with,  it would be a big change but it shouldn't change the line count very much, and have thought about changing it.

phlptp:
I don't want to make any major changes to the clang-format until after @nightlark  gets the automated formatting action operational, which likely won't be until after 2.3 comes out.  Then we will have a way of ensuring the completeness of the formatting and correcting any issues quickly and we can experiment with it a little more freely

phlptp:
This is waiting for #893 to be completed before we do anything further

phlptp:
I rebased the change and tweaked a few of the specifications and got rid of the boost/unit_test stuff as that is irrelevant now.

phlptp:
@nightlark @kdheepak @eberleim   Please see #944 for the changes this creates.

Do we want to adjust the line length?  I left a few comments in #944 with some questions

phlptp:
I think this is ready now. Any Objections?  @nightlark @kdheepak

phlptp:
@kdheepak  how do you feel about the current version?  we customized the bracket to only break after a function and before a catch statement.  Otherwise it attaches.

kdheepak:
I think it looks good and I'd be okay merging. The only changes I have a minor preference for are these. There are still some lines that are hard to read when there are multiple arguments or parameters on the same line.

```
BinPackParameters: false
BinPackArguments: false
```

See this as example: https://github.com/GMLC-TDC/HELICS/pull/944/files#r354487271

phlptp:
I don't think those parameters make a difference in that example.  I am going to try it with 100 character lines, see how much difference that makes.

phlptp:
100 as the max width seems to get rid of nearly all the cases where github wrapped the lines in the review panel, which is a nice benefit I suppose.    Any objections?

phlptp:
I think we had those as false already, so I think that example was a matter of line length

phlptp:
Unless there are objections I am going to merge once the tests pass,  then I want to get #945 merged, then merge a formatting PR that makes all the specified formatting changes.

kdheepak:
Looks good to me!

.clang-format
```diff
@@ -1,57 +1,82 @@
...
+#AlignConsecutiveMacros: 'true'
+AlignEscapedNewlines: Right
+AlignOperands: 'false'
+AlignTrailingComments: 'false'
+#AllowAllArgumentsOnNextLine: 'false'
+#AllowAllConstructorInitializersOnNextLine: 'false'
+AllowAllParametersOfDeclarationOnNextLine: 'false'
+AllowShortBlocksOnASingleLine: 'false'
+AllowShortCaseLabelsOnASingleLine: 'false'
+AllowShortFunctionsOnASingleLine: 'true'
```
eberleim: `'true'` does not seem to be one of the allowable values for this option. 
It seems depending on version, clang-format allows specifics of `All`, `None`, `Inline`, `InlineOnly`, and `Empty`.

phlptp: good catch,  I changed it to Inline

.clang-format
```diff
@@ -1,57 +1,82 @@
...
+AlignOperands: 'false'
+AlignTrailingComments: 'false'
+#AllowAllArgumentsOnNextLine: 'false'
+#AllowAllConstructorInitializersOnNextLine: 'false'
+AllowAllParametersOfDeclarationOnNextLine: 'false'
+AllowShortBlocksOnASingleLine: 'false'
+AllowShortCaseLabelsOnASingleLine: 'false'
+AllowShortFunctionsOnASingleLine: Inline
+AllowShortIfStatementsOnASingleLine: 'true'
+#AllowShortLambdasOnASingleLine: 'true'
```
kdheepak: We have commented this so it doesn't matter, but it should be.
    
    ```
    #AllowShortLambdasOnASingleLine: 'Inline'
    ```

kdheepak:
I think this is the default, but we should add it anyway.

```
	SpaceAfterLogicalNot: false
```

.clang-format
phlptp: I will make that change at some point later.  That parameter was not recognized by the automatic format generator so it got commented out.

phlptp:
That is another one that got commented out since the automatic formatter didn't recognize it.  It is in the file just commented out

PR #865, opened 2019-10-28
Master merge

phlptp:
### Summary

If merged this pull request will merge in v2.2.2 back into develop to get the bug fixes and other code changes

### Proposed changes

- timeout bug fix
- updates to readme and changelog
- remove brew install mentions from docs


phlptp:
I was hoping to figure out a way to just get the last few commits into the chain but that wasn't going to work since there was conflicts with the changes in develop.  
So this PR should be merged not squashed otherwise the next release will get complicated.  

This is basically just to get the v2.2.2 changes from master back into the develop branch so we have all the latest bugfixes in develop.

kdheepak:
I can get these commits into develop using cherry picking and fixing conflicts if you'd like.

phlptp:
@kdheepak  that is what I had done, the problem was you had to resolve the conflicts, so they would have been conflicts again when we tried to merge to master for the 2.3 release and we would likely have had to do an even messier merge at that point.

PR #871, opened 2019-10-29
Add inproc core

phlptp:
### Summary

If merged this pull request will split the test core used in HELICS into two parts which can have different goals.  The inproc core will be intended to provide in process memory transfer between federates and can be optimized for that purpose.  The test core will be focused on provided convenient means of testing communication robustness in HELICS.  Previously the test core provided both of these purposes however these roles have started to come into conflict necessitating a split.  

### Proposed changes

- add two new core type definitions
- split the test core into two (mostly similar at this point) core types
- add dedicated tests for the in proc core
- add compiler option for the core
- change most tests to use the inproc core

### DIFF
--- a/src/helics/helics_enums.h
+++ b/src/helics/helics_enums.h
@@ -39,7 +39,10 @@ extern "C"
         helics_core_type_nng = 9, /*!< for using the nanomsg communications */
         helics_core_type_tcp_ss =
           11, /*!< a single socket version of the TCP core for more easily handling firewalls*/
-        helics_core_type_http = 12 /*!< a core type using http for communication*/
+        helics_core_type_http = 12, /*!< a core type using http for communication*/
+        helics_core_type_websocket = 14, /*!< a core using websockets for communication*/
+        helics_core_type_inproc = 18, /*!< an in process core type for handling communications in shared memory
+                                   it is pretty similar to the test core but stripped from the "test" components*/
     } helics_core_type;
 
     /** enumeration of allowable data types for publications and inputs*/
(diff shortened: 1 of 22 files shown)


src/helics/helics_enums.h
```diff
@@ -39,7 +39,10 @@ extern "C"
         helics_core_type_nng = 9, /*!< for using the nanomsg communications */
         helics_core_type_tcp_ss =
           11, /*!< a single socket version of the TCP core for more easily handling firewalls*/
-        helics_core_type_http = 12 /*!< a core type using http for communication*/
+        helics_core_type_http = 12, /*!< a core type using http for communication*/
```
nightlark: Unless this core is going to implement the http protocol for communication, I don't think it is something we should have. If it's a shortcut for a tcpss core using port 80, then it should be named helics_core_type_tcpss_port80 - though the default ports for websockets are port 80 and 443, so if a websocket core is added I'm not sure a port 80 alias to tcpss is needed.

nightlark: Actually, if it's just an alias for the websocket core then the name is okay(ish).

phlptp: I think we had intended that to be an http core using RESTful queries and responses.  The name has been in there for quite a while.  I doubt anyone is using it so we can probably remove it, and I am not sure if that core type has that much purpose.    I was just going to leave it in for the time being as a place holder but I don't have a big issue removing it.

nightlark: Well, that is an accurate name. It is hard to imagine a situation where a websocket connection or an intermediate federate that makes any needed REST queries couldn't be used instead. I think this isn't really an issue related to the PR and it has been a placeholder for a while so it's fine sticking around a while longer.

PR #880, opened 2019-11-06
Filter benchmark

phlptp:
### Summary

If merged this pull request will add a set of Filter benchmarks based on echoMessageBenchmark.  It adds benchmarks with source and destination filters, and some code refactoring to split out EchoMessage class into separate file as it is used by both EchoMessageBenchmarks and filterBenchmarks.  

### Proposed changes

- add FilterBenchmarks
- split EchoMessage class into separate file for use by two Benchmarks


phlptp:
address filter benchmark in #848

PR #881, opened 2019-11-07
PHOLD gbenchmark

nightlark:
### Summary
If merged this pull request will add a PHOLD benchmark to the set of google benchmarks.

### Proposed changes
- Add PHOLD benchmark

### DIFF
--- a/benchmarks/helics/PholdFederate.hpp
+++ b/benchmarks/helics/PholdFederate.hpp
@@ -1,0 +1,120 @@
+/*
+Copyright (c) 2017-2019,
+Battelle Memorial Institute; Lawrence Livermore National Security, LLC; Alliance for Sustainable Energy, LLC.  See
+the top-level NOTICE for additional details. All rights reserved.
+SPDX-License-Identifier: BSD-3-Clause
+*/
+
+#include "helics/application_api/Endpoints.hpp"
+#include "helics/application_api/MessageFederate.hpp"
+#include "helics/core/ActionMessage.hpp"
+#include "helics/helics-config.h"
+#include <random>
+
+using namespace helics;
+/** class implementing the hub for an echo test*/
+class PholdFederate
+{
+  public:
+    helics::Time deltaTime = helics::Time (10, time_units::ns);  // sampling rate
+    helics::Time finalTime = helics::Time (10000, time_units::ns);  // final time
+    int evCount = 0; // number of events handled by this federate
+
+  private:
+    std::unique_ptr<helics::MessageFederate> mFed;
+    helics::Endpoint* ept;
+
+    int index_ = 0;
+    int maxIndex_ = 0;
+    int initEvCount_ = 16; // starting number of events
+    double localProbability_ = .1; // probability of local events
+    double randTimeMean_ = deltaTime * 2;
+
+    std::random_device rd;
+    std::mt19937 rand_gen;
+    std::exponential_distribution<> rand_exp;
+
+    bool initialized{false};
+    bool readyToRun{false};
+
+  public:
+    PholdFederate () = default;
+
+    void run (std::function<void ()> callOnReady = nullptr)
+    {
+        makeReady ();
+        if (callOnReady)
+        {
+            callOnReady ();
+        }
+        mainLoop ();
+    };
+
+    void initialize (const std::string &coreName, int index, int maxIndex)
+    {
+        std::string name = "phold_" + std::to_string (index);
+        index_ = index;
+        maxIndex_ = maxIndex;
+        helics::FederateInfo fi;
+        fi.coreName = coreName;
+        mFed = std::make_unique<helics::MessageFederate> (name, fi);
+        ept = &mFed->registerEndpoint ("ept");
+
+        // en.cppreference.com/w/cpp/numeric/random/exponential_distribution
+        rand_gen.seed(rd());
+        rand_exp = std::exponential_distribution<> (1/randTimeMean_);
+
+        initialized = true;
+    }
+
+    void makeReady ()
+    {
+        if (!initialized)
+        {
+            throw ("must initialize first");
+        }
+        mFed->enterExecutingMode ();
+
+        // create initial events (not included in benchmark time)
+        for (int i = 0; i < initEvCount_; i++)
+        {
+            createNewEvent ();
+        }
+
+        readyToRun = true;
+    }
+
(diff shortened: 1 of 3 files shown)


benchmarks/helics/PholdFederate.hpp
```diff
@@ -0,0 +1,122 @@
...
+        std::string name = "phold_" + std::to_string (index);
+        index_ = index;
+        maxIndex_ = maxIndex;
+        helics::FederateInfo fi;
+        fi.coreName = coreName;
+        mFed = std::make_unique<helics::MessageFederate> (name, fi);
+        ept = &mFed->registerEndpoint ("ept");
+
+        // en.cppreference.com/w/cpp/numeric/random/exponential_distribution
+        rand_gen.seed(rd());
```
phlptp: It occurs to me that we might want the capability of doing repeatable runs, with the same seed here,  whether we use that in the benchmarks is another question, but I think in the federate definition being able to set a consistent seed for deterministic runs might make sense

benchmarks/helics/PholdFederate.hpp
```diff
@@ -0,0 +1,122 @@
...
+        rand_exp = std::exponential_distribution<> (1/randTimeMean_);
```
phlptp: might need "1.0" instead so some compiler doesn't get the idea to do integer division and just to make it clear this is a floating point value

benchmarks/helics/PholdFederate.hpp
```diff
@@ -0,0 +1,122 @@
...
+        }
+
+        readyToRun = true;
+    }
+
+    void createNewEvent ()
+    {
+        // decide if the event is local or remote
+        auto destIndex = index_;
+        if (static_cast<float>(rand()) / static_cast<float>(RAND_MAX) > localProbability_)
```
phlptp: you are mixing random methodologies here, might be more consistent to use something from the <random> library.  I don't think it is wrong here just odd to see the mixed usage.  something like std::uniform_real_distribution<float> distribution(0.0,1.0); might be a little clearer and more consistent

benchmarks/helics/pholdBenchmarks.cpp
```diff
@@ -0,0 +1,197 @@
...
+        state.PauseTiming ();
+        for (auto &thrd : threadlist)
+        {
+            thrd.join ();
+        }
+
+        int totalEvCount = 0;
+        for (auto &f : feds)
+        {
+            totalEvCount += f.evCount;
```
phlptp: what is this counting?

nightlark: The total number of events, since typically the PHOLD benchmark is events/second handled.

benchmarks/helics/PholdFederate.hpp
phlptp: hmmm maybe we want to have a fixed seed for the tests,  just have to make sure that seed is different for each federate.

benchmarks/helics/PholdFederate.hpp
```diff
@@ -26,21 +26,29 @@ class PholdFederate
 
     int index_ = 0;
     int maxIndex_ = 0;
-    int initEvCount_ = 16; // starting number of events
+    unsigned int initEvCount_ = 16; // starting number of events
     double localProbability_ = .1; // probability of local events
```
phlptp: did Peter recommend this value,  might want a comment on why these values were chosen

benchmarks/helics/PholdFederate.hpp
```diff
@@ -26,21 +26,29 @@ class PholdFederate
...
     double randTimeMean_ = deltaTime * 2; // mean for the exponential distribution used when picking event times
 
-    // classes related to the exponential distribution random number generator
-    std::random_device rd;
+    // classes related to the exponential and uniform distribution random number generator
+    bool generateRandomSeed = true; 
+    unsigned int seed = 0;
```
phlptp: Generally a seed of 0 is not the best idea.  My guess is it is fine, but some other numbers might be a better choice here.  pretty sure <random> handles it properly, but would prefer some defined (non-zero number) as the default.

benchmarks/helics/PholdFederate.hpp
```diff
@@ -89,7 +105,7 @@ class PholdFederate
     {
         // decide if the event is local or remote
         auto destIndex = index_;
-        if (static_cast<float>(rand()) / static_cast<float>(RAND_MAX) > localProbability_)
+        if (rand_uniform(rand_gen) > localProbability_)
```
phlptp: :+1  I like this better, looks cleaner and easier to understand

benchmarks/helics/PholdFederate.hpp
phlptp: If the seed is not different for each federate you have the potential to get some oddly coherent behavior between the federates.

benchmarks/helics/PholdFederate.hpp
nightlark: The values are vaguely based on a paper for the Time Warp optimistic synchronization protocol for PDES.

benchmarks/helics/PholdFederate.hpp
nightlark: I found some suggestions that not having the majority of the bits be 0 in the seed might make the generated numbers "better". The alternate choice, 0x600d5eed had quite a few more 0 bits.

benchmarks/helics/PholdFederate.hpp
nightlark: I think the latest changes have helped a lot with making the results vary less between runs.

benchmarks/helics/pholdBenchmarks.cpp
```diff
@@ -0,0 +1,210 @@
...
+    {
+        std::cout << "starting run" << std::endl;
+        state.PauseTiming ();
+
+        int fed_count = static_cast<int> (state.range (0));
+        gmlc::concurrency::Barrier brr (static_cast<size_t>(fed_count));
+        auto wcore = helics::CoreFactory::create (core_type::INPROC, std::string ("--autobroker --federates=") +
+                                                                       std::to_string (fed_count));
+        std::vector<PholdFederate> feds (fed_count);
+        std::mt19937 rand_gen(0x600d5eed);
```
phlptp: this is nice, having a single random seed generate the others

PR #883, opened 2019-11-08
Fix swig doxygen python documentation

kdheepak:
### Summary

If merged this pull request will autogenerate Python documentation in the source code so users can access it from the REPL.

This PR takes the following documentation:

```cpp
    /** create a broker object
    @param type the type of the broker to create
    @param name the name of the broker , may be a nullptr or empty string to have a name automatically assigned
    @param initString an initialization string to send to the core-the format is similar to command line arguments
    typical options include a broker address  --broker="XSSAF" if this is a subbroker or the number of federates or the address
    @forcpponly
    @param[in,out] err an error object that will contain an error code and string if any error occurred during the execution of the function
    @endforcpponly
    @return a helics_broker object, will be NULL if there was an error indicated in the err object
    */
    HELICS_EXPORT helics_broker helicsCreateBroker (const char *type, const char *name, const char *initString, helics_error *err);

```

and generates the following Python documentation:

```python
def helicsCreateBroker(type: "char const *", name: "char const *", initString: "char const *") -> "helics_broker":
    r"""
    create a broker object
       :type type: string
       :param type: the type of the broker to create
       :type name: string
       :param name: the name of the broker , may be a nullptr or empty string to have a name automatically assigned
       :type initString: string
       :param initString: an initialization string to send to the core-the format is similar to command line arguments
           typical options include a broker address  --broker="XSSAF" if this is a subbroker or the number of federates or the address

       :rtype: void
       :return: a helics_broker object, will be NULL if there was an error indicated in the err object
    """
    return _helics.helicsCreateBroker(type, name, initString)
```

Users can now request for help in a REPL and get the doxygen documentation.

Screen Shot 2019-11-08 at 10 29 28 AM


phlptp:
if it works we should update the java swig commands as well so they can add the documentation.

phlptp:
probably need a check to detect the swig version and only add the flag if the version is 4.0 or higher

kdheepak:
What do you think about this `forcpponly` doxygen command?

phlptp:
kind of wish we didn't have to add it to every use,  have you checked that it doesn't impact the actual C documentation generation results?

interfaces/python/CMakeLists.txt
```diff
@@ -26,6 +26,10 @@ else()
 
             set(CMAKE_SWIG_FLAGS "-py3")
 
+            if("${SWIG_VERSION}" VERSION_GREATER "4.0.0")
```
phlptp: we should add this to the Java CMakeLists as well

kdheepak:
I ran `cmake -DHELICS_GENERATE_DOXYGEN_DOC=ON ..` and then `make`. How do I view the documentation once this finishes successfully? I don't see any html files anywhere.

interfaces/python/CMakeLists.txt
```diff
@@ -131,6 +135,10 @@ else()
...
```
phlptp: pretty sure this will work like
```
if (SWIG_VERSION VERSION_GREATER 4.0.0)

**Comment by phlptp** [comment]
RTD generate the doxygen stuff too right?  If so might be sufficient to look in there just to make sure everything looks correct on the doxygen side

**Comment by phlptp** [comment]
I think they get generated in build/doc/html.  I don't often run it like that so I am not sure entirely sure

**Comment by kdheepak** [comment]
I had to run `make helics_doxygen` locally to run doxygen and generate html files, but the documentation is blank. I'm not sure what I'm missing.

**Comment by phlptp** [comment]
At some point I would like to work on getting the documentation generated and installed with the builds.  I don't think we ever worked through all the steps that requires yet though.  So I am not 100% sure the target generation of the documentation works.

**Comment by kdheepak** [comment]
I think they both look good. 

https://external-builds.readthedocs.io/html/helics/883/doxygen/helics_8h.html#a9f90691b63ca3c15f49a8df316975745

[Screen Shot 2019-11-08 at 10 43 21 AM](https://external-builds.readthedocs.io/html/helics/883/c-api-reference/index.html)

File: interfaces/python/CMakeLists.txt (hunk 2 above)
Comment (kdheepak) [reply interfaces/python/CMakeLists.txt]: Resolved in e050c2c3a756d8e4a22f034a919b2b4a3e7cba69.

File: interfaces/python/CMakeLists.txt (hunk 1 above)
Comment (kdheepak) [reply interfaces/python/CMakeLists.txt]: Resolved in 012e2867bdb64e04c957eb4cbd0e82b22ce6e441.

File: interfaces/python/CMakeLists.txt [hunk 3]
```diff
@@ -26,6 +26,10 @@ else()
 
             set(CMAKE_SWIG_FLAGS "-py3")
 
+            if(SWIG_VERSION VERSION_GREATER_EQUAL "4.0.0")
```
phlptp: I don't think this comparison is available in cmake 3.4 so I don't think we can use it until HELICS 3.0   may just need to use SWIG_VERSION VERSION_GREATER 3.9 or something like that so 4.0 works but any of the 3.X series don't, since I don't believe swig ever had a 3.2 or higher

kdheepak:
> kind of wish we didn't have to add it to every use

I don't like it either but it looks better for the user. I tried using other `%feature` directives in swig and avoid changing the c/cpp source files but those didn't do what I wanted.

phlptp:
how does this look in python for helicsInputGetVector or something like that where we have a large typmap translation?

interfaces/python/CMakeLists.txt
kdheepak: Ah that's annoying. Okay, I've changed it back to `VERSION_GREATER` but left it at 4.0.0.

Resolved in 7639096280c260a3e90720497aa94e08f09440ce.

kdheepak:
> how does this look in python for helicsInputGetVector or something like that where we have a large typmap translation?

Not great.

```
def helicsInputGetVector(ipt: "helics_input") -> "int *":
    r"""
    get a vector from a subscription
       :type ipt: void
       :param ipt: the input to get the result for
           data the location to store the data
       :type maxlen: int
       :param maxlen: the maximum size of the vector
           actualSize location to place the actual length of the resulting vector
    """
    return _helics.helicsInputGetVector(ipt)
```

kdheepak:
We'd have to go in and add `forcpponly` and `forpythononly` and `forjavaonly` everywhere to generate really good documentation all from doxygen. 

I think looking at this:

```python
def helicsInputGetVector(ipt: "helics_input") -> "int *":
```

a user might know what is expected of them, and at best they'll they'll guess that the docstring is generated from doxygen comments and at worst they'll think our documentation / docstring is broken.

In either case it seems like the tradeoff is worth it to having the docstring in there, versus having no docstring at all.

phlptp:
I agree that having the doxygen comments of any kind is a big +, regardless if some functions aren't quite right.  Maybe you can try fixing it for one function just to see what it looks like on a common function that might have that issue, then later we can go through an fix the others as the opportunity arises.

nightlark:
Does the option not work in SWIG 4.0.0?

phlptp:
I thought it worked in 4.0

kdheepak:
> Does the option not work in SWIG 4.0.0?

It works in 4.0.0. I think `VERSION_GREATER` than 4.0.0 is fine, since the latest release is 4.0.1 and it is more readable than `VERSION_GREATER` than 3.9.9. I think very few people are going to want to build extensions from source with documentation and even if they do, they can change one line :) This will be fixed when we use `VERSION_GREATER_EQUAL` in the future anyway.

kdheepak:
> Maybe you can try fixing it for one function just to see what it looks like on a common function that might have that issue, then later we can go through an fix the others as the opportunity arises.

I'd rather we be consistently broken right now :) It'll look very much like this PR, where certain docstrings / params are ignored if certain directives are set.

phlptp:
is the Automatic generator using 4.0.1?  if it isn't then we need to fix this if it is it probably doesn't matter that much.

phlptp:
Ok, we can take a closer look at some point down the road

interfaces/python/helicsPython.i
```diff
@@ -1,3 +1,5 @@
+%feature("doxygen:ignore:forcpponly", range="end");
```
phlptp: do we need to add this command in the python2 .i file as well?

kdheepak: Good catch! Resolved in a8794035029c1b426e9161e7cad4d6b57bd0f2d0.

phlptp:
@kdheepak  I think you can go ahead and merge if you think it is ready.  then we can take a look at the auto generated PR

PR #884, opened 2019-11-08
Add benchmarks varying message size and count

nightlark:
### Summary

If merged this pull request will add a set of benchmarks that exercise sending messages of various sizes and quantities. Related to issue #848.

Message count benchmarks go up to 512 because beyond that there wasn't much of interest happening (no crashes), just much longer run times.

Message size benchmarks caused certain cores to crash (ipc and tcp both can't handle 4096, udp core packets are too long at 65K).

Combining the message size and count benchmarks enables interesting results, mostly that show how for many core types increasing the message size (up to a certain point) has no affect compared to sending more messages.

### Proposed changes
- Add message send size and count benchmarks

### DIFF
--- a/benchmarks/helics/messageSendBenchmarks.cpp
+++ b/benchmarks/helics/messageSendBenchmarks.cpp
@@ -1,0 +1,282 @@
+/*
+Copyright (c) 2017-2019,
+Battelle Memorial Institute; Lawrence Livermore National Security, LLC; Alliance for Sustainable Energy, LLC.  See
+the top-level NOTICE for additional details. All rights reserved.
+SPDX-License-Identifier: BSD-3-Clause
+*/
+
+#include "helics/core/BrokerFactory.hpp"
+#include "helics/core/CoreFactory.hpp"
+#include <benchmark/benchmark.h>
+#include <chrono>
+#include <fstream>
+#include <iostream>
+#include <thread>
+#include <random>
+
+#include <gmlc/concurrency/Barrier.hpp>
+
+#include "helics/helics-config.h"
+
+using namespace helics;
+
+// headers used by the MessageExchangeFederate class
+#include "helics/application_api/Endpoints.hpp"
+#include "helics/application_api/MessageFederate.hpp"
+#include "helics/core/ActionMessage.hpp"
+
+using helics::operator"" _t ;
+// static constexpr helics::Time tend = 3600.0_t;  // simulation end time
+/** class implementing a federate that sends messages to another (and vice versa)*/
+class MessageExchangeFederate
+{
+  public:
+    helics::Time deltaTime = helics::Time (10, time_units::ns);  // sampling rate
+    helics::Time finalTime = helics::Time (10, time_units::ns);  // final time
+  private:
+    std::unique_ptr<helics::MessageFederate> mFed;
+    helics::Endpoint ept;
+    bool initialized = false;
+    bool readyToRun = false;
+    
+    int msgCount;
+    std::string msg;
+    std::string dest;
+
+  public:
+    MessageExchangeFederate () = default;
+
+    void run (std::function<void()> callOnReady = {}, std::function<void()> callOnEnd = {})
+    {
+        if (!readyToRun)
+        {
+            makeReady ();
+        }
+        if (callOnReady)
+        {
+            callOnReady ();
+        }
+        mainLoop ();
+        if (callOnEnd)
+        {
+            callOnEnd ();
+        }
+    };
+
+    void initialize (const std::string &coreName, int i, int msg_size, int msg_count)
+    {
+        std::string name = "msgExchange_" + std::to_string (i);
+        helics::FederateInfo fi;
+        fi.coreName = coreName;
+        mFed = std::make_unique<helics::MessageFederate> (name, fi);
+        ept = mFed->registerEndpoint ("ept");
+
+        // set the destination to the other federate in the pair
+        dest = "msgExchange_" + std::to_string ((i+1)%2) + "/ept";
+
+        // create a message string to send
+        msg = std::string (msg_size, '0');
+        msgCount = msg_count;
+
+        initialized = true;
+    }
+
+    void makeReady ()
+    {
+        if (!initialized)
+        {
+            throw ("must initialize first");
+        }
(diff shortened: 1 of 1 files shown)


benchmarks/helics/messageSendBenchmarks.cpp
```diff
@@ -0,0 +1,282 @@
...
+    }
+
+    void mainLoop ()
+    {
+        auto cTime = 0.0_t;
+        while (cTime < finalTime)
+        {
+            while (ept.hasMessage ())
+            {
+                auto m = ept.getMessage ();
```
phlptp: I would imagine this generates a warning of some kind.    Might want to do (void)m or not bother with the m since it is just to get the message out

benchmarks/helics/messageSendBenchmarks.cpp
```diff
@@ -0,0 +1,282 @@
...
+        for (int ii = 0; ii < fed_count; ++ii)
+        {
+            threadlist[ii] = std::thread ([&] (MessageExchangeFederate &f) { f.run ([&brr] () { brr.wait (); }, [&brr] () { brr.wait (); }); },
+                                          std::ref (feds[ii]));
+        }
+
+        // synchronize the federates and run the benchmark with timing
+        brr.wait ();
+        state.ResumeTiming ();
+        brr.wait ();
```
phlptp: This isn't a big deal in this case since there is only two federates, but with this set up it seems possible the timing could be off since the threads will pass the barrier at different times so it seems possible quite a bit of execution could occur before the resumeTiming is actual executed particularly on low core count machines

phlptp: what do we think of the wait and resume in the main thread vs one of the federates running in the timing thread.    I think it has the potential to lead to incorrect timing, since the other threads could have already started once the resume is called.

nightlark: My thought (not sure about accuracy) was that the wait and resume in the main thread followed by a wait and pause timing was that a small delay in resuming timing would be partially offset by pausing timing being a bit late, but that it would give something close to the wallclock time for all federates.

With the other benchmarks running a federate in the timing thread what is actually getting measured is closer to the time it takes for that federate to run, since it might be a bit faster or slower than other federates at exiting and calling pause timing. The times might also be slightly higher due to including the time taken for the main thread federate to synchronize with the other federates at the enterExecutingMode() function.

My impression is that getting accurate high precision benchmark times is tricky. It would be useful to try measuring how much of an effect multiple threads on low core count machines has on resume/pause timing.

phlptp: I guess there are some potential issues either way, and likely no perfect solution.  In the other cases there are a bunch of loops so at most one loop iteration on the non main federates has executed outside the timing loops.  How many loops execute in this test?  If it is only 1 then maybe we need some additional care in this test to make sure we are actually measuring the correct thing.  I suspect for the longer tests the effect is really small, but might be important for shorter tests

nightlark: There's one loop right now we could bump that up by 10x or 100x, but for the tests using larger packet sizes or higher message counts those 2-3 second tests will start taking 20-30s or 200-300s.

phlptp: I think a single loop is appropriate for this test, we could do 2 I suppose but I don't think it resolves the issue.  having the first one just be a dummy might be useful.  Putting both waits the timing loop might also be appropriate in this case.    I think the goal needs to be to make sure we get the entire process in the timing loop.  right now I am not sure that is the case.

phlptp: I suppose one simple way of doing it would be to add another wait in the first callback to the federates.  Then in the main loop the first wait is outside the timing loop.  then the second inside it, along with the final wait.  Then you are assured both federates are in the callback when the timing loop starts and the overhead should be pretty minimal.

nightlark: At most one loop iteration could happen in the non-main federates for starting timing, but there is an unknown amount of time that it could take for the context switch back to the timing federate to happen at the end of the test to stop timing.

I think the bounding properties of starting and stopping timing on one of the federates may be nicer though, and one less thread is involved in context switches. Maybe a barrier that calls a function just before it releases the other threads could get better timing.

There should be a benchmark that does as close to nothing except advancing in time, so that we can get the overhead for starting execution and synchronizing times.

nightlark: I guess a barrier that calls a function would be similar to adding a second wait.

phlptp: That is a good idea for an additional benchmark.  Probably won't put that in 2.3 but we can add it soon after.  We may find we need a few other ones once we start actually accumulating results from different systems on the existing benchmarks.  

Since this test only does one loop and there isn't a federate running in the timing loop we may have to do something a little different

phlptp:
```suggestion
        brr.wait ();
        brr.wait();
```

benchmarks/helics/messageSendBenchmarks.cpp
```diff
@@ -0,0 +1,282 @@
...
+            {
+                feds[ii].initialize (wcore->getIdentifier (), ii, msg_size, msg_count);
+            }
+        }
+
+        // create threads for federates and get most of them running
+        std::vector<std::thread> threadlist (static_cast<size_t> (fed_count));
+        for (int ii = 0; ii < fed_count; ++ii)
+        {
+            threadlist[ii] = std::thread ([&] (MessageExchangeFederate &f) { f.run ([&brr] () { brr.wait (); }, [&brr] () { brr.wait (); }); },
```
phlptp:
```suggestion
            threadlist[ii] = std::thread ([&] (MessageExchangeFederate &f) { f.run ([&brr] () { brr.wait ();brr.wait(); }, [&brr] () { brr.wait (); }); },
```

phlptp: one idea,  using a double wait, with the resume timing in the middle.

PR #888, opened 2019-11-09
Update the Windows installer release build action

nightlark:
### Summary
If merged this pull request will add various updates to the GitHub action for building Windows installers on releases. The main changes are updating it to use renamed CMake HELICS options, a Win32 shared library build is added, and building the installers nightly (relative to UTC-8).

### Proposed changes
- Update HELICS CMake options
- Enable a nightly build of installers using the develop branch
- Add a build for a Win32 shared library
- Switch to using the windows-2019 GH action image to windows-latest in preparation for the consolidation of the two images
- Move some conditionals for steps to conditionals for running the job


.github/workflows/release-build.yml
```diff
@@ -77,7 +83,7 @@ jobs:
         cpack_dir="$(which cmake)"
         cpack_dir="${cpack_dir%/cmake}"
         mkdir build && cd build
-        cmake -G "${BUILD_GEN}" -A "${BUILD_ARCH/x86/Win32}" -DCMAKE_BUILD_TYPE=Debug -DENABLE_PACKAGE_BUILD=ON -DBUILD_SHARED_LIBS=ON -DBUILD_HELICS_EXAMPLES=OFF -DBUILD_TESTING=OFF -DBUILD_APP_EXECUTABLES=OFF -DDISABLE_C_SHARED_LIB=ON ..
+        cmake -G "${BUILD_GEN}" -A "${BUILD_ARCH/x86/Win32}" -DCMAKE_BUILD_TYPE=Debug -DHELICS_ENABLE_PACKAGE_BUILD=ON -DHELICS_BUILD_CXX_SHARED_LIB=ON -DHELICS_BUILD_EXAMPLES=OFF -DBUILD_TESTING=OFF -DHELICS_BUILD_APP_EXECUTABLES=OFF -DHELICS_BUILD_APP_LIBRARY=ON -DHELICS_DISABLE_C_SHARED_LIB=ON ..
```
phlptp: why do we have the C shared library disabled for these builds?

phlptp: are these the ones that are specific to a MSVC compiler?

.github/workflows/release-build.yml
```diff
@@ -220,7 +228,7 @@ jobs:
         echo "Building shared library with ${BUILD_GEN} for ${BUILD_ARCH}"
         choco install -y swig
         mkdir build && cd build
-        cmake -G "${BUILD_GEN}" -A "${BUILD_ARCH/x86/Win32}" -DCMAKE_BUILD_TYPE=Release -DENABLE_PACKAGE_BUILD=ON -DDISABLE_STATIC_LIB_INSTALL=ON -DSTATIC_STANDARD_LIB=ON -DBUILD_APP_EXECUTABLES=OFF -DBUILD_HELICS_EXAMPLES=OFF -DBUILD_TESTING=OFF .. || true
+        cmake -G "${BUILD_GEN}" -A "${BUILD_ARCH/x86/Win32}" -DCMAKE_BUILD_TYPE=Release -DHELICS_ENABLE_PACKAGE_BUILD=ON -DHELICS_DISABLE_STATIC_LIB_INSTALL=ON -DSTATIC_STANDARD_LIB=ON -DHELICS_BUILD_APP_EXECUTABLES=OFF -DHELICS_BUILD_APP_LIBRARY=OFF -DHELICS_BUILD_EXAMPLES=OFF -DBUILD_TESTING=OFF .. || true
```
phlptp: The disable static Lib install option has been removed completely I think, so we should be able to remove it

.github/workflows/release-build.yml
nightlark: Yes, this is the ones specific to a MSVC compiler.

.github/workflows/release-build.yml
nightlark: Do you want the two occurrences of `HELICS_DISABLE_STATIC_LIB_INSTALL` removed?

The only place where it is used and has an effect is https://github.com/GMLC-TDC/HELICS/blob/develop/config/cmake/addlibzmq.cmake

There is also the place where it is set to ON in https://github.com/GMLC-TDC/HELICS/blob/develop/CMakeLists.txt#L121

phlptp: lets remove them here in this PR,  I will remove the other extraneous references in the warning_fixes PR since I am already touching some of those files.

nightlark: Okay, the builds no longer use that option.

phlptp:
What is the test to verify this is working correctly?  have you run it on a fork, or what should we be looking for to verify it is working as intended?

nightlark:
I ran it on my fork and tagged "releases" -- if you want to download and check any of the generated installers they're at https://github.com/nightlark/HELICS/releases

For the nightly build I set it to build every 2 hours and made sure it was building the artifacts, those can be seen under the actions tab of my fork. It should at least let us know if we need to update the workflow for future releases.

PR #895, opened 2019-11-11
Add macOS installers and Linux shared library to release build workflow

nightlark:
### Summary

If merged this pull request will add builds for a macOS and Linux shared library, and a macOS archive that includes some extras like the compiled apps to the releases workflow.

Latest test run at https://github.com/nightlark/HELICS/releases/tag/vMac.7 (MSVC archive builds failed due to swig archive not downloading, probably sourceforge issues -- rerun is at https://github.com/nightlark/HELICS/releases/tag/vMac.8)

### Proposed changes
- Add a macOS and Linux shared libraries (+ headers and swig *.i files) to the release build workflow
- Add a macOS archive that includes pre-compiled apps and some interfaces to the release build workflow

### DIFF
--- a/.github/workflows/release-build.yml
+++ b/.github/workflows/release-build.yml
@@ -285,7 +285,7 @@ jobs:
       run: |
         brew install swig
         mkdir build && cd build
-        cmake -DCMAKE_BUILD_TYPE=Release -DHELICS_ENABLE_PACKAGE_BUILD=ON -DSTATIC_STANDARD_LIB=ON -DHELICS_BUILD_APP_EXECUTABLES=OFF -DHELICS_BUILD_APP_LIBRARY=OFF -DHELICS_BUILD_EXAMPLES=OFF -DBUILD_TESTING=OFF .. || true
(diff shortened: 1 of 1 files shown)


phlptp:
@kdheepak can you take a look at this as well

phlptp:
@nightlark can you add a note in the changelog referencing this(if I didn't to it already), you can do it in this PR or #894

phlptp:
This will close #648

nightlark:
The changelog in the doc update branch has been updated to reference this change.

The binaries were built on macOS 10.15 and Ubuntu 18.04, so for now I'm putting those as the supported versions. The Linux one should work on other distributions and older versions. I'm not sure about on macOS, there was an Xcode setting file that mentioned 10.11 as a minimum. For the upcoming release I don't want to make any guarantees about them working on older versions in case something needs to change like building the Linux one in a docker container with an old version of glibc.

nightlark:
Some notes on building a macOS package using CPack and why it's currently using TGZ:
- Bundle generator creates a .app file which doesn't really work since we don't have a main binary with a gui
- DragNDrop dmg generator had the bin/lib/etc folders at the top level, which looked messy
- PackageMaker is deprecated/no longer in macOS
- productbuild generator creates a .pkg installer which is nice, except it doesn't allow picking an install folder other than a drive (and it asked for a password to run)

The Qt IFW generator might be one to explore for future releases -- since it's cross-platform, it could be a way to get a gui-based installer that's consistent across Windows/macOS/Linux.

PR #899, opened 2019-11-12
Update readme introduction for 2.3 release

bpalmintier:
### Summary

### Proposed changes

-


phlptp:
I think this needs to be against develop, otherwise the merge process is going to be a big pain

phlptp:
@bpalmintier  you might want to just add a small change and recommit since all the PR builds were trying against master which wasn't going to work very well.

README.md
```diff
@@ -43,9 +43,13 @@
...
+Today the core uses for HELICS are in the energy domain, where there is extensive and growing support for a wide-range of electric power system, communications and control-schemes, natural gas, transportation, built environment, and related domain tools. However, it is possible to use HELICS for co-simulation in any domain. Previous and existing HELICS efforts have stretched across a wide range of scales in time—e.g. from transient dynamics (e.g. power system frequency response or electromechanical transient simulation) to long-term planning studies—and space—from individual appliances to nationwide simulations.
```
nightlark: What is `built environment`? Build environments?

I found the end of the line about "in time" and "in space" a bit confusing to read, probably from the arrangement of -'s and nested e.g.'s (forgot about it being an example for "in time" when I reached the "-and space-" part of it).

bpalmintier:
> I think this needs to be against develop, otherwise the merge process is going to be a big pain

Thanks. The branch was from develop, just mixed up the pull request.

README.md
nightlark: Ah, learned a new term and found out what a "built environment" is.

bpalmintier: Built environment is generalization for buildilngs/ Clarified

bpalmintier:
I am also tempted to reduce the scope of the README a bit now that the other documentation is better defined. Thoughts on:

Removing (or switching to documentation link for):
* Install

And merging the content from other pieces into the documentation (likely with some updates on the way). Specifically:
* Getting Started (Read the docs doesn't have youtube among other items. Also should add link to the PESGM tutorial, and maybe remove the old tutorial link if no longer maintained)
* Docker
* Tools with HELICS support (needs updates both to list and to relevant links).

phlptp:
I have noted we do need to update the tools list in some way and some of the other suggestions as well, but I also want to get this release out.    
Doc updates I am not opposed a direct PR on Master so can be done outside the release cycle.  So we should make it a point to do a more thorough review and discussion on the README in the next couple weeks and make an accurate list of the tools

phlptp:
@nightlark @kdheepak   can you make a final review of this.  Then I can merge and make the PR into master.

Issue #877, opened 2019-11-01
Build option to disable ASIO


phlptp:
**Is your feature request related to a problem? Please describe.**
There are a few situations where ASIO doesn't work well or at all (Cygwin, WASM) or other times when it is not needed.  It would useful if we had a CMAKE flag and library support for disconnecting the use of ASIO from the library.    This will by definition mean severe restrictions on which core types can be used, but most functionality should be possible.  

**Describe the solution you'd like**
A build option for HELICS_DISABLE_ASIO

PR #907, opened 2019-11-14
Disable asio option

phlptp:
### Summary

If merged this pull request will add a cmake option to disable linking with ASIO.  This can be useful in some environments where ASIO isn't supported.  Such as cygwin 64 bits.  It turns off the TCP and UDP cores as well as any timeout capabilities, and real time capabilities of HELICS.  Some of options related to real time operation have been disabled and will result in an error if used.  Others are left in place since there wasn't a good way to disable them.  Some potential options in the future may be available in later version of CLI11.  But the use cases for this are expected to be rare.  

### Proposed changes

- Add cmake (advanced option) to disable ASIO
- disable code and tests that make use of ASIO
- disable some options and configurations based on ASIO

will resolve Issue #877

### DIFF
--- a/src/helics/core/BrokerBase.cpp
+++ b/src/helics/core/BrokerBase.cpp
@@ -489,7 +489,7 @@ void BrokerBase::queueProcessingLoop ()
             }
             messagesSinceLastTick = 0;
             // reschedule the timer
-#ifndef HELICS_DISABLE_ASIO
+			#ifndef HELICS_DISABLE_ASIO
             ticktimer.expires_at (std::chrono::steady_clock::now () + tickTimer.to_ns ());
             active = std::make_pair (true, true);
             ticktimer.async_wait (timerCallback);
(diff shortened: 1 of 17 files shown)


src/helics/core/BrokerBase.cpp
```diff
@@ -475,9 +489,11 @@ void BrokerBase::queueProcessingLoop ()
             }
             messagesSinceLastTick = 0;
             // reschedule the timer
+#ifndef HELICS_DISABLE_ASIO
```
nightlark:
```suggestion
			#ifndef HELICS_DISABLE_ASIO
```

Matching the indentation for the #endif

phlptp: There is a clang-format option for formatting indentation of pragmas in the newer version of clang-format.  I am hoping when we get that automated the pragmas such as this will be consistent

PR #913, opened 2019-11-20
Update the release builds workflow

nightlark:
### Summary
If merged this pull request will move the release build scripts from the GitHub Actions workflow YAML file to script files, so that the version run in the nightly scheduled builds can be updated. Compatibility with release binaries across a wider variety of Windows and Linux versions is also improved.

Static linking with the standard libraries is now properly enabled for the Linux shared library, avoiding a dependency on the libstdc++ shared library, and static linking is now working for the Windows installers and shared library builds.

To (hopefully) enable the Linux shared library to work as far back as RHEL 6, CentOS 6, Ubuntu 12.04, and Debian 7, a docker container using CentOS 6 with gcc 8.3.1 is used for building. A compiler bug with 7.3.1 was uncovered while setting this up, see https://gcc.gnu.org/bugzilla/show_bug.cgi?id=84785 and https://github.com/mpark/variant/issues/43 for details. tl;dr is use any gcc version other than 7.3.1, since 7.3.0 and 7.4.0 are not affected.

### Proposed changes
- Move the scripts for release builds from the workflow YAML file to shell scripts, to enable updates in the develop branch for nightly builds
- Change how ZeroMQ is installed when used as a subproject on Linux and macOS to make sure shared libraries and required symlinks are created in the right places (requires CMake 3.13+ to install HELICS built from source on Linux/macOS when using ZMQ as a subproject; building from source with ZeroMQ already installed on the system will still work with older CMake versions)
- Update documentation for the correct usage of the `STATIC_STANDARD_LIB` CMake option
- Properly static link with the standard library for Windows installer and shared library builds (but not for msvc archives, because use of those implies the needed standard libraries are installed)
- Static link with the standard library for the Linux shared library build
- Build the Linux shared library for releases on a CentOS 6 image with glibc 2.12 to improve the chance of it working on older versions


nightlark:
This should fix the currently failing nightly release build test whenever the workflow yaml file makes it into the master branch.

config/cmake/addlibzmq.cmake
```diff
@@ -159,11 +159,21 @@ endif()
...
+               LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
+               FRAMEWORK DESTINATION "Library/Frameworks"
+		   )
+		elseif(WIN32)
+		   install(
+               FILES $<TARGET_FILE:${zmq_target_output}>
+               DESTINATION ${CMAKE_INSTALL_BINDIR}
+               COMPONENT libs
+		   )
+		endif()
```
phlptp: Is there anything we can put in an additional else() clause that would be useful and correct?

config/cmake/addlibzmq.cmake
```diff
@@ -159,11 +159,21 @@ endif()
 if(HELICS_BUILD_CXX_SHARED_LIB OR NOT HELICS_DISABLE_C_SHARED_LIB)
 
 	if(NOT HELICS_USE_ZMQ_STATIC_LIBRARY)
-		install(
-			FILES $<TARGET_FILE:${zmq_target_output}>
-			DESTINATION ${CMAKE_INSTALL_BINDIR}
-			COMPONENT libs
-		)
+		set_target_properties(${zmq_target_output} PROPERTIES PUBLIC_HEADER "")
+		if(NOT ${CMAKE_VERSION} VERSION_LESS "3.13")
```
phlptp: Is this going to work properly with a package builder on windows/linux?    Just wondering if the WIN32 clause maybe should be first before the version 3.13 check of CMAKE.

phlptp:
So in general what this is doing is shifting most of the builds to run in scripts that we can update on a branch basis and fixing the linux build?  

What exactly was the issue that caused the Linux build to fail?

nightlark:
The Linux builds failed because this PR https://github.com/GMLC-TDC/HELICS/pull/896 updated some places where the ZMQ subproject option name was not updated, one of which is the `FATAL_ERROR` line  that requires CMake 3.13 -- the build images for Ubuntu on GitHub Actions have CMake 3.12.

This does 4 main things:
- shifts most of the build to run in scripts so that it uses the branch version of the scripts instead of whatever is in master
- moves the Linux shared library build to run in a CentOS 6 container, which has a version of CMake 3.13 (fixing the build)
- fixes documentation on `STATIC_STANDARD_LIB` and updates builds using the option to use it correctly (reducing dependencies for the resulting binaries on shared libraries)
- fixes the ZeroMQ library install location, and fixes symlink creation on Linux/macOS

config/cmake/addlibzmq.cmake
nightlark: CMake does the right thing for Windows and places .dll files (zmq, libhelicsShared, etc) in the bin folder, and .lib files in the lib folder.

config/cmake/addlibzmq.cmake
nightlark: The only thing would be an install with the TARGET_FILE for zmq, and then trying to create the correct symlinks manually. I think that would be more trouble to get working and maintain than it is worth.

Maybe an error message saying "upgrade to CMake 3.13+ or enable HELICS_USE_ZMQ_STATIC_LIBRARY=ON".

config/cmake/addlibzmq.cmake
nightlark: The way some of the HELICS libraries are getting installed right now with an install TARGET_FILE command might be redundant, but that's another issue.

phlptp: Can we verify that this generates the correct package with ZMQ on windows with CMAKE>=3.13.

phlptp: Should we just disable packaging completely on windows if cmake<3.13 and ZMQ non-static subproject is used?

nightlark: Yes, the default CMake version included in the Windows GitHub Actions images has been >3.13 for a while now. The shared library archives from the most recent build are at https://github.com/nightlark/HELICS/releases/tag/vLinuxFix.17

I noticed the helics app binaries weren't getting included, since the executables getting built depend on the shared library being enabled. It looks like there is no way to create an installer with the helics apps that doesn't also include the helics app shared library. I'll run another test build for the installers to confirm that's working again in a bit.

nightlark: I think users building with ZMQ as a static subproject would make their lives easier, so they never need to worry about forgetting to copy the ZMQ dll to the right place. It would also simplify this part of the addlibzmq.cmake file.

phlptp: Lets review this over the next couple weeks make sure we have the right defaults and error conditions, and see if we can resolve the lingering questions of what we can and cannot do with ZMQ licensing to make sure any packaged distributions are BSD compliant.

nightlark: Alright. For now I added a warning about an unsupported configuration for installing from source if a Linux or macOS build with CMake <3.13 is using ZeroMQ as a subproject.

nightlark: The final test build is at https://github.com/nightlark/HELICS/releases/tag/vLinuxFix.18
From going through the different tar and zip files, it looks like the expected files are in each one.

phlptp:
Is this ready to merge?

nightlark:
I think so, I’m not seeing any other issues with the release files for Windows and Linux at the moment.

phlptp:
Ok, I will merge this and we can check a few things on master.   Today I will make a release update commit to develop and check with @kdheepak  if anything else needs to be fixed for a 2.3.1 release and we can tag the release by this afternoon, or tomorrow morning and make sure everything is working correctly with the release builds

Issue #918, opened 2019-11-23
helics::HelicsException: TEST core is not available


kdheepak:
I'm getting a segfault when running Julia with HELICS v2.3.1.

```julia
$ julia --startup-file=no --project 
               _
   _       _ _(_)_     |  Documentation: https://docs.julialang.org
  (_)     | (_) (_)    |
   _ _   _| |_  __ _   |  Type "?" for help, "]?" for Pkg help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.3.0-rc4.1 (2019-10-15)
 _/ |\__'_|_|_|\__'_|  |  Official https://julialang.org/ release
|__/                   |

julia> using HELICS

julia> const h = HELICS
HELICS

julia> h.helicsGetVersion()
"2.3.1 (2019-11-22)"

julia> h.helicsCreateCore("test", "core1", "--autobroker")
libc++abi.dylib: terminating with uncaught exception of type helics::HelicsException: TEST core is not available

signal (6): Abort trap: 6
in expression starting at REPL[4]:1
__pthread_kill at /usr/lib/system/libsystem_kernel.dylib (unknown line)
Allocations: 1757642 (Pool: 1757254; Big: 388); GC: 2
[1]    37599 abort      /Users/USER/Applications/Julia-1.3.app/Contents/Resources/julia/bin/julia
```

kdheepak:
<img width="1551" alt="Screen Shot 2019-11-23 at 5 54 30 AM" src="https://user-images.githubusercontent.com/1813121/69479018-035f2f00-0db6-11ea-95f1-6d65395c8258.png">

[This change](https://github.com/GMLC-TDC/HELICS/commit/02c450f37b3dbe140ecd7611da67cc78eebc976b#diff-af3b638bc2a3e6c650974192a53c7291R391) is what is causing the issue. I have tests in HELICS.jl that use the `test` core but I'm building the shared library with `-DHELICS_BUILD_TESTS=OFF`. I'll update the Julia tests to not use the `test` core since we don't expect users to be using that.

kdheepak:
I'm reopening this since this causes a segfault. This should have thrown an error in Julia instead.

phlptp:
[...]

PR #919, opened 2019-11-23
Helics create core leaked exception

phlptp:
### Summary

If merged this pull request will fix a potential leaked exception if a known core type was unavailable and used in the helicsCreateCore function.
Fixes issue #918

### Proposed changes

- add try/catch block around helicsCoreCreate section which could throw the exception.  
- add a test method to execute it.
- add "nng"  string to getCoreType so we can generate a known unavailable core type for use in the test.

### DIFF
--- a/src/helics/core/coreTypeOperations.cpp
+++ b/src/helics/core/coreTypeOperations.cpp
@@ -95,6 +95,7 @@ static const std::unordered_map<std::string, core_type> coreTypes{{"default", co
                                                                   {"websocket", core_type::WEBSOCKET},
                                                                   {"web", core_type::WEBSOCKET},
                                                                   {"inproc", core_type::INPROC},
+                                                                  {"nng", core_type::NNG},
                                                                   {"http", core_type::HTTP},
                                                                   {"HTTP", core_type::HTTP},
                                                                   {"web", core_type::HTTP},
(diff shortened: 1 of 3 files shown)


src/helics/core/coreTypeOperations.cpp
```diff
@@ -95,6 +95,7 @@ static const std::unordered_map<std::string, core_type> coreTypes{{"default", co
                                                                   {"websocket", core_type::WEBSOCKET},
                                                                   {"web", core_type::WEBSOCKET},
                                                                   {"inproc", core_type::INPROC},
+                                                                  {"nng", core_type::NNG},
```
kdheepak: What does `NNG` stand for?

phlptp: I believe it stands for [Nanomsg Next Generation](https://github.com/nanomsg/nng), which is an alternative to ZMQ, which at some points in the past we had debated creating a core type for, so the type names existed for it, but we haven't had a requirement that drove it in a while.   When we get to some large scales it may be worth a second look since it is MIT licensed, but we will see.  For the time being it was just convenient for this purpose.    For HELICS 3.0 We probably should just make a dummy one has the type and string conversion but it guaranteed to not exist.

PR #922, opened 2019-11-26
Move build environment Dockerfiles to GMLC-TDC/helics-buildenv

nightlark:
### Summary

If merged this pull request will finish moving the build environment related Dockerfiles (ones that don't use any part of HELICS in the resulting container) to the [GMLC-TDC/helics-buildenv](https://github.com/GMLC-TDC/helics-buildenv) repository by removing the migrated files.

### Proposed changes
- Move the HELICSBuilder (helics/buildenv:builder docker image tag), sanitizer, octave, and interface-gen Dockerfiles to GMLC-TDC/helics-buildenv repository
- Move the sanitizer_tests.sh script from config/Docker to scripts/ in this repository


config/Docker/sanitizer_tests.sh
```diff
@@ -1,19 +0,0 @@
...
-for test in "${tests[@]}"; do
-    echo "${test}"
-    eval "${test}"
-    RESULT=$?
-    echo "***Latest test result: "${RESULT}
-    SUMRESULT=$(( SUMRESULT + RESULT ))
-done
-# Return 0 or a positive integer for failure
-exit ${SUMRESULT}
-
```
phlptp: Is there a similar file to this one in the scripts folder.  I think at some point I was planning on shifting the circle ci builds to use a script like this, not sure if there was a duplicate in the scripts folder or not.  If not I think I want a copy of this in that folder so it can be used on the HELICS build tests.

phlptp:
I assume the corresponding changes were made to the buildenv repo?

config/Docker/sanitizer_tests.sh
nightlark: I thought something weird was going on with git because this showed up as deleted instead of moved. There is an identical copy of this file in the scripts/ folder: https://github.com/GMLC-TDC/HELICS/blob/move-dockerfiles/scripts/sanitizer_tests.sh

nightlark:
Yes, all the Dockerfiles are in the buildenv repo. It has also been setup with a GitHub actions workflow that will build and push just the Dockerfiles that have been changed by a commit.

PR #925, opened 2019-11-27
enable benchmarks to be built by older cmake versions

phlptp:
### Summary

If merged this pull request will fix an issue gitUtils that impacted building google benchmark with older cmake versions

### Proposed changes

- fix issue with gitUtils
- add building benchmarks on travis gcc 6 build
- add some more flags to turn off building internal tests on google benchmark on linux systems
- fix bug trying to build java interface with cmake 3.7.X where X is 1 or 2 
- Add RUN_ALL_BENCHMARKS and RUN_KEY_BENCHMARKS targets to generate benchmark files automatically
- add circle ci benchmark run


phlptp:
@nightlark Do we have any builds on travis on LINUX with cmake greater than 3.5 and less than 3.11?

nightlark:
It looks like google benchmarks requires at least CMake 3.5.1

nightlark:
No, it doesn't look like there is. The gcc 6 daily build using 3.11 could be downgraded to 3.10.

phlptp:
so it appears.  I am going to add a check on that more directly to not enable that option if the version of cmake is not high enough.

phlptp:
I think that version was using 3.11 to test the other path so I don't think we should change that

phlptp:
can we bump the gcc 6 build to use 3.8 or something like that

nightlark:
That should be okay, most of the other builds are using 3.4.3.

phlptp:
How do I do that?

phlptp:
- USE_CMAKE_VERSION=3.7.2

?

nightlark:
In the env section for the .travis.yml file job, add USE_CMAKE_VERSION=3.8.2

nightlark:
Some of the Travis caches might also need clearing (master/develop).

phlptp:
there are a lot of old caches around on travis.  before I merge this I am going to delete them all.

phlptp:
I am running through all the benchmarks, making sure they run in a reasonable amount of time and making some targets that can be run to generate the results files in a consistent fashion.   So I am going to add that to this PR as well, and @nightlark I will ask you to review again once completed.

phlptp:
I added a few targets for running benchmarks and made it so there is a regular run (3x/week) on develop on circleci with the key benchmarks.   Someday maybe we can make that into a test for performance regressions.    @nightlark  can you take a look through the recent changes.  Then I can merge it.

nightlark:
I can't see the settings page for the plan on CircleCI (requires admin), but if it's the free plan (not sure if open source/organizations have a different plan than individuals) there's a bit in the documentation that mentions contacting support about resource classes: https://circleci.com/docs/2.0/configuration-reference/#resource_class

phlptp:
I am trying to run the RUN_ALL_BENCHMARKS target on a linux system.  My first attempts caused an issue in one of the tests so I made a few tweaks.  If those run to completion I will merge this.

PR #930, opened 2019-11-28
Add a clang-format actions workflow

nightlark:
### Summary
If merged this pull request will open automated clang-format PRs for commits to develop and master that don't match the format given in the `.clang-format`. Updates to code in the ThirdParty folder or automated swig generated source files in the interfaces folder will not trigger a formatting run.

~A manual run can be triggered by commenting on a PR with `/format`. A PR with the formatting changes will be opened that can be merged into the PR (or the base branch could be changed to develop for merging later).~

The `format-trigger-testing` branch shows the results of some tests of the path/file triggers from some commits that did not result in any open PRs (or workflow runs) and two that did (#927 and #928).

Update on comment triggers: the Git SHA in the comment trigger is for the latest commit in the default branch, which could be entirely unrelated to the source and target branches involved in a PR. It looks like there might not be the required information in the event context to add support in an easy/straightforward way.

### Proposed changes
- Add a clang-format GitHub Actions workflow that automatically runs clang-format and opens a PR with formatting changes when HELICS (not swig generated or third party) code or formatting related files are updated.
- ~Add a `/format` comment as a trigger for the automatic formatting workflow to format a branch~

### DIFF
--- a/.github/actions/create-file-update-pr/action.yml
+++ b/.github/actions/create-file-update-pr/action.yml
@@ -1,5 +1,12 @@
 name: 'Create File Update PR'
 description: 'Pushes a new branch with changed files and opens a PR for updating the changed files in the base branch'
+inputs:
+  no-hash: # turn off inclusion of the hash in the created branch name
+    description: 'Do not include the changed file hash in the PR branch name'
+    default: false
+  replace-branch: # if a PR branch already exists, replaces it with updated contents
+    description: 'Replace the contents of an existing PR branch of the same name'
+    default: false
 runs:
   using: 'docker'
   image: 'Dockerfile'
--- a/.github/actions/create-file-update-pr/entrypoint.sh
+++ b/.github/actions/create-file-update-pr/entrypoint.sh
@@ -11,7 +11,12 @@ files_changed=$(git diff --staged --name-only)
 if [[ "$files_changed" != "" ]];
 then
   hash=$(sha256sum "${files_changed}" | sha256sum | cut -c 1-12 -)
-  current_branch="${GITHUB_REF#refs/heads/}"
+  if [[ "${GITHUB_REF}" == "refs/pull/"* ]];
+  then
+    current_branch="${GITHUB_HEAD_REF}"
+  else
+    current_branch="${GITHUB_REF#refs/heads/}"
+  fi
 
   # Set the git origin url for committing using a GITHUB_TOKEN
   git remote set-url origin "https://x-access-token:${GITHUB_TOKEN}@github.com/${GITHUB_REPOSITORY}"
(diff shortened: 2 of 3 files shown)


nightlark:
Somewhat annoying, but it seems like responding to `issue_comment` events will require the workflow file to make its way into the master branch before the feature becomes active.

phlptp:
Is this going to work like the interface files PR generator in that it generates a branch based on the SHA?  I am guessing we will need to try this out for a while and may want to make some tweaks as we try it out, to see if it plays nicely with our workflow.

nightlark:
Yea, it works like the swig PR generator. I was hoping to get something working that could be run just before a branch gets merged to make sure the formatting is good (a server running probot might be the best way to deal with handling commands in comments)

I’m guessing we might want to switch it to running on a schedule or when a PR to master is opened to avoid a new formatting update after every code change we make.

phlptp:
I think I want to see how things shape up over December to better understand what we want going forward.  

Would there be a way to have it make a commit to a branch if there was a specific keyword in the PR of that branch?  or is that what was giving you trouble?

phlptp:
I think the way we will deal with this in the short term is get #925 and #930 merged, which I presume will generate a fairly large PR.    Then we take a rigorous look at .clang-format in #862 and make sure we have that the way we want it for now before merging what I suspect will be a PR with a lot of files.

nightlark:
Responding to a keyword in the PR for a branch has challenges similar to responding to a comment on a PR - the commit sha info provided in the event is all for the default branch. It seems to be because PRs are treated like a type of issue by GitHub.

It would be easy to add something that lets it run if the branch name contains a particular keyword.

phlptp:
How practical would a HELICS-bot comment on whether the PR would generate a format change, like for HELICS-examples be?    I don't want it to be a check since it is probably ok either way, but an indicator would be useful if the submitter is still working on it and can run the formatting locally.

nightlark:
That could work, though if the PR is a work in progress it could result in quite a few comments.

phlptp:
Can it be disabled if there is [WIP] in the PR name or it is marked as draft?  It wouldn't be the worst thing for use to get into better habits in that regard.       We may want to consider doing the same for the Examples and NS3 tests if it isn't hard.

nightlark:
For ns-3 and the examples, a draft PR shouldn’t trigger builds because Travis isn’t running the jobs that send the api request.

I think a pull request trigger might not be affected by drafts (need to check), so could easily set it to open a formatting PR if the target is develop/master.

Drafts would be preferable over WIP because it results in less builds getting queued.

nightlark:
The formatting PR for this branch is https://github.com/GMLC-TDC/HELICS/pull/942 -- if any additional commits are made, the formatting PR will be updated with a force push. https://github.com/GMLC-TDC/HELICS/pull/941 shows what it would look like in the force push scenario.

Issue #893, opened 2019-11-11
Github Action for clang-format


phlptp:
It would be useful to have an automated github action that could reformat the code in the HELICS repository and generate a PR for it to ensure consistent formatting in the code repository, and allow changes to the formatting on a coherent basis.

phlptp:
#930 resolves this

PR #935, opened 2019-12-02
Add pre-commit hook to run clang-format and CMake flag to enable/disable it

eberleim:
### Summary

If merged this pull request will add a clang-format verification git hook, and set a flag in CMake that toggles the hook `ON` or `OFF` (default `ON`), so HELICS can still be built in environments where clang-format is not available. 
Additionally, when enabled, the CMake script will verify the availability of clang-format prior to setting the `core.hooksPath` in git, and will unset `core.hooksPath` if `ENABLE_GIT_HOOKS` is set to `OFF`. 

The detection is non-changing, and simply notifies the user of any C/C++ files which do not pass the current formatting requirements. If the user is unable to run clang-format on the specified files, it can be manually bypassed by giving the `--no-verify` flag to `git commit`. 

The operation of this component requires Git and clang-format, and will appropriately check for the presence of Git prior to invoking them. 

Relates to #893 and other clang-format changes, as this should help reduce repository noise when used.


eberleim:
All CI systems will probably fail first attempt, as the default value of the new flag `ON` will cause `FATAL_ERROR` in CMake on those platforms.

config/cmake/addGitHooks.cmake
```diff
@@ -0,0 +1,31 @@
...
+    )
+endfunction()
+
+function(set_git_hooks_disabled)
+    execute_process(
+            COMMAND ${GIT_EXECUTABLE} config --local --unset core.hooksPath
+    )
+endfunction()
+
+option(ENABLE_GIT_HOOKS "Activate git hooks to run clang-format on committed files." ON)
```
nightlark: I think this should use cmake_dependent_option to make the option default to on when clang-format is available, or off if it isn't.

```suggestion
cmake_dependent_option(ENABLE_GIT_HOOKS "Activate git hooks to run clang-format on committed files." ON "CLANG_FORMAT" OFF)
```

config/cmake/addGitHooks.cmake
```diff
@@ -0,0 +1,31 @@
+cmake_minimum_required(VERSION 2.8.7)
+
+find_package(Git)
+if(NOT GIT_FOUND)
```
nightlark: This shouldn't be a fatal error, otherwise it may be impossible to build on a system that's using the tar archive for a release on a system that doesn't have git installed. I think wrapping the rest of this file inside `if(GIT_FOUND)` instead would avoid problems. It should probably also include a version check to make sure the version of git used supports setting a hook path. So maybe use a condition like `if(GIT_FOUND AND (GIT_VERSION_STRING VERSION_GREATER "2.9"))`.

config/cmake/addGitHooks.cmake
```diff
@@ -0,0 +1,31 @@
...
+    message(FATAL_ERROR "git not found!")
+endif()
+
+include(clang-cxx-dev-tools)
```
nightlark: This may already be included earlier @ https://github.com/GMLC-TDC/HELICS/blob/master/CMakeLists.txt#L694

```suggestion
```

config/cmake/addGitHooks.cmake
```diff
@@ -0,0 +1,31 @@
...
+
+function(set_git_hooks_disabled)
+    execute_process(
+            COMMAND ${GIT_EXECUTABLE} config --local --unset core.hooksPath
+    )
+endfunction()
+
+option(ENABLE_GIT_HOOKS "Activate git hooks to run clang-format on committed files." ON)
+if (ENABLE_GIT_HOOKS)
+    if (CLANG_FORMAT)
```
nightlark: Should probably also make sure the HELICS_ENABLE_CLANG_TOOLS option is enabled, since that's the option that includes the clang tools earlier in the root CMakeLists.txt file.
```suggestion
    if (HELICS_ENABLE_CLANG_TOOLS AND CLANG_FORMAT)
```

nightlark:
Thanks for the clang-format githooks PR!

nightlark:
Other than cloning the submodules and getting a commit-ish for a version string, what else in the CMake scripts require git? If the source archive that includes submodules is used it should be possible to build HELICS without a copy of git -- any existing features preventing that are things that should be changed.

config/cmake/addGitHooks.cmake
phlptp: We need to be able to build HELICS in some configuration with no git.  So we can't have a requirement on Git that can't be bypassed.  If there are some requirements on GIT that can't be bypassed already in place we need to fix them.    No issue using git if it is available but we do need to be able to build without it.

config/cmake/addGitHooks.cmake
phlptp: We probably want a minimum version of clang-format.  An old one might not work properly.

phlptp: I don't think there are any issues linking to clang-format if a different compiler is used.

eberleim:
@nightlark That may be correct, I saw the FATAL_ERROR in GitUtils.cmake, but that is only invoked from submodule acquisition, or the test suite, so it may be that the packaged version would not require it. 
I am making some additional changes to this PR and will submit them shortly. 

Thanks for your feedback!

phlptp:
I don't know of any other places we are using GIT,  grabbing a commit description, and updating the submodules.  Both of which I think can be bypassed easily or automatically.

config/cmake/addGitHooks.cmake
eberleim: `HELICS_ENABLE_CLANG_TOOLS` is currently always `OFF`, and only works if the compiler is `Clang`, looking at it a bit closer now, depending on this flag is problematic. it may make more sense to simply verify `clang-format` is available separately for this code.

phlptp: The clang tools targets probably need some updating since I haven't used it in a while.  So I don't think a dependency for this work is necessary or even desirable if we are adding a commit hook and auto formatting PR generator in #930

config/cmake/addGitHooks.cmake
```diff
@@ -0,0 +1,36 @@
...
+
+find_package(Git)
+
+if (GIT_FOUND AND (GIT_VERSION_STRING VERSION_GREATER_EQUAL "2.9"))
+    if (HELICS_ENABLE_GIT_HOOKS)
+        set_git_hooks_enabled()
+    else ()
+        set_git_hooks_disabled()
+    endif ()
+elseif (GIT_FOUND AND (GIT_VERSION_STRING VERSION_LESS "2.9"))
```
phlptp: This warning should only be triggered if HELICS_ENABLE_GIT_HOOKS is active, otherwise there should be no warning since there isn't any issue.

phlptp: Likewise with the next message,  and I think that MESSAGE needs WARNING level indicator on it as well.

config/cmake/addGitHooks.cmake
```diff
@@ -0,0 +1,36 @@
...
+
+if (NOT DEFINED CLANG_FORMAT)
+    find_program(CLANG_FORMAT "clang-format")
+endif ()
+
+cmake_dependent_advanced_option(HELICS_ENABLE_GIT_HOOKS "Activate git hooks to run clang-format on committed files." ON "CLANG_FORMAT" OFF)
+
+find_package(Git)
+
+if (GIT_FOUND AND (GIT_VERSION_STRING VERSION_GREATER_EQUAL "2.9"))
```
nightlark: The older CMake versions we support don't have VERSION_GREATER_EQUAL. I think the simplest alternative is:
```suggestion
if (GIT_FOUND AND (NOT GIT_VERSION_STRING VERSION_LESS "2.9"))
```

eberleim: Good catch, thanks, I've made this update.

config/cmake/addGitHooks.cmake
```diff
@@ -0,0 +1,67 @@
...
+
+if (HELICS_ENABLE_GIT_HOOKS AND NOT CLANG_FORMAT_VERSION_OK)
+    message(WARNING
+            "HELICS git hooks require clang-format version ${CLANG_FORMAT_MIN_VERSION} or higher."
+            "\nInstall or update clang-format to use git hooks."
+            "\nTo select a different clang-format exectuable, update the CLANG_FORMAT property in CMake.")
+    SET(HELICS_ENABLE_GIT_HOOKS OFF CACHE BOOL "Activate git hooks to run clang-format on committed files." FORCE)
+endif ()
+
+if (GIT_FOUND AND (NOT GIT_VERSION_STRING VERSION_LESS "2.9"))
```
phlptp: I don't think cmake needs the inner parenthesis

phlptp:
based on the GCC build the GIT HOOKS integration needs to be turned off by default when HELICS is used as a subproject,  it also probably should be turned off on most if not all of the CI builds.

.githooks/pre-commit
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+files=()
+for file in `git diff --cached --name-only --diff-filter=ACMRT | grep -E "\.(cpp|hpp|c|h)$"`; do
+  if ! cmp -s <(git show :${file}) <(git show :${file}|clang-format); then
+    files+=("${file}")
```
nightlark: Somewhere here or in the grep on line 4 there should be a filter to exclude source files in two subfolders:
- `interfaces` to avoid formatting/interface file generation cycles
- `ThirdParty` to make sure we aren't messing up the formatting of some vendored code that doesn't follow the same style

kdheepak:
This might be better off implemented as part of this framework: https://github.com/pre-commit

There's a bunch more pre-commit hooks that are extremely useful that I have installed globally in my environment.

phlptp:
@eberleim if we can get this finished by the end of the week we will be able to get it into a patch release for 2.3.2

eberleim:
@phlptp I have most of the remaining review suggestions implemented, I'm curious if we currently have a CMake flag that can be easily used to identify when used as a subproject or CI build. I saw some logic that uses `if (CMAKE_PROJECT_NAME STREQUAL PROJECT_NAME)` but wanted to verify if this is our current "best practice" or if there's a better way which I've missed.

eberleim:
@kdheepak The pre-commit framework looks like something worth investigating in the future, but is a more comprehensive process than I have time to implement this pass.

phlptp:
As I understand it the best practice is to not use local variables
so this is the most common way to do it
`if (CMAKE_PROJECT_NAME STREQUAL PROJECT_NAME)`

.githooks/pre-commit
eberleim: I have added filters for these two directories.

phlptp:
I think you will need to rebase on the latest develop to get the tests to pass

phlptp:
would it be possible to check the clang-version before turning the hook on.  Then it would be default off if the version of clang-format wasn't supported or it wasn't available?

eberleim:
@phlptp It currently checks if clang-format exists, and then checks if clang-format's version is greater than 7.0 (minimum version required for our clang-format file), and displays a non-fatal warning if an insufficient version is installed, and the process for assigning a specific clang-format executable. 
When no clang-format is installed, it is disabled automatically.
Summary of defaults:
no clang-format -> no git hook
old clang-format -> warning and no git hook
acceptable clang-format and acceptable git version -> git hook

phlptp:
I think what I don't want is the warning on the CI builds
```
CMake Warning at config/cmake/addGitHooks.cmake:50 (message):
  HELICS git hooks require clang-format version 7.0 or higher.
  Install or update clang-format to use git hooks.
  To select a different clang-format exectuable, update the CLANG_FORMAT
  property in CMake.
Call Stack (most recent call first):
  CMakeLists.txt:1082 (include)
```

That means either turning it off by default via cmake command on the CI builds or not activating it if the clang-format version is too old.

nightlark:
It might be easier to just not enable it if the clang-format version is too old. In this repository there is also the circle ci, drone, various GitHub action workflows, and azure pipelines configurations, and there are dependent repositories like helics-ns3 and HELICS-Examples that also build HELICS.

Another thought is that if someone is on a system with an old version of clang, it might not be managed by them so upgrading could be difficult (university clusters?).

eberleim:
@nightlark That's easy enough. I added the warning, because the default Ubuntu 18.04 version of clang-format is still on 6.0 unless you manually target clang-format-7 or 8, and as such doesn't actually run our format file completely. 
Making the cmake_advanced_dependent_option depend on the clang-format version as well is easy enough. I'd like to keep the warning if the user has specified "ON" and their clang-format version is inadequate, so they'll at least know why it failed. I'll rebase that last change out and adjust.

PR #945, opened 2019-12-04
turn off printing in the cores for the benchmarks and the ones which …

phlptp:
### Summary

If merged this pull request will alter some of the cmake config files to not be HELICS specific
and change the benchmarks to not print if things don't go perfectly.  

### Proposed changes

- addLibZMQ and a few other cmake files are changed to not be HELICS specific
- all benchmarks are set to no print for both brokers and cores.


phlptp:
@nightlark  can you double check the benchmarks to make sure I got them consistent as far as naming goes

phlptp:
@nightlark  can you take a look at this, after this I want to get the big formatting change in.  But this would likely cause conflicts if this PR wasn't merged first.

PR #957, opened 2019-12-06
Fix shellcheck issues on shell scripts

nightlark:
### Summary
If merged this pull request will fix various shellcheck warnings and errors. Mostly quoting variables to make sure spaces in the value work correctly, replacing `which` with the more standard `command -v` to get program paths, and adding `|| exit/return` for failed directory changes.

### Proposed changes
- Fix shellcheck warnings, including some bugs
- Remove no longer used Boost sanitizer and compilation step; copy the Boost headers instead (160 seconds -> 50 seconds for installing dependencies without a cache on Travis)

### DIFF
--- a/.github/workflows/interface_gen_script.sh
+++ b/.github/workflows/interface_gen_script.sh
@@ -1,16 +1,16 @@
-#!/usr/bin/sh
+#!/usr/bin/bash
 
 # Generate Matlab interface using swig matlab
 mkdir build_matlab
-pushd build_matlab
+pushd build_matlab || exit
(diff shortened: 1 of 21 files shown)


phlptp:
Why are the Azure builds still failing?

nightlark:
Probably because the build was for the branch (which was created before the Python 3.4 change got merged into develop) and not the PR. I rebased on develop to get the updated Azure yml file into this branch and now it's using the right version.

.github/workflows/interface_gen_script.sh
```diff
@@ -1,16 +1,16 @@
-#!/usr/bin/sh
+#!/usr/bin/bash
 
 # Generate Matlab interface using swig matlab
 mkdir build_matlab
-pushd build_matlab
+pushd build_matlab || exit
```
phlptp: what does the ||exit do exactly.  I would guess it just just forces the return result of this command to be 0?

phlptp:
are all these scripts running somewhere?

.github/workflows/interface_gen_script.sh
nightlark: The default exit status is 0. The main thing is that if changing directories fails it stops running the rest of the script.

nightlark:
Only one of the git archive related scripts runs as part of the release/nightly build action.

mktempdir.sh is a mystery, but it looks like I committed it for some unknown reason back in May.

The rest all run as part of various CI builds.

phlptp:
Is this ready to merge?

nightlark:
I think it's okay to merge. I got rid of mktempdir.sh since it isn't used anywhere, I think I'd used it to test making a temporary directory in a portable way (macOS mktemp command has quirks that Linux doesn't have).

PR #964, opened 2019-12-09
Tsan failure fixes

phlptp:
### Summary

If merged this pull request will fix some TSAN related issues.  

### Proposed changes

- move the octave build to using the buildenv version so the standalone octave repo can be retired after merging to master.  
- rework the container and concurrency repos to fix some issues
- add some extra thread join processes in case the threads are abnormally terminated at program end.  
- add an extra shared_ptr pass to a callback in case the thread terminates then the this point wouldn't be deleted.


src/helics/core/tcp/TcpHelperClasses.cpp
```diff
@@ -37,8 +37,9 @@ namespace tcp {
             if (!triggerhalt) {
                 socket_.async_receive(
                     asio::buffer(data.data() + residBufferSize, data.size() - residBufferSize),
-                    [this](const std::error_code& error, size_t bytes_transferred) {
-                        handle_read(error, bytes_transferred);
+                    [ptr = shared_from_this()](
+                        const std::error_code& error, size_t bytes_transferred) {
```
nightlark: Slightly wrapping. Is this from the line length?

phlptp: the formatting will automatically wrap from the line length, though I think the algorithm is some penalty function so in some cases it can be slightly longer.    You think this is an issue that we need to tweak?

nightlark: I think in most other cases it would look fine, it's just with the lambda that it looks a bit strange because the lambda body is indented to the same level as the args.

phlptp: I have noticed different version of clang-format do handle lambdas a little different.  Do you think it is something we should tweak.  Changing the variable names can probably shift it around a little if you think it worthwhile.

nightlark: Maybe try a variable name change. If it doesn't work then it's okay, most of the other lambdas I saw didn't wrap like this one.

PR #966, opened 2019-12-09
Add EditorConfig and gitattributes files

nightlark:
### Summary
If merged this pull request will add an EditorConfig file with some basic source code formatting rules (tabs = 4 spaces, newline at the end of the file) for IDEs/editors.

### Proposed changes
- Add an EditorConfig file


kdheepak:
We should also add instructions in the documentation for how to set up extensions if needed to use `.editorconfig` in popular IDEs/editors, such as Visual Studio, Visual Studio Code, Atom, Vim, Emacs, etc.

nightlark:
What about just saying "go to https://editorconfig.org/#download to see if your favorite IDE has built-in editorconfig support or requires a plugin"? Then we aren't maintaining or own version of something that already exists.

phlptp:
looks like this branch might need to be rebased on develop to get the tests to pass properly.

nightlark:
Okay, rebased on develop and cleaned up the commit history.

PR #970, opened 2019-12-10
Build macOS release assets with the macOS 10.10 SDK

nightlark:
### Summary
If merged this pull request will build the macOS installer and shared library release assets using a macOS 10.10 SDK to support older versions of macOS.

### Proposed changes
- Setup a copy of the macOS 10.10 SDK and use it to build macOS release assets
- Cleanup some of the macOS release package settings
- Switch macOS installer CPack generator from TGZ to ZIP

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -248,7 +248,7 @@ add_library(HELICS::compile_flags_target ALIAS compile_flags_target)
 add_library(HELICS::build_flags_target ALIAS build_flags_target)
 
 get_target_property(EXTRA_BUILD_FLAGS build_flags_target INTERFACE_COMPILE_OPTIONS)
-set(HELICS_BUILD_FLAGS_OUTPUT "${CMAKE_CXX_FLAGS} ${CMAKE_CXX_FLAGS_RELEASE} ${CMAKE_STATIC_LINKER_FLAGS} ${EXTRA_BUILD_FLAGS} ") 
+set(HELICS_BUILD_FLAGS_OUTPUT "${CMAKE_CXX_FLAGS} ${CMAKE_CXX_FLAGS_RELEASE} ${CMAKE_STATIC_LINKER_FLAGS} ${EXTRA_BUILD_FLAGS} ")
 
 if(CMAKE_PROJECT_NAME STREQUAL PROJECT_NAME)
     if(NOT USE_LIBCXX)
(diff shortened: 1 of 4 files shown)


phlptp:
Will this add another file to the release assets?

nightlark:
No, the names just change. Darwin => macOS suffix, and the non-shared library to macOS package is a zip file instead of tar.gz file.

phlptp:
So the summary is change the name to be a little more obvious,  use a more common extension and use an older SDK so it works on more systems?

nightlark:
Sounds about right. If there's a place we can archive files, we should create a backup of the older macOS SDKs just in-case something happens to the current GitHub repository they're in.

phlptp:
would it make sense(or even a possibility) to store them in a docker image on dockerhub?

nightlark:
Unfortunately macOS docker just runs a Linux VM (it also isn’t installed by default, workflow steps using a container just fail).

CMakeLists.txt
```diff
@@ -1061,13 +1061,9 @@ if(HELICS_ENABLE_PACKAGE_BUILD)
...
-        set(CPACK_BUNDLE_NAME "libhelics")
-        configure_file(
-            "${HELICS_SOURCE_DIR}/config/Info.plist.in"
-            "${CMAKE_CURRENT_BINARY_DIR}/Info.plist" @ONLY
-        )
-        set(CPACK_BUNDLE_PLIST "${CMAKE_CURRENT_BINARY_DIR}/Info.plist")
-        set(CPACK_BUNDLE_ICON "${HELICS_SOURCE_DIR}/docs/img/HELICS.ico")
+	set(CPACK_PACKAGE_FILE_NAME "${CPACK_PACKAGE_NAME}-${CPACK_PACKAGE_VERSION}-macOS")
+	set(CPACK_DMG_DISABLE_APPLICATIONS_SYMLINK ON)
+	set(CPACK_PACKAGE_ICON "${HELICS_SOURCE_DIR}/docs/img/HELICS.ico")
```
kdheepak: Change to use spaces instead of tabs.

kdheepak:
This is tangentially related but we should add this to the `CMakeLists.txt` file.

```
    if(WIN32)
        set(CPACK_RESOURCE_FILE_README "${HELICS_SOURCE_DIR}\\\\README")
    else(WIN32)
        set(CPACK_RESOURCE_FILE_README "${HELICS_SOURCE_DIR}/README")
    endif(WIN32)
```

kdheepak:
https://github.com/GMLC-TDC/HELICS/blob/3c2af0430fa212108633bf805a9310d3df1271a2/CMakeLists.txt#L952

This should be a description, not just a single word:

```
set(CPACK_PACKAGE_DESCRIPTION_SUMMARY "Hierarchical Engine for Large-scale Infrastructure Co-Simulation (HELICS)")
```

PR #985, opened 2019-12-12
Adds a CMake function to Git Submodules logic to check if all submodules are current

eberleim:
### Summary

If merged this pull request will add a non-blocking CMake function to Git Submodules logic to check if all submodules are current.

### Proposed changes

Uses the `git submodule status --recursive` command to verify all submodules are
up to date with the parent repository, and displays a non-blocking warning if any
module is out of date, as well as the command to update submodules.
Output from git is piped through grep on *nix style machines, and powershell Select-String on Win32 machines to verify all submodules and their dependencies are clean. 
If submodules are out of date, the user is notified with a CMake warning, and the command to update the submodules is displayed.

### DIFF
--- a/config/cmake/updateGitSubmodules.cmake
+++ b/config/cmake/updateGitSubmodules.cmake
@@ -48,3 +48,29 @@ macro(submod_update target)
      message(WARNING "SUBMODULE update has been disabled")
    endif()
 endmacro()
+
+function(check_submodule_status)
+    if (ENABLE_SUBMODULE_UPDATE)
+        execute_process(COMMAND ${GIT_EXECUTABLE} submodule status --recursive
+                WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
+                RESULT_VARIABLE GIT_RESULT
+                OUTPUT_VARIABLE GIT_OUTPUT)
+
+        if (GIT_RESULT)
+            message(WARNING "Automatic submodule verification with `git submodule status --recursive` failed with error ${GIT_RESULT} and output ${GIT_OUTPUT}. Verify submodules before building.")
+        endif ()
+
+        if (WIN32 AND NOT (MSYS OR CYGWIN))
+            execute_process(COMMAND powershell "-c" "echo '${GIT_OUTPUT}' | Select-String -Pattern '^[-\\|+\\|U].*'"
+                    OUTPUT_VARIABLE SUBMODULE_STATUS
+                    )
+        else ()
+            execute_process(COMMAND bash "-c" "echo '${GIT_OUTPUT}' | grep '^[-\\|+\\|U].*'"
+                    OUTPUT_VARIABLE SUBMODULE_STATUS
+                    )
+        endif ()
+        if (NOT "${SUBMODULE_STATUS}" STREQUAL "")
+            message(WARNING "Submodules are not up to date. Update submodules by running `git submodule update --init --recursive` before building HELICS.")
+        endif ()
+    endif ()
+endfunction()
--- a/config/cmake/updateGitSubmodules.cmake
+++ b/config/cmake/updateGitSubmodules.cmake
@@ -12,15 +12,15 @@
 find_package(Git QUIET)
 if(GIT_FOUND AND (GIT_VERSION_STRING VERSION_GREATER "1.5.2"))
     if(EXISTS "${PROJECT_SOURCE_DIR}/.git")
-        option(ENABLE_SUBMODULE_UPDATE "Checkout and update git submodules" ON)
-		mark_as_advanced(ENABLE_SUBMODULE_UPDATE)
+        option(${PROJECT_NAME}_ENABLE_SUBMODULE_UPDATE "Checkout and update git submodules" ON)
+		mark_as_advanced(${PROJECT_NAME}_ENABLE_SUBMODULE_UPDATE)
     else()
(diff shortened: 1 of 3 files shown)


phlptp:
Is there any way to test this on travis or one of the other CI services?

eberleim:
I imagine it must be possible to deliberately checkout an incorrect submodule tag, run cmake, run submodule update, and rerun cmake to verify both the warning and the successful run, but I do not have any experience setting up CI configurations.
Since I wrote wrappers for the change string detection in both bash and powershell, we probably need one of each type of test. 

Who's our resident expert on the CI build processes? It might be worth checking to see if this is a "wrap it into one of the existing builds" or a "create a whole new build specifically to test this."

phlptp:
@nightlark  what is your thoughts on how something to test this might get added to a CI build.  
In the mean time in the next day or two I am going to test this out manually on my windows machine to verify it is working on windows and does what is claimed. If that works I think we can merge it, and work on a CI test later on.

phlptp:
does this actually update the submodules if they are not at the correct commit or just check them?

eberleim:
@phlptp It checks the status, and notifies the user if there are inconsistent submodules, as well as the appropriate command to update them manually. 
A target to run the update could be added if we decide that path.

I made the warning non-blocking for testing purposes, however given that incorrect submodules are likely to break a build anyways, it could easily be changed to a fatal error if we ultimately decide that's the right path forward. Alternately, it could check the status, and if submodules are out of date, run the update all automatically, however my concern with this is it having the potential to swap out user-committed changes without them noticing, so I elected not to do this at the time.

nightlark:
When you say test this, do you mean a test to make sure that the verifying submodule status is working or a test to see if the submodules are up to date when a PR is made?

phlptp:
I was under the impression that this PR automatically updated the modules as well, but if it is just a warning, that isn't as important to test.  I am going to try it out on my system before approving, but the test would probably be more useful if/when it does an automatic update.

nightlark:
It looks like it's just a warning.

Is the check just seeing if the submodule is up to date with the current commit hash for the submodule associated with the checked out HELICS commit, or is it comparing against the latest commit to the submodule repository?

eberleim:
@nightlark This check is for matching hash. It only verifies that you have the same version as the HELICS repo identifies as correct, not that the submodule is in its most up-to-date state with its own repository.

nightlark:
Okay, that's probably fine then.

I guess a test would be cd to a cloned submodule, checkout a very old commit, and then see if running cmake again updates the submodule. Maybe a new job run on drone that just does that.

config/cmake/updateGitSubmodules.cmake
```diff
@@ -47,4 +47,30 @@ macro(submod_update target)
    else()
      message(WARNING "SUBMODULE update has been disabled")
    endif()
-endmacro()
\ No newline at end of file
+endmacro()
+
+function(check_submodule_status)
+    if (ENABLE_SUBMODULE_UPDATE)
```
phlptp: Should this variable be made HELICS specific list ${PROJECT_NAME}_ENABLE_SUBMODULE_UPDATE ?  @nightlark @eberleim

config/cmake/updateGitSubmodules.cmake
```diff
@@ -47,4 +47,30 @@ macro(submod_update target)
...
+            execute_process(COMMAND powershell "-c" "echo '${GIT_OUTPUT}' | Select-String -Pattern '^[-\\|+\\|U].*'"
+                    OUTPUT_VARIABLE SUBMODULE_STATUS
+                    )
+        else ()
+            execute_process(COMMAND bash "-c" "echo '${GIT_OUTPUT}' | grep '^[-\\|+\\|U].*'"
+                    OUTPUT_VARIABLE SUBMODULE_STATUS
+                    )
+        endif ()
+        if (NOT "${SUBMODULE_STATUS}" STREQUAL "")
+            message(WARNING "Submodules are not up to date. Update submodules by running `git submodule update --init --recursive` before building HELICS.")
```
phlptp: This should probably have something more noticeable when running as an update.    It showed up reasonably clear when running in cmake, but when cmake just updated as part of the build it would be easy to miss,  It might need something like **WARNING** in front on the message to make it more noticeable.

phlptp: Is there any way it could specify exactly which submodule was not up to date?

phlptp:
Even after running submodule update the warning doesn't seem to turn off.    Is this doing a full recursive check on the submodules themselves.  If so that is probably not what we want.  We only want to go one layer deep.  Since some of the submodules themselves have optional submodules which are used for testing which we don't really need or want to download.

eberleim:
@phlptp The submodule check will no longer perform recursive depth check, and will now print out a list of the hashes and names of out-of-date submodules. 
I did not change the `ENABLE_SUBMODULE_UPDATE` variable name, but we can make that additional change if we feel it is meaningful.

phlptp:
I think it works a little better now.  @nightlark  what do you think of the cmake variable name `${PROJECT_NAME}_ENABLE_SUBMODULE_UPDATE` vs `ENABLE_SUBMODULE_UPDATE`  

Either way this option should be documented in https://github.com/GMLC-TDC/HELICS/blob/master/docs/installation/helics_cmake_options.md

nightlark:
In general I prefer options with a unique prefix or namespace to make sure it doesn't conflict if someone tries to integrate it with a project that uses an identical non-prefixed name.

phlptp:
Thanks!!

Issue #906, opened 2019-11-14
Code coverage to 75%


phlptp:
I would like to set a goal of 75% code coverage for the 2.4 release.  

Ultimately I would like to 95% but I don't know that we can get there by the next release on top of everything else, so this is an intermediate goal which I think might be doable in the time frame.

phlptp:
@nightlark @kdheepak   Can you take a look through codecov results and maybe pick out a couple places we should add some coverage, or places that are showing up that wouldn't need to be covered by HELICS tests.

kdheepak:
I think we should aim to have complete code coverage in the Python and Julia interfaces first. Maybe that means getting to 100% code coverage in the C interface in the short run.

phlptp:
What kind of automated ways do we have to measure and track the coverage for Julia and Python interfaces?  I don't think we measuring coverage in those interfaces currently.

phlptp:
as of #990 we are at 66.68%

phlptp:
#1024 increased to 71.49%

phlptp:
#1082 increased to 72.5%
#1088 increased to 73.2%

kdheepak:
I think for Python it's harder to do. But for Julia, I'm tracking it here:

[![Codecov](https://img.shields.io/codecov/c/github/gmlc-tdc/HELICS.jl.svg)](https://codecov.io/gh/GMLC-TDC/HELICS.jl) [![Coverage Status](https://coveralls.io/repos/github/GMLC-TDC/HELICS.jl/badge.svg?branch=HEAD)](https://coveralls.io/github/GMLC-TDC/HELICS.jl?branch=HEAD)

phlptp:
#1158 to 75.04%

PR #990, opened 2019-12-17
Evil tests shared lib

phlptp:
### Summary

If merged this pull request will update the code coverage flags to better reflect actual coverage and add a set of "evil library tests"  which give bad input to all the shared library functions to test susceptibility to mistaken input and null values.  It also calls all the functions and therefore bumps up the coverage a couple %

### Proposed changes

- add evil input tests
- change the coverage flags to get a little more accuracy


phlptp:
make a little progress for #906

tests/helics/shared_library/comboConfigTests.cpp
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright � 2017-2019,
+Copyright © 2017-2019,
```
nightlark:
```suggestion
Copyright (c) 2017-2019,
```

tests/helics/coreTypeLists.hpp
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright � 2017-2019,
+Copyright © 2017-2019,
```
nightlark:
```suggestion
Copyright (c) 2017-2019,
```

src/helics/shared_api_library/MessageFederateExport.cpp
```diff
@@ -402,6 +403,7 @@ helics_message_object helicsEndpointGetMessageObject(helics_endpoint endpoint)
     auto message = endObj->endPtr->getMessage();
 
     if (message) {
+        message->messageValidation = messageKeyCode;
```
nightlark: What is the messageValidation trying to detect?

phlptp: the most common case would be if someone passed a federate in what should have been a message object.  So all the objects have a coded sequence in one of the fields to have a first order check against an incorrect argument being passed.  The most likely of which is a different helics object passed incorrectly, since the compiler can't detect that in C like it would in C++, we need a run time check.

nightlark: Okay, so it's not networking related.

phlptp: no, It is using a gap in the structure and the rest of HELICS will ignore it.  Primarily for shared library to detect mistakes at the first level.  All the validation keys in the object, are just there for object checks.

Issue #842, opened 2019-10-09
Toml Support


phlptp:
It looks like the current toml library we are using is not being updated, at least it hasn't been updated in quite a while, and is not compliant with the latest toml standard.  

It might be time to move to a different library perhaps [toml11](https://github.com/ToruNiina/toml11)

not entirely sure how much work this would be.   Given the nature of this library we would probably would move it to a submodule vs direct inclusion.

PR #994, opened 2019-12-20
Toml11 conversion

phlptp:
### Summary

If merged this pull request will convert HELICS to use toml11 library which is being actively maintained and updated unlike the previous toml library.  

### Proposed changes

- update toml library to use toml11 via a submodule.
- remove toml.h
- update some of the flags to remove some warnings.


phlptp:
address issue #842

phlptp:
need to update the contributors before merging

CONTRIBUTORS.md
```diff
@@ -47,8 +47,8 @@ If you would like to contribute to the HELICS project see [CONTRIBUTING](CONTRIB
...
 
-### [tinytoml](https://github.com/mayah/tinytoml)
-  tinytoml is used for parsing toml files.  tinytoml is licensed under [BSD 2-clause](https://github.com/mayah/tinytoml/blob/master/LICENSE) license. The header file is included in HELICS source.
+### [toml11](https://github.com/ToruNiina/toml11)
+  toml11 is a C\++11 header-only toml parser/encoder depending only on C++ standard library.  Compatible to the latest version of TOML v0.5.0. toml11 is licensed under [MIT](https://github.com/ToruNiina/toml11/blob/master/LICENSE) license. HELICS includes it through a submodule of a library [fork](https://github.com/GMLC-TDC/toml11) until the changes can be worked upstream.
```
nightlark:
```suggestion
  toml11 is a C\++11 header-only toml parser/encoder depending only on the C++ standard library.  Compatible with the latest version of TOML v0.5.0. toml11 is licensed under the [MIT](https://github.com/ToruNiina/toml11/blob/master/LICENSE) license. HELICS includes it through a submodule of a library [fork](https://github.com/GMLC-TDC/toml11) until the changes can be merged upstream.  
```

src/helics/application_api/Federate.cpp
```diff
@@ -834,17 +834,17 @@ void Federate::registerFilterInterfacesJson(const std::string& jsonString)
...
     catch (const std::invalid_argument& ia) {
         throw(helics::InvalidParameter(ia.what()));
     }
 
-    auto filts = doc.find("filters");
-    if (filts != nullptr) {
-        auto& filtArray = filts->as<toml::Array>();
+    if (isMember(doc, "filters")) {
+        auto filts = toml::find(doc, "filters");
+        auto& filtArray = filts.as_array();
```
nightlark: Should this have a check to make sure filts is an array?

src/helics/application_api/ValueFederate.cpp
```diff
@@ -281,9 +281,10 @@ void ValueFederate::registerValueInterfacesToml(const std::string& tomlString)
     }
     bool defaultGlobal = false;
     replaceIfMember(doc, "defaultglobal", defaultGlobal);
-    auto pubs = doc.find("publications");
-    if (pubs != nullptr) {
-        auto& pubArray = pubs->as<toml::Array>();
+
+    if (isMember(doc, "publications")) {
+        auto pubs = toml::find(doc, "publications");
+        auto& pubArray = pubs.as_array();
```
nightlark: Should publications, subscriptions, and inputs (and MessageFederate endpoints) have checks to make sure the toml element is an array?

src/helics/common/TomlProcessingFunctions.cpp
```diff
@@ -28,68 +28,56 @@ toml::Value loadToml(const std::string& tomlString)
...
-            if (val->is<int64_t>()) {
-                return {val->as<int64_t>(), defaultUnits};
+        toml::value emptyVal;
+        auto val = toml::find_or(timeElement, "value", emptyVal);
+        if (!val.is_uninitialized()) {
+            if (val.is_integer()) {
+                return {val.as_integer(), defaultUnits};
             }
-            return {val->as<double>() * toSecondMultiplier(defaultUnits)};
+            return {val.as_floating() * toSecondMultiplier(defaultUnits)};
```
nightlark: Should there be a check here to make sure the value is a floating point number?

src/helics/core/fileConnections.hpp
```diff
@@ -53,14 +54,13 @@ void makeConnectionsToml(brkX* brk, const std::string& file)
             }
         }
     }
-    auto filts = doc.find("filters");
-    if (filts != nullptr) {
-        auto& filtArray = filts->as<toml::Array>();
+    auto filts = toml::find_or(doc, "filters", uVal);
+    if (!filts.is_uninitialized()) {
+        auto& filtArray = filts.as_array();
```
nightlark: Same as most of the others -- looks like another place that might need an is_array check.

src/helics/application_api/Federate.cpp
phlptp: It didn't before, but I am not entirely clear whether everything is an array or it has to be a purposeful array in toml11.   Same comment applies to the other places

src/helics/common/TomlProcessingFunctions.cpp
phlptp: if it isn't that is an incorrectly specified file, but the same was true before.

nightlark: Yea, I noticed a few other places that had checks. I don't know if a linting tool for input files is needed to catch issues with fields that are given unexpected types of values.

src/helics/application_api/Federate.cpp
nightlark: That's an interesting question. Reading the toml spec seems to imply that arrays are values that use `[]`, but there isn't anything in the spec or toml11 that mentions implicitly conversion of a value to an array.

 Since some other places already had checks, I having it be consistent is good.

phlptp: Anyway I put in checks that throw an exception if it is not array in all the locations.  I also added a function that does the conversion to string regardless of the underlying type.  It doesn't look the toml11 does the automatic conversion.

PR #1000, opened 2020-01-02
Add an MSYS build to Travis CI

nightlark:
### Summary
If merged this pull request will add an MSYS build to Travis and moves some of the build steps to separate scripts so that the right PATH can be setup for an MSYS build. It caches some chocolatey files for the msys install, with whatever minimal benefit that might give.

This build runs under similar conditions to the MinGW build:
- `develop` nightly builds
- pushes + PRs to `master`
- branch names containing `msys`
- commits with messages containing `msys`

### Proposed changes
- Add an MSYS build to the Travis CI build matrix
- Move the Travis cmake configure + build commands to a separate script
- Move the core ci test command to a separate script
- Add +x bit to sanitizer_tests.sh permissions


.travis.yml
```diff
@@ -172,7 +180,32 @@ jobs:
...
-      
+    
+    - os: windows
+      name: "MSYS"
+      if: (branch = develop AND type IN (cron)) OR (branch = master AND type IN (push, pull_request)) OR (branch =~ /(msys)/) OR (commit_message =~ /(msys)/)
+      env:
+        - DISABLE_INTERFACES="Python,Java"
+        - SKIP_TEST_RUN=true
+        - CORE_CI_TEST=true
+        - DISABLE_EXAMPLES=true
```
phlptp: any particular reason to disable the examples?

.travis.yml
```diff
@@ -6,9 +6,17 @@ cache:
...
     - build/libs
+    - $HOME/AppData/Local/Temp/chocolatey
+    - /c/tools/msys64
 
 before_cache:
   - ${TRAVIS_BUILD_DIR}/scripts/cleanup-cmake.sh
+  - |
+    case $TRAVIS_OS_NAME in
+      windows)
+        $SHELL_CMD pacman --sync --clean --noconfirm
```
phlptp: is SHELL_CMD something you defined or is that a predefined variable

.travis.yml
```diff
@@ -249,17 +279,17 @@ script:
   - |
     if [[ "$SKIP_TEST_RUN" != "true" ]]; then
       echo "travis_wait ../scripts/run-ci-tests.sh ${CI_TEST_FLAGS}"
-      travis_wait ../scripts/run-ci-tests.sh ${CI_TEST_FLAGS}
+      travis_wait $SHELL_CMD ../scripts/run-ci-tests.sh ${CI_TEST_FLAGS}
```
phlptp: are we only using the scripts on the MSYS build?

.travis.yml
nightlark: No, just similar settings to the MinGW build.

.travis.yml
nightlark: Scripts are used for all of the builds. The commands used for builds/tests are the same other than using a SHELL_CMD wrapper (if defined) and adding the MSYS bin folder to the PATH.

nightlark: This entire script section in the yaml could have been put in a .sh file, but I think that would make Travis not display as much information on how long subcommands take to run.

.travis.yml
nightlark: It's defined in the install script for the MSYS build. It's undefined/empty for the other builds.

Issue #1004, opened 2020-01-03
Can't build Java interface with MSYS2


temcdrm:
**Describe the bug**
The "make" process stops on libJNIHelics.dll as shown

`[100%] Linking C shared library libJNIhelics.dll`
`-- Configuring done`
`-- Generating done`
`-- Build files have been written to: C:/src/HELICS-src/build/interfaces/java/buildjar`
`Building jar file`
`The system cannot find the file specified`
`CMake Error: Generator: execution of make failed. Make command was: nmake /nologo helics &&`
`make[2]: *** [interfaces/java/CMakeFiles/JNIhelics.dir/build.make:90: interfaces/java/libJNIhelics.dll] Error 1`
`make[2]: *** Deleting file 'interfaces/java/libJNIhelics.dll'`
`make[1]: *** [CMakeFiles/Makefile2:1939: interfaces/java/CMakeFiles/JNIhelics.dir/all] Error 2`
`make: *** [Makefile:141: all] Error 2`

**What is the expected behavior?**
 Would like to ship some Java agent examples in transactive energy simulation

**To Reproduce**
From the build subdirectory:
`cmake -G "MSYS Makefiles" -DCMAKE_INSTALL_PREFIX=/usr/local -DBUILD_SHARED_LIBS=ON -DBUILD_PYTHON_INTERFACE=ON -DBUILD_JAVA_INTERFACE=ON -DJAVA_AWT_INCLUDE_PATH=NotNeeded -DHELICS_DISABLE_BOOST=ON -DCMAKE_BUILD_TYPE=Release ..`
`make`

**Environment (please complete the following information):**
 - Operating System: [Windows MSYS2]
 - Language Extension: [Java]
 - what compiler or setup process did you use: gcc/g++ 9.2.0 and Java 11.0.5
 - HELICS version: master branch, pulled today

**Additional context and information**
This build step used to work in MSYS2. At this time, `pacman -Su` or a new installation of MSYS2 comes with gcc9.  Also, older versions of Java aren't so readily available now.

phlptp:
I have never tried building The JAVA interface on MSYS2 before, I will look into it.

What is `-DJAVA_AWT_INCLUDE_PATH=NotNeeded`  for?

phlptp:
Alright I tried it and it seemed to work, but I think it is picking up my windows native jdk rather than an msys2 version of java so that probably isn't a good test.

temcdrm:
[...]

PR #1006, opened 2020-01-04
Java msys

phlptp:
### Summary

If merged this pull request will fix issue #1004 by modifying the cmake files of the java interaface and hacking a fix in CMAKE for MSYS.  

### Proposed changes

- propagate the generator to the cmake build for the jar file
- propagate the JAVA related variables
- make MSYS think it is CYGWIN for purposes of add_jar


nightlark:
I think this looks okay. Do you remember the reason building the jar uses a separate CMake build?

phlptp:
The main reason is that if we building with swig the files are dynamically generated at build time from swig in a separate step, so we can't really generate any sort of file list or dependency checks ahead of time.  So to make the process the same for both the swig and non-swig cases we build the jar in a separate generate/build process that only gets run at build time.

PR #1008, opened 2020-01-06
Add .git directory detection to fix fatal error

### LINKED ISSUE #996: fatal: --local can only be used inside a git repository
**Describe the bug**
cmake is showing a 
`fatal: --local can only be used inside a git repository`  

There really shouldn't be any warning or error like this showing up in the build files

**Environment (please complete the following information):**
 - Operating System: Windows

**Additional context and information**
I think it is related to the recent git hooks changes

eberleim:
### Summary

Resolves #996 

### Proposed changes

- Add logic to detect CMake is running inside a git repository prior to attempting to toggle git hooks on or off.


phlptp:
How is this change causing the nonlings example test to fail?

eberleim:
@phlptp That's an excellent question, because it shouldn't be. I'll rebase onto the most recent develop branch in case I missed something when I branched to fix this, and push that up, but it really shouldn't cause any tests to fail, as it doesn't change any code or actual build properties.

phlptp:
@nightlark  you got any ideas why this causing a failure pretty consistently in the nonlings test?

nightlark:
I don't know. It is interesting that it is just the branch builds and not the PR/merge builds.

phlptp:
hmm thats kind of funny.  The test is failing since due to the name fatal-error in the branch.  The test just prints out the helics version and is just looking to make sure the exit code is 0 and no errors showed up.  Well when the git branch has fatal-error in it it triggers the test to fail.  Pretty sure it is working properly.

phlptp:
I am just going to merge it since it most likely would have passed if not for the branch name

Issue #433, opened 2018-08-15
MSYS 2 package for HELICS


phlptp:
* **I'm submitting a ...**
  - [ x] feature request

### Feature Request / Support Question

Add Helics to the MSYS 2 Repos 
[adding packages](https://github.com/msys2/msys2/wiki/Creating-Packages)

phlptp:
There is a PR to get this into the main repository for MSYS2

https://github.com/msys2/MINGW-packages/pull/6173

phlptp:
The package for 2.4.0 has been merged into the main repo.  It looks like it is in the next phase of review before (hopefully) becoming available in the package repositories for MSYS2

phlptp:
The package is now available in the msys2 repositories 
```
pacman -Ss helics
```
to see it.  
I am going to add some documentation and instructions to the installation pages for msys2 and when merged will close this issue

PR #1009, opened 2020-01-07
make a few minor tweaks to the build to make the configuration work w…

phlptp:
…ith the package builder/installer for msys2.  The main fix was turning off the installation of zmq libraries on MSYS and CYGWIN, that was interfering with the default installation.

### Summary

If merged this pull request will make some tweaks to the installation for MSYS to fix some issues from building an MSYS2 Mingw package

### Proposed changes

- turn off the installation of libzmq static lib in msys2 and cygwin
- update the package recipe


config/packaging/mingw-w64-helics/PKGBUILD
```diff
@@ -3,27 +3,26 @@
 _realname=helics
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
-pkgver=2.0.0
-pkgrel=1
-pkgdesc="A C++ library for co-simulation"
+pkgver=2.3.1
+pkgrel=3
+pkgdesc="Hiearchical Engine for large-scale infrastructure co-simulation (mingw-w64)"
```
nightlark:
```suggestion
pkgdesc="Hiearchical Engine for Large-scale Infrastructure Co-Simulation (mingw-w64)"
```

config/packaging/mingw-w64-helics/PKGBUILD
```diff
@@ -3,27 +3,26 @@
 _realname=helics
 pkgbase=mingw-w64-${_realname}
 pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
-pkgver=2.0.0
-pkgrel=1
-pkgdesc="A C++ library for co-simulation"
+pkgver=2.3.1
+pkgrel=3
```
nightlark: This should probably be 1, unless there have been previous releases of a 2.3.1 package.

```suggestion
pkgrel=1
```

phlptp: probably right, I didn't really change that from the template I was using

phlptp:
at some point I need to test the recipe on a fresh install of MSYS2 to make sure the dependencies are correct.  Any ideas on how to go about that.  I don't really want to erase my own environment.    I wonder if I can have two installs of MSYS?

nightlark:
Maybe if the installer doesn't change already set environment variables a second install could work. Or try a Windows 10 VM; MS has some 90-day VMs with various edge versions installed that are pretty minimal.

config/packaging/mingw-w64-helics/PKGBUILD
```diff
@@ -3,27 +3,27 @@
...
+pkgver=2.3.1
 pkgrel=1
-pkgdesc="A C++ library for co-simulation"
+pkgdesc="Hiearchical Engine for Large-scale Infrastructure Co-Simulation (mingw-w64)"
 arch=('any')
 url="https://github.com/GMLC-TDC/HELICS"
-license=('BSD')
+license=('BSD-3-clause')
+depends=("${MINGW_PACKAGE_PREFIX}-zeromq"
+         "${MINGW_PACKAGE_PREFIX}-python3")
```
nightlark: The python3 package could be added to optdepends instead, so that users using the shared libraries directly can choose if they want to install python.

phlptp: I was debating removing the python stuff completely.  I think it if we could get `pip install` working on MSYS2 then there would be no need to build the python interface as part of this package since that could be loaded if needed through `pip`

phlptp: This PR was more about making some changes in the HELICS build, than about finalizing the msys2 package build script.  That script has to work its way through the MSYS2 checking gauntlet yet, so I expect a number of additional changes before it is finalized.  So I am not concerned if the PKGBUILD is not perfect yet.

nightlark: Is MSYS2 able to use typical Windows dll's?

phlptp: everything on the windows path is in the MSYS2 path, so as long as names don't intersect with the msys2 libraries,  yes

phlptp: I also will not submit the PR to the mingw repos until we get HELICS 2.4 released,  hopefully by then we will have a clear idea of what we can do with pip install.

phlptp:
This is a step towards #433

PR #1012, opened 2020-01-08
Multinode PHOLD benchmark

nightlark:
### Summary
If merged this pull request will add support for running a multinode PHOLD benchmark.

Results should be deterministic based on the seed used, though they differ from the standalone phold benchmark.

Differences in time measured by federates within small (max 8 federate) benchmark runs seem to be within a range of 0.5ms, though most of the federates are within 0.3ms of each other. For 1-8 federates the differences in timing seem to be consistent. With a 100 federate run, timing differences were over a range of ~2ms.

After comparing steady_clock vs high_precision_clock results, it looked like the results from steady_clock were similar to high_precision_clock (on Quartz, the high_precision_clock is not steady).

### Proposed changes
- Add multinode support to the PHOLD benchmark
- Create a benchmark framework-agnostic header for printing system information
- Add an sbatch script for launching a PHOLD benchmark run on multiple nodes


phlptp:
Is there another script that executes the .pbatch file with different arguments to generate a full suite of test results?

phlptp:
How would we set this up on a cloud cluster with different core types?

nightlark:
The slurm sbatch files require slurm to be installed. A cloud cluster would either need to be set up with slurm to manage jobs. Or a bash script can be written that ssh's into each node in the cloud cluster to start federates (requires knowing the hostnames/ip addresses, and the script would be specifically for that cluster).

phlptp:
Do you have some results from initial runs?

nightlark:
Added a bash script that will run sbatch to submit a set of jobs (matrix?) using the zmq, zmqss, tcp, tcpss, and udp core types on 1, 2, 4, and 8 nodes. Optionally, arguments can be given to the script and they will be treated as an array that takes the place of the number of nodes array. The script exiting does not mean any of the benchmarks have run, just that they are in the slurm job queue; telling it to run with 100 nodes may take a few hours before the benchmark starts running depending on how many other people are also using the cluster.

As it is, the sbatch scripts aren't completely portable to different clusters and there isn't really a way around requiring some slurm basics. In particular two things come to mind:
- making sure they launch the job with the right account/bank for charging CPU time
- making sure the job is submitted to the right queue/partition, which will depend on the cluster

I don't think there a way around some cluster specific configuration being required to launch a multinode benchmark run.

phlptp:
I mostly wanted an example of it being run.  Some notes about it in the documentation might be appropriate of what might need to be done to run on different clusters?

nightlark:
Add a multinode benchmark section to the docs mentioning the doing a multinode benchmark run will depend on their environment, and that the sbatch files are setup to work on LC's Quartz cluster.

Here's a copy of the output from a run of the jobs submitted by the phold sbatch submitter script:
https://github.com/GMLC-TDC/helics_benchmark_results/tree/multinode-results/multinode_benchmark_results/2020-01-08

The output from each node includes information about the node (there are some heterogeneous clusters).

benchmarks/helics/PholdFederate.hpp
```diff
@@ -47,18 +47,27 @@ class PholdFederate {
     std::uniform_real_distribution<double> rand_uniform_double;
     std::uniform_int_distribution<unsigned int> rand_uniform_int;
 
+    // callbacks for instrumenting the code
```
phlptp: Could you envision doing something like this for some of the other benchmarks, potentially making a base class or something of that nature?

nightlark: Yes, there's parts that would make sense to pull out into a base class that could provide default callbacks and command line arguments.

PR #1014, opened 2020-01-10
Set HELICS_SWIG_INCLUDE_DIRS variable for standalone interface builds

nightlark:
### Summary
If merged this pull request will create a HELICS_SWIG_INCLUDE_DIRS variable that is populated with the directories needed to generate an interface using the installed swig files.

### Proposed changes
- Create a HELICS_SWIG_INCLUDE_DIRS variable with all of the directories needed for generating a swig library (set in two places, to work for current interface build process and a standalone interface build that uses find_package(HELICS))
- Move the install line for the helics.i swig file to the interfaces CMake file
- Remove output.i, a file generated by doxy2swig that isn't used


CMakeLists.txt
```diff
@@ -722,15 +722,6 @@ if(CMAKE_PROJECT_NAME STREQUAL PROJECT_NAME)
 
     add_subdirectory(interfaces)
 
-    if(NOT HELICS_BINARY_ONLY_INSTALL AND NOT HELICS_DISABLE_C_SHARED_LIB)
```
phlptp: This part just got moved to the interfaces cmake correct?

interfaces/CMakeLists.txt
```diff
@@ -133,10 +133,23 @@ if(HELICS_ENABLE_SWIG OR swig_required)
...
+        COMPONENT swig
+    )
+endif()
+
+# cmake-format: off
+# add a helper variable for standalone interface swig builds
+set(HELICS_SWIG_INCLUDE_DIRS "${CMAKE_CURRENT_SOURCE_DIR}" "${HELICS_SOURCE_DIR}/src/helics/shared_api_library" "${HELICS_BINARY_DIR}/src/helics/shared_api_library")
+# cmake-format: on
+
+# TODO: remove this after all of the interfaces work as standalone builds
```
phlptp: Is this something we are planning to work on?

CMakeLists.txt
nightlark: Yes, it just got moved.

interfaces/CMakeLists.txt
nightlark: For now just the Python interface to get pip builds working. The changes to the other interfaces should be easy to make later on.

phlptp: Sounds good the others can wait

PR #1020, opened 2020-01-13
Add support for standalone Python3 interface builds

nightlark:
### Summary
If merged this pull request will refactor the Python3 interface build system and add support for standalone builds. Standalone builds (configure+swig+compile) can take under 10 seconds.

RPATH findings:
- Debian packaging suggests not using it at all
- $ORIGIN is the way to handle relative paths on Linux
- For Python wheels it doesn't really matter... part of fixing up a binary wheel on Linux involves copying in needed shared libraries and adjusting the rpath

When building as a subproject of a HELICS build, the rpath using an absolute path for non-system install locations as part of the rpath still works.

For standalone builds of the interface, the rpath will just be relative paths using $ORIGIN, which means it will work the same cases as a precompiled binary would -- it needs to be installed in a prefix folder that has HELICS installed. 

### Proposed changes
- Add standalone build support to the Python3 interface
- Clean-up the CMake build system files for the Python 3 interface, reduce duplication and add comments
- Fix bug with the version argument location for the SWIG find_package call use by the other interfaces (resulted in old SWIG versions below the minimum getting found)
- Remove ./* from the rpath in favor of $ORIGIN/* paths instead
- Give the path to the shared_api_library subfolder for SWIG includes in HELICSConfig.cmake
- Copying shared libraries to the python interface folder only happens on Windows during subproject builds (unix systems have rpath); things can get fixed up with Windows


interfaces/python/CMakeLists.txt
```diff
@@ -7,284 +7,272 @@
...
+endif()
+
+# CMake UseSWIG doesn't properly propagate include dirs when using target_link_libraries
+# This is also the only way that works with CMake < 3.12
+include_directories(${HELICS_SWIG_INCLUDE_DIRS})
+
+###############################
+# UseSWIG feature support
+###############################
+# This can be removed in HELICS 3 if the minimum CMake version is >= 3.12
```
phlptp: I think at least for the first while the minimum cmake will move to 3.10 for HELICS 3.  I want to try to keep it able to build natively and easily on Ubuntu 18.04.

interfaces/python/helicsPython.i
```diff
@@ -3,7 +3,7 @@
 %include "cstring.i"
 %include "python_maps.i"
 
-%include "../helics.i"
+%include "helics.i"
```
phlptp: Did we move the file or is that now on the include path for swig?

interfaces/python/CMakeLists.txt
```diff
@@ -7,284 +7,272 @@
 # SPDX-License-Identifier: BSD-3-Clause
 # ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-if(HELICS_SWIG_GENERATE_INTERFACE_FILES_ONLY)
-    include(${CMAKE_CURRENT_SOURCE_DIR}/pythonSwigGenerateOnly.cmake)
-else()
+cmake_minimum_required(VERSION 3.4)
+project(helicsPYTHON)
```
phlptp: should this have a version associated with it?

interfaces/python/CMakeLists.txt
```diff
@@ -7,284 +7,272 @@
...
+        list(APPEND CMAKE_INSTALL_RPATH "@executable_path/.")
+    endif()
+endif()
+
+###############################
+# Find required dependencies
+###############################
+# Find HELICS
+if(NOT hasParent OR NOT TARGET HELICS::helicsSharedLib)
+    find_package(HELICS 2 REQUIRED)
```
phlptp: Is there a minimum version of HELICS this will work with.  like 2.3 or something like that?  Is it going to work with 2.0 for example.

interfaces/python/CMakeLists.txt
```diff
@@ -7,284 +7,272 @@
...
+endif()
+
+if(HELICS_ENABLE_SWIG OR HELICS_SWIG_GENERATE_INTERFACE_FILES_ONLY)
+    # Find SWIG
+    if(NOT DEFINED SWIG_FOUND OR SWIG_VERSION VERSION_LESS 3)
+        if(POLICY CMP0078)
+            cmake_policy(SET CMP0078 OLD)
```
phlptp: In HELICS 3.0 we really need to clean up some of these policy issues.

interfaces/python/helicsPython.i
nightlark: It's on the include path for swig.

interfaces/python/CMakeLists.txt
nightlark: 2.2.2 is the first version that bundled the helics.i file, but in theory it could work with 2.0 if they provided a path to a valid helics.i file in HELICS_SWIG_helics.i_FILE and set HELICS_SWIG_INCLUDE_DIRS to the right include paths.

phlptp: Does it make sense to leave it at 2 or should we set it to something like 2.3?

interfaces/python/CMakeLists.txt
nightlark: Since the only way it is distributed is in the HELICS repository and source tarballs, I think that gives pretty strong ties to a particular HELICS version. If we add a separate version for the Python interface we'd need to figure out what version numbers to use (every time swig updates the .c wrapper file?).

interfaces/python/CMakeLists.txt
nightlark: I'd rather leave it more open to start -- though hoping there aren't any users still on 2.0 might be optimistic.

interfaces/python/CMakeLists.txt
phlptp: I was assuming it would be the same the HELICS_VERSION, I just don't want CMAKE to spit out a warning if nothing is there.  Not sure if it does that on Newer cmake or not.  If not I am fine leaving it off.

nightlark: The docs say the VERSION argument is optional and needs CMP0048 set to new — it mentions 3.16.2+ will output a warning if VERSION is used and the policy is not set (seems to be for compatibility with pre-3.0 projects that manage version variables themselves).

kdheepak:
Should we add instructions (maybe just a single line or a small section) to the docs on how to create a standalone build?

nightlark:
That could be useful for people trying to use Python versions we don't provide packages for.

PR #1021, opened 2020-01-14
Combine NSIS+ZIP release job and add CPU architecture to Linux/macOS release package names

nightlark:
### Summary
If merged this pull request will add the CPU architecture to the file name for the Linux and macOS release packages to distinguish between different CPUs and 32/64 bit processors. It also combines the NSIS+ZIP installer creation on Windows into a single job, because compiling takes more time than packaging.

After this change, the set of files on a release is:
Helics-2.3.1-macOS-x86_64.zip (11.5 MB)
Helics-2.3.1-msvc2017-win64.zip (7.9 MB)
Helics-2.3.1-msvc2019-win64.zip (7.09 MB)
Helics-2.3.1-win64.exe (6.59 MB)
Helics-2.3.1-win64.zip (11.3 MB)
Helics-shared-2.3.1-Linux-x86_64.tar.gz (2.82 MB)
Helics-shared-2.3.1-macOS-x86_64.tar.gz (1.92 MB)
Helics-shared-2.3.1-win32.tar.gz (1.56 MB)
Helics-shared-2.3.1-win64.tar.gz (1.79 MB)
Helics-v2.3.1-SHA-256.txt (975 Bytes)
Helics-v2.3.1-source.tar.gz (17.8 MB)

### Proposed changes
- Add CPU architecture to the Linux and macOS release package names
- Combine the NSIS and ZIP installer packaging for Windows into a single job


CMakeLists.txt
```diff
@@ -1071,10 +1071,12 @@ if(HELICS_ENABLE_PACKAGE_BUILD)
     endif(WIN32)
 
     if(APPLE)
-        set(CPACK_PACKAGE_FILE_NAME "${CPACK_PACKAGE_NAME}-${CPACK_PACKAGE_VERSION}-macOS")
+        set(CPACK_SYSTEM_NAME "macOS-${CMAKE_SYSTEM_PROCESSOR}")
```
phlptp: is CPACK_SYSTEM_NAME automatically used to generate CPACK_PACKAGE_FILE_NAME?

nightlark: Yes.

.github/workflows/release-build.yml
```diff
@@ -98,18 +98,12 @@ jobs:
...
             os: windows-latest
             arch: x64
-            cpack_gen: NSIS
-            cmake_gen: 'Visual Studio 16 2019'
-            msvc_ver: 'msvc2019'
-          - id: windows-x64-zip
-            os: windows-latest
-            arch: x64
-            cpack_gen: ZIP
+            cpack_gen: NSIS;ZIP
```
phlptp: cpack can do more than one at once? or is this used by a script some where?

nightlark: CPack takes a list for the generator(s) to use -- the default on Linux (if not set by an argument or in the CMake project) is creating tgz, stgz, and (tbz2?) archives when cpack is run.

PR #1023, opened 2020-01-14
Remove dependency on Python for generating the swig interface

nightlark:
### Summary
If merged this pull request will remove the dependency on Python when only generating updated interface files with swig.

### Proposed changes
- Handle the swig generate only option before finding Python (removes dependency on Python)
- Include the directory helics.i is in for swig generate only builds so that it can work standalone
- Add a note to the docs mentioning support for building the Python3 interface using an existing HELICS install

### DIFF
--- a/docs/installation/language.md
+++ b/docs/installation/language.md
@@ -28,8 +28,10 @@ In [1]: import helics
 
 In [2]: helics.helicsGetVersion()
 Out[2]: 'x.x.x (XXXX-XX-XX)'
-
 ```
+
+Note: If you already have a HELICS installation with the C shared library, it is possible to build the Python3 interface
+using the `interfaces/python` folder as a standalone CMake project. This can be much faster than rebuilding HELICS.
 
 ## HELICS with Python2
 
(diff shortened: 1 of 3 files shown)


docs/installation/language.md
```diff
@@ -28,9 +28,11 @@ In [1]: import helics
 
 In [2]: helics.helicsGetVersion()
 Out[2]: 'x.x.x (XXXX-XX-XX)'
-
 ```
 
+Note: If you already have a HELICS installation with the C shared library, it is possible to build the Python3 interface
+using the `interfaces/python` folder as a standalone CMake project. This can be much faster than rebuilding HELICS.
+
```
kdheepak: Can you add the install command here?

kdheepak:
If we can get this to work with Python2 that would allow us to drop Python2 support in the future and tell people if they need to they can just install it using the standalone method.

docs/installation/language.md
nightlark: Would something like “e.g. `cmake <path to interfaces/python folder>`”  be enough to make it clear what it means by using the folder as a standalone cmake project? For a system directory install of HELICS that will work, otherwise the default CMake error messages will tell them what variables can get set to make it find a custom HELICS install.

nightlark:
@kdheepak do you know if there is anything for SWIG type maps that is unique to Python 3 that would break generating the Python 2 interface if it were ever added to our Python type maps file?

nightlark:
#1040 combines the Python 2 and 3 build systems, adding an option to build a Python 2 compatible interface separate from building HELICS.

PR #1029, opened 2020-01-16
Make sure the python module can find a copy of helicsSharedLib

nightlark:
### Summary
If merged this pull request will put a copy of the helicsSharedLib file in the same folder as the Python module was built with on Windows, because it doesn't have anything like rpath.

### Proposed changes
- Copy helicsSharedLib on Windows to the same folder the Python module is in

### DIFF
--- a/interfaces/python/CMakeLists.txt
+++ b/interfaces/python/CMakeLists.txt
@@ -267,13 +267,26 @@ install(
 # Standalone interface builds for wheels will need the files to be copied to a different location anyway,
 # so not copying the files for now isn't a big loss
 if(WIN32)
+    # Realistically for bundling into a Python wheel, the dll should be given a unique name
+    # to avoid conflicts with other packages using a different version of the dll with the same name
+    add_custom_command(
+        TARGET ${HELICS_PYTHON_TARGET_NAME} POST_BUILD
+        COMMAND
+            ${CMAKE_COMMAND} -E copy_if_different # which executes "cmake - E
+                                                  # copy_if_different..."
+            "$<TARGET_FILE:HELICS::helicsSharedLib>" # <--this is in- file
+            "$<TARGET_FILE_DIR:${HELICS_PYTHON_TARGET_NAME}>/" # <--this is out- file path
+    )
     if(COMMAND copy_key_files_to_target_location)
         copy_key_files_to_target_location(${HELICS_PYTHON_TARGET_NAME})
     endif()
-    if(COMMAND copy_shared_target)
-        copy_shared_target(${HELICS_PYTHON_TARGET_NAME} HELICS::helicsSharedLib)
-    endif()
     if(COMMAND install_key_files_with_comp)
         install_key_files_with_comp(python)
+    else
+        install(
+            FILES $<TARGET_FILE:HELICS::helicsSharedLib>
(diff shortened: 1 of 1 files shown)


nightlark:
I think the RTD build has gotten into some bad state with reporting statuses - maybe related to the PR starting out as a draft.

interfaces/python/CMakeLists.txt
```diff
@@ -267,13 +267,26 @@ install(
 # Standalone interface builds for wheels will need the files to be copied to a different location anyway,
 # so not copying the files for now isn't a big loss
 if(WIN32)
+    # Realistically for bundling into a Python wheel, the dll should be given a unique name
+    # to avoid conflicts with other packages using a different version of the dll with the same name
```
phlptp: So this isn't changing the name yet is it?  looks like just copying the file in a more robust way

interfaces/python/CMakeLists.txt
```diff
@@ -267,13 +267,26 @@ install(
...
+    add_custom_command(
+        TARGET ${HELICS_PYTHON_TARGET_NAME} POST_BUILD
+        COMMAND
+            ${CMAKE_COMMAND} -E copy_if_different # which executes "cmake - E
+                                                  # copy_if_different..."
+            "$<TARGET_FILE:HELICS::helicsSharedLib>" # <--this is in- file
+            "$<TARGET_FILE_DIR:${HELICS_PYTHON_TARGET_NAME}>/" # <--this is out- file path
```
phlptp: So on Windows there is a distinction between the target linking file. (*.lib) and the target file (*.dll)  do you need both?  And if we are building afterwards on windows do we need the .lib file in a different location than it is already?

interfaces/python/CMakeLists.txt
nightlark: No, it isn't changing the name. A name conflict is more likely to be a problem with the libzmq library than helicsSharedLib (more Python projects could be using something with a copy of ZeroMQ).

I think the name mangling should probably be handled by a step when generating the wheels that goes through and bundles any needed shared libraries on the system PATH into the generated wheel and gives them unique names.

interfaces/python/CMakeLists.txt
nightlark: Only the runtime needs to be in the wheel so the compiled Python module (a dll with a pyd extension) can find it. If someone is doing C/C++ development that needs the lib file for linking they should download the helics-shared archive for the release instead of using pip install.

The lib file is one of the reasons changing the library names gets annoying.

nightlark: The way building Python wheels works right now, I have GH actions download a copy of the compiled Helics-shared release archive and pull a copy of the HELICS source code into a subfolder. Then I have a setup.py script that uses CMake to do a standalone Python build for creating the precompiled packages, with the idea that we can also publish a source package that will let users with weird OS or Python version requirements build from source with pip install.

phlptp:
Do you want to do something to get the branch to build the docs or should I just merge it?

nightlark:
The docs have been built a few times for this PR, the latest one is https://readthedocs.org/projects/helics/builds/10273955/

But I guess opening another PR for this branch into another target messed up its status reporting.

PR #1035, opened 2020-01-17
Add STATUS mode to all cmake message() calls

nightlark:
### Summary
If merged this pull request will ensure all cmake message() calls have a mode set to STATUS instead of no mode. ucm and cmake_clone_git modules are updated to the latest versions.

### Proposed changes
- Update cmake_clone_git
- Update ucm
- Add STATUS mode to a message() call in the python interface CMakeLists.txt

### DIFF
--- a/config/cmake/ucm.cmake
+++ b/config/cmake/ucm.cmake
@@ -240,6 +240,37 @@ macro(ucm_set_runtime)
     endif()
 endmacro()
 
+# ucm_set_embedded_debug on MSVC
+# Sets the runtime (static/dynamic) for msvc
+macro(ucm_set_embedded_debug)
+    cmake_parse_arguments(ARG "EMBEDDED;EXTERNAL" "" "" ${ARGN})
+
+    if(ARG_UNPARSED_ARGUMENTS)
+        message(FATAL_ERROR "unrecognized arguments: ${ARG_UNPARSED_ARGUMENTS}")
+    endif()
+     
+	 if (MSVC)
+		ucm_gather_flags(0 flags_configs)
+    
+		# add/replace the flags
+		# note that if the user has messed with the flags directly this function might fail
+		# - for example if with MSVC and the user has removed the flags - here we just switch/replace them
+		if("${ARG_EMBEDDED}")
+			foreach(flags ${flags_configs})
+					if(${flags} MATCHES "/Zi")
+						string(REGEX REPLACE "/Zi" "/Z7" ${flags} "${${flags}}")
+					endif()
+			endforeach()
+		elseif("${ARG_EXTERNAL}")
+			foreach(flags ${flags_configs})
+					if(${flags} MATCHES "/Z7")
+						string(REGEX REPLACE "/Z7" "/Zi" ${flags} "${${flags}}")
+					endif()
+			endforeach()
+		endif()
+	endif()
+endmacro()
+
 # ucm_print_flags
 # Prints all compiler flags for all configurations
 macro(ucm_print_flags)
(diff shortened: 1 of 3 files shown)


config/cmake/ucm.cmake
```diff
@@ -32,18 +15,18 @@ cmake_minimum_required(VERSION 2.8.12)
...
-#
-#if(COMMAND cotire AND "1.7.9" VERSION_LESS "${COTIRE_CMAKE_MODULE_VERSION}")
-#    set(ucm_with_cotire 1)
-#else()
-#    set(ucm_with_cotire 0)
-#endif()
-#
-#option(UCM_UNITY_BUILD          "Enable unity build for targets registered with the ucm_add_target() macro"                     OFF)
-#option(UCM_NO_COTIRE_FOLDER     "Do not use a cotire folder in the solution explorer for all unity and cotire related targets"  ON)
+if(NOT COMMAND cotire)
```
phlptp: I think I had commented most of this out since it wasn't in use and was cluttering up the build somewhat

phlptp: we probably want it commented out for the time being since it would be confusing

nightlark: What do you mean by cluttering up the build?

phlptp: we are not using Cotire,  we don't have any plans to use cotire, so searching for that and any options there are pointless,    the UCM_UNIT_BUILD and UCM_NO_COTIRE_FOLDER are then options that are meaningless and should not show up in the cmake-gui or in the cache.  They are not meaningful for the project, so I don't think they should even show up.

nightlark: Having less custom changes for future updates would be nice.

How about something like:
```
if(COMMAND cotire AND "1.7.9" VERSION_LESS "${COTIRE_CMAKE_MODULE_VERSION}")
    set(ucm_with_cotire 1)
    option(UCM_UNITY_BUILD          "Enable unity build for targets registered with the ucm_add_target() macro"                     OFF)
    option(UCM_NO_COTIRE_FOLDER     "Do not use a cotire folder in the solution explorer for all unity and cotire related targets"  ON)
else()
    set(ucm_with_cotire 0)
endif()
```

I think that might be a change upstream would accept.

phlptp: For our purposes I don't really even want to search for COTIRE.  which happens on the earlier line.

nightlark: It checks a fixed location for the existence of a file and doesn't output anything if the file isn't found -- that should take a fraction of a second.

phlptp: ok

nightlark: Another option that might be possible to get upstream would be surrounding the entire cotire block with a check for "UCM_DISABLE_COTIRE" -- which could be set right before including ucm.

nightlark: Any preference? That could have some benefit in a scenario with subprojects using cotire, allowing individual projects to opt-in/out.

phlptp: That would be acceptable I think.  Have you ever used cotire?   Have you ever seen anyone use cotire?

nightlark: I've seen it included in some projects, though cmake 3.16 adds precompiled header support and an upcoming version might add unity build support.

nightlark: It looks like it is an opt-in feature -- `ucm_add_target` needs to get used instead of the built-in CMake `add_library` or `add_executable` functions -- so there probably isn't much value in having an option for (sub)projects completely disabling the check to see if it exists.

I changed it to only add the options if cotire is available, so it won't show up for building HELICS.

PR #1036, opened 2020-01-17
Set macOS rpaths to make installs movable

nightlark:
### Summary
If merged this pull request will set the rpaths on macOS so that installs can be moved around (useful for tar.gz release download for language interfaces). Some CMake rpath settings will also no longer be forced to their default values, or override a user that configures a different setting; this is aimed at making packaging for different systems more flexible.

### Proposed changes
- Add relative path to lib folder to macOS rpath
- Don't override user set CMake rpath settings or force the default values

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -578,8 +578,10 @@ if(NOT APPLE)
                 "\$ORIGIN/../${CMAKE_INSTALL_LIBDIR}"
     )
 else()
-    list(APPEND CMAKE_INSTALL_RPATH "@loader_path/.")
-    list(APPEND CMAKE_INSTALL_RPATH "@executable_path/.")
+    list(APPEND CMAKE_INSTALL_RPATH "@loader_path")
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -559,17 +559,15 @@ endif()
 # -------------------------------------------------------------
 # setting the RPATH
 # -------------------------------------------------------------
-# use, i.e. don't skip the full RPATH for the build tree
-set(CMAKE_SKIP_BUILD_RPATH FALSE)
-
-# when building, don't use the install RPATH already (but later on when installing)
-set(CMAKE_BUILD_WITH_INSTALL_RPATH FALSE)
```
phlptp: Is not having this variable set going to cause any potential issues on LINUX or mac for that matter if you are just installing normally.

nightlark: It seems to work -- the options just affect the files in the build tree (so the binaries used for testing with ctest). The default values for both options is false, so the it should behave the same as before.

Issue #1018, opened 2020-01-11
remove unneeded files from the all source package


phlptp:
**Describe the bug**
There are extra files from the units library that get bundled up into the all source tar.gz file

**What is the expected behavior?**
 What is the motivation / use case for changing the behavior?
To reduce the size of the all source package a little these files can be removed before packaging
The docs,  ThirdParty, and Test folders of the Units library can be removed These account for 6.5 MB of the file(before compression)  probably about 1.2 MB after compression.  

The doc and test folder of fmtlib can also be removed most likely.

PR #1038, opened 2020-01-19
Slimmer all source release tarball

nightlark:
### Summary
If merged this pull request will remove some unneeded folders from 3rd party submodules. The overall size drops from 17.9MB to 15.7MB (compressed). Addresses #1018.

### Proposed changes
- Remove unneeded files from the all source release tarball

### DIFF
--- a/scripts/_git-all-archive.sh
+++ b/scripts/_git-all-archive.sh
@@ -115,7 +115,7 @@ rmdir_list=(
 )
 for i in "${rmdir_list[@]}"
 do
-   tar --delete -f "${OUTPUT_BASENAME}.tar" "$i"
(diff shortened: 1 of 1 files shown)


nightlark:
An archive with this change can be downloaded at https://github.com/nightlark/HELICS/releases/download/vSlimTar3/Helics-vSlimTar3-source.tar.gz

phlptp:
have you tried building from the slimmed tarball?
I was hoping for a little more size reduction, but 12% is decent.

nightlark:
I compiled with the tarball on Quartz; benchmarks, apps, examples, tests, and shared libraries.

PR #1040, opened 2020-01-19
Combine the Python 2 and 3 interface build systems

nightlark:
### Summary
If merged this pull request will combine the Python 2 and 3 build systems (type maps and main swig i files remain separate). The same restrictions on needing swig for the Python 2 interface still apply, but it can now also be built standalone.

### Proposed changes
- Combine the Python 2 and 3 build systems, adding an option to build the Python 2 interface in a standalone build


nightlark:
The swig *.i files are still separate because the Python 3 one takes advantage of some functions only available in Python 3 (specifically `PyUnicode_*` functions added in 3.3).

interfaces/python/CMakeLists.txt
```diff
@@ -157,16 +179,23 @@ if(HELICS_ENABLE_SWIG AND SWIG_EXECUTABLE)
...
-    set(CMAKE_SWIG_FLAGS "-py3")
+    if(_py_ver EQUAL 3)
+        set(CMAKE_SWIG_FLAGS "-py3")
+    endif()
 
     if(SWIG_VERSION VERSION_GREATER "4.0.0")
         set(CMAKE_SWIG_FLAGS ${CMAKE_SWIG_FLAGS} "-doxygen")
     endif()
 
+    set(_helicsPython.i_FILE "helicsPython.i")
```
phlptp: I suppose we could rename the .i file to helicsPython3.i and then you could use the _py_ver variable to find it

nightlark: If we're okay with requiring SWIG 3.0.9+ (released in May 2016), we can use the same helicsPython.i and type_maps.i file for both Python 2 and 3 -- there's just one section that would need to use the `SWIGPYTHON_PY3` define.

phlptp: Not sure we can make that a requirement yet. Definitely can for HELICS 3.0 though.

phlptp: Not sure we can make that a requirement yet. Definitely can for HELICS 3.0 though.

phlptp: Not sure we can make that a requirement yet. Definitely can for HELICS 3.0 though.

phlptp: The other option would be to include the codegen for python2 in the repo.  Then we could make the requirement.  But we need to be able to build the python 2 on older systems yet.

nightlark: The SWIG 3.0.9+ requirement would be for building the Python 3 interface with SWIG; any SWIG 3 version would work for building the Python 2 interface since the fallback code if `SWIGPYTHON_PY3` is undefined (always for pre-3.0.9) would be the Python 2 compatible type map.

nightlark: I'm also planning on updating the packaging code to build Python 2 binary distributions for Windows/Linux/macOS that can be installed with pip.

phlptp: I think @kdheepak  should weigh in, but I don't like the idea of adding an additional requirement right now for the 2.4 release.   From the sound of it we can make it pretty easy to workaround so I want to take a closer look after the release, and I do like the idea of having a combined python2 and 3 section of the build.  So we can evaluated it closely for a 2.5 release an definitely the 3.0 release.  But I really hesitate to change it for the 2.4 release.

phlptp:
@kdheepak  can you take a look at this?

kdheepak:
This looks good! Always nice to delete files :)

kdheepak:
Before we merge to master, I'd like to make sure all the conda builds work as expected.

PR #1041, opened 2020-01-19
Add a Linux installer to the release builds

nightlark:
### Summary
If merged this pull request will add a Linux installer release archive (manylinux2010 compliant, CentOS 6+). The main difference from the shared library archive is that it includes HELICS app binaries that have been static linked to the system stdlibc++ and libgcc for portability. It does not include pre-built Python/Java interfaces (likely requires compiling Java from source and installing a copy of Python; pip install is easier than picking a single Python version to include).

### Proposed changes
- Add a Linux installer with the HELICS apps and shared library to the release builds


nightlark:
It's mostly the same as the windows zip and macos tar archives -- user extracts it to wherever they want on their system. Should probably also update the new helics website with some download links and basic instructions.

PR #1042, opened 2020-01-20
Set gitattributes hunk headers for common source file types

nightlark:
### Summary
If merged this pull request will set the hunk headers to use for several common source files types in the HELICS repository. These are the lines of text that show up when doing a `git diff` or using the `-p` (`--show-function`) or `-W` (`--function-context`) arguments to `git add`, `git grep`, and `git diff`.

Further details:
https://git-scm.com/docs/gitattributes#_defining_a_custom_hunk_header
https://github.blog/2019-08-16-highlights-from-git-2-23/
https://github.blog/2020-01-13-highlights-from-git-2-25/

### Proposed changes
- Set git hunk header built-in patterns to use for common source file types

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -3,3 +3,12 @@
 *.jpg -text
 *.png -text
 *.mat -text
+
+*.cpp  diff=cpp
+*.hpp  diff=cpp
+*.c    diff=cpp
+*.h    diff=cpp
+*.py   diff=python
+*.java diff=java
+*.m    diff=matlab
+*.cs   diff=csharp
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,8 +1,8 @@
 * text=input
 
-*.jpg -text
-*.png -text
-*.mat -text
+*.jpg binary
+*.png binary
+*.mat binary
 
 *.cpp  diff=cpp
 *.hpp  diff=cpp


nightlark:
Adding defaults to gitattributes didn't seem to break anything on older versions of git (or setting the value to an invalid random string).

phlptp:
What do these changes do that is beneficial?


---
Sent from Workspace ONE Boxer<https://whatisworkspaceone.com/boxer>

On January 20, 2020 at 3:14:18 PM PST, Ryan Mast <notifications@github.com> wrote:

Adding defaults to gitattributes didn't seem to break anything on older versions of git (or setting the value to an invalid random string).

nightlark:
When you do a `git diff` you get a bunch of output for the changes like `@@ -k,l +n,m @@ TEXT` -- where `TEXT` is something that gives some context for where the change is. The default selection matches `GNU diff -p` which is according to the docs "a line that begins with an alphabet, an underscore or a dollar sign" and not aware of the language context for a file.

This affects things like `git diff` output or `git add -p` for making separate commits for different changes to a file.

I've noticed the main difference is with the non-C++ files like Python/Java/Matlab where before it just kind of picks a random nearby line, and after it uses a function or class name that the change was in.

Before (change isn't even in the if or catch blocks):
```
--- a/tests/matlab/messageFilterTests.m
+++ b/tests/matlab/messageFilterTests.m
@@ -23,7 +23,7 @@ if (~helicsBrokerIsValid(fedStruct.broker))
     success=false;
     return;
 end
-fedInfo=helicsCreateFederateInfo();
+   fedInfo=helicsCreateFederateInfo();
 if (fedInfo==0)
     success=false;
     return;
@@ -40,7 +40,7 @@ catch ec
     return
 end
 for ii=1:count
-fedStruct.mFed{ii}=helicsCreateMessageFederate(['fed',num2str(ii)],fedInfo);
+   fedStruct.mFed{ii}=helicsCreateMessageFederate(['fed',num2str(ii)],fedInfo);
 if (fedStruct.mFed{ii}==0)
     success=false;
 end
```

After (pulls out the function the change is in):
```
--- a/tests/matlab/messageFilterTests.m
+++ b/tests/matlab/messageFilterTests.m
@@ -23,7 +23,7 @@ function [fedStruct,success]=generateFeds(count)
     success=false;
     return;
 end
-fedInfo=helicsCreateFederateInfo();
+   fedInfo=helicsCreateFederateInfo();
 if (fedInfo==0)
     success=false;
     return;
@@ -40,7 +40,7 @@ function [fedStruct,success]=generateFeds(count)
     return
 end
 for ii=1:count
-fedStruct.mFed{ii}=helicsCreateMessageFederate(['fed',num2str(ii)],fedInfo);
+   fedStruct.mFed{ii}=helicsCreateMessageFederate(['fed',num2str(ii)],fedInfo);
 if (fedStruct.mFed{ii}==0)
     success=false;
 end

```

.gitattributes
```diff
@@ -3,3 +3,12 @@
 *.jpg -text
```
phlptp: Does it matter that these are listed as text, even though they really binary files.

nightlark: `-text` is unsetting the text attribute on those files to make sure "line endings" don't get normalized.

nightlark: Maybe it would be better to set that using the `binary` macro, which might be clearer than unsetting text.

phlptp: Ah, I see now.  having some mention of binary would make it clearer if possible.

PR #1044, opened 2020-01-20
Fix issue where git --local flag would cause fatal error

### LINKED ISSUE #996: fatal: --local can only be used inside a git repository

eberleim:
When building in an external directory, the git repository would
be detected, but git would be unable to run in-folder. This fix
simply ensures that git is run from within the project directory
and not the build directory.

A new branch has been created, to avoid the "error" trigger word from the previous version of this issue. 

Resolves #996

### DIFF
--- a/config/cmake/addGitHooks.cmake
+++ b/config/cmake/addGitHooks.cmake
@@ -4,6 +4,7 @@ set(CLANG_FORMAT_MIN_VERSION "7.0")
 function(set_git_hooks_enabled)
     execute_process(
             COMMAND ${GIT_EXECUTABLE} config --local core.hooksPath .githooks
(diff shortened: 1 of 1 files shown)


eberleim:
This technically sets a Git config flag, rather than any environmental flags. It affects the whole repository, but shouldn't affect other build platforms, though platforms which do not know about the bash terminal may be unable to execute the hook itself (Git seems to take care of this internally for most platforms).

phlptp:
would it make sense to have the GIT Hook stuff default to off if the build path is outside the source path?

Issue #663, opened 2019-05-03
man pages


nightlark:
Write/generate man pages for the helics_* executables and install them on Linux/Mac(/BSD).

It seems like man pages are typically written using GNU troff for typesetting, but there are converters for Markdown and other alternatives. It would be nice to be able to generate HTML pages from the same source as the man pages for including in RTD.

nightlark:
Some discussion in PR #662 on man pages being easier to read for detailed information on program usage than long help messages. Caveat is that not all platforms (Windows) support man pages.

PR #1052, opened 2020-01-24
Add man pages for HELICS executables

nightlark:
### Summary
If merged this pull request will add a set of man pages for the helics_* executables that get installed to the share/man(/man1) folder of the install prefix. WSL is able to launch Windows apps and supports viewing man pages, and people have made other manpage viewing tools for Windows, so the manpages are included in installs for all platforms.

Re: #663.

### Proposed changes
- Add man pages describing the options and purpose of the various helics_* executables and their subcommands


docs/man/manpage_out/helics_app-broker.1
```diff
@@ -0,0 +1,268 @@
+'\" t
```
phlptp: why is this one in troff format?  Is that just a trial?

phlptp:
There seem to be generated files and the asciidoc files.  Are both necessary or is on generated from the other?

docs/man/manpage_out/helics_app-broker.1
nightlark: The manpage_out subfolder has a copy of the generated troff files so that there's something that can be installed using `make install`  or for packaging without a dependency on installing a2x.

nightlark:
The *.1 files are generated from the asciidoc files.

phlptp:
Are you thinking it is something that could be added to the automatic generators, like the interface files?

nightlark:
That's probably the most practical way to handle the man pages (fortunately the output is still text files, so diffs for updates should be relatively small).

nightlark:
I was thinking about our discussion on CLI11 generating manpages -- something that would be really useful is outputting the options and a synopsis into a format that supports includes (asciidoc, rst). Then the free form parts of the page could be written up (multi-paragraph descriptions, examples, see also, etc), and the options/synopsis could have their relevant include files updated by CLI11.

This could be more flexible and easier to get a good man page than having CLI11 try to generate the troff (no include directive?) directly.

phlptp:
That will likely be a couple months before the CLI11 stuff would be ready, but I think it is doable.  I might start an issue on the CLI11 github and see what other people think.

nightlark:
I'm going to change the extension from .txt to .adoc before merging -- GitHub has some amount of AsciiDoc support, though I've heard their include directive might not quite be working.

PR #1058, opened 2020-01-29
Change macOS SDK target from 10.10 to 10.9

nightlark:
### Summary
If merged this pull request will change the macOS SDK built against from 10.10 to 10.9 for the macOS release builds. This seems to be the lowest common SDK targeted by binary packages on PyPI (and the lowest version any of the python.org macOS installers target, at least for Python versions that aren't EOL).

### Proposed changes
- Target the macOS 10.9 SDK for HELICS release builds


nightlark:
Checked the compiled artifacts in my fork to confirm that they have 10.9 set as the minimum sdk version.

phlptp:
Is the reason for this change just to get a slightly older version working?

nightlark:
Yea. Also changed the script slightly to make switching the sdk version easier.

PR #1067, opened 2020-02-04
Develop

phlptp:
### Summary

If merged this pull request will merge develop to master for the HELICS 2.4 release


phlptp:
@kdheepak I am expecting a couple more commits into develop before this can be merged.   But we can review most of the changes already.

phlptp:
@kdheepak  @nightlark take another look through the changes.  There are a lot since this is where we did the formatting changes,  but if you see anything major lets fix it otherwise it will be good to get this release out.

nightlark:
Found two 2.3.1 version references, I think both are unlikely to be found by a user.

Otherwise I think the changes LGTM.

kdheepak:
We should update the developer docs to say that `clang-format` and the corresponding version is now required  because of the `pre-commit` step.

nightlark:
It's not required, if `clang-format` isn't present the githook bypasses the format step (https://github.com/GMLC-TDC/HELICS/blob/develop/.githooks/pre-commit#L3), and the git hook is only activated if the available version of `clang-format` meets the minimum required (https://github.com/GMLC-TDC/HELICS/blob/develop/config/cmake/addGitHooks.cmake#L39-L44).

kdheepak:
Okay, great!

phlptp:
once the builds finish I will merge, then once the master builds finish I will make the release

nightlark:
Testing - lgtm

nightlark:
Testing LGTM

PR #1093, opened 2020-02-23
Cmake new policy

phlptp:
### Summary

If merged this pull request will remove the cmake_policy old statements to support new cmake editions more easily.  

### Proposed changes

- specify the cmake policy to the new versions as they become available.  
- use the swig target variable for specifying the later target modifications
- update the copyright date on a few components
- change the name of the Java target since the new policy affect the module name and resulting in some odd names.


phlptp:
@nightlark  is there anyway we can run the interface file gen on this branch and run the tests before merging?

nightlark:
Maybe you could try making a copy of this branch, and in the GitHub workflow for interface generation add the branch name to see if it runs on the copy of the branch -- then revert the workflow file change or cherry-pick the updated interface file commit?

phlptp:
interesting, I guess it didn't change the files at all yet.  At some point we should make the ones generated consistent with the swig generated ones but for now this is probably ok yet.  

I created a new branch and the files showed up as exactly the same.

nightlark:
It looks like the swig generated ones that we keep in the repository match whatever the workflow generated when swig ran.

interfaces/python/CMakeLists.txt
```diff
@@ -15,11 +15,6 @@ get_directory_property(hasParent PARENT_DIRECTORY)
 if(NOT hasParent)
     message(STATUS "Standalone Python interface build")
 
-    # find_package(<PackageName>) searches prefixes given in <PackageName>_ROOT CMake variable
-    if(POLICY CMP0074)
-        cmake_policy(SET CMP0074 NEW)
```
nightlark: This should be replaced by:
```
cmake_policy(VERSION ${CMAKE_MAJOR_VERSION}.${CMAKE_MINOR_VERSION})
```

For standalone builds, the policy set in the parent will not apply.

phlptp:
I had thought one of the java ones would be different but might have been wrong.  It is also possible the generators between swig and the interface gen are now slightly different as well.  

One thing to check is that I think the names of the generated files in the python generation are slightly different.  I don't know if that will affect things in the pypi generation or not.

nightlark:
I'll try running the generation locally.. it's possible that if the file names have changed (e.g. added files) then maybe the change detection is missing those files. Based on that, I'm pretty sure it has no way to handle removing files that swig is no longer generating.

Maybe an upcoming change that will need to get made is ensuring swig generated files are all put in a subfolder that can be cleared out for the swig generate interface files action.

phlptp:
I know during some intermediate stages of development some of the files changed names, but I resolved some of them,   just wanted to make sure I got them all with the updated cmake policy.   Thanks for checking.

PR #1100, opened 2020-02-26
Fix swig interface file generation handling of added/removed files

### LINKED ISSUE #1095: Automated SWIG interface generation needs to handle new/removed files
**Describe the bug**
A recent PR resulted in some generated swig interface file names changing -- the automated SWIG PR setup doesn't appear to detect those changes.

The automated SWIG generation workflow should handle new/deleted/renamed files.

nightlark:
### Summary
If merged this pull request will make move the generated files for the matlab and python interfaces to an `interface` subfolder, to match the java interface and make so the automated swig interface generating action can handle added/removed/renamed files. Resolves #1095.

https://github.com/GMLC-TDC/HELICS/pull/1099 is a test run of the action to regenerate the matlab interface files. For easier reviewing, I'll remove the matlab files and get the moved/regenerated ones when the rest of the PR is ready.

### Proposed changes
- Move the generated matlab and python swig interface files to `interface` subfolders. Update the CMake scripts to copy files from the right place for non-swig builds, and overwrite files in the subfolder instead of the root interfaces/<language> folder.
- Change how the swig interface gen action adds changed files, to handle cases where swig generates a new file, removes a file, or renames a file.


nightlark:
I left the extra directory handling the same. The overall installed/build directory structures should still be the same, and only the generated files in the source directory should be moving around.

I added in another glob to capture the (non-generated) *.m files that won't be moved to the subfolder.

nightlark:
I'm going to move the generated matlab files to their new location and run the matlab tests before merging.

phlptp:
Do we have any way to test the matlab build?

phlptp:
If we don't have another way I will try it my local machine tonight.

nightlark:
I’m going to try running it on LC.

nightlark:
The latest commit copies matlab files to the HELICS build directory for running tests, and the one right before it moves the swig generated matlab files to the interface subfolder (the number of files moved matches the number that the automated test PR generated).

@phlptp I think you should check on your system to make sure it works, on Linux there are a number of issues getting it running that have likely been around for some time.

phlptp:
I was able to build successfully the matlab interface on matlab 2013a and 2017b on Windows.

PR #1107, opened 2020-03-02
Renormalize line endings for md, json, player, and bat files

nightlark:
### Summary
If merged this pull request will renormalize the line endings for json and player files to lf, and bat files to crlf (via .gitattributes eol settings). A *.bat rule is added to the editorconfig file to set it's line endings to crlf since it's a Windows batch script.

### Proposed changes
- Renormalize line endings for existing json, player, and bat files
- Set .gitattributes rules for md, player, json, and bat file line endings
- Set .editorconfig rule for bat file line endings (other file types care covered by another rule)


.gitattributes
```diff
@@ -1,7 +1,9 @@
 * text=input
 
-*.md eol=lf
-*.json eol=lf
+*.bat text eol=crlf
+*.md text eol=lf
+*.json text eol=lf
+*.player text eol=lf
```
phlptp: Some of these player files are from gridlabd, do we know if gridlab can handle either line ending?  I presume it could but might be something to at least consider

nightlark: Given that GridLab-D support for Windows is using MinGW, it's more likely to have trouble with Windows style (crlf) line endings or [old-macintosh (cr) line endings](https://github.com/gridlab-d/gridlab-d/issues/846).

It [looks like they expect \n but not \r for blank lines](https://github.com/gridlab-d/gridlab-d/blob/f56223aa6e10dd130668ec9f8092da96f5d3f895/tape/player.c#L318-L321).

nightlark: All of the [recorder](https://github.com/gridlab-d/gridlab-d/blob/f56223aa6e10dd130668ec9f8092da96f5d3f895/tape/recorder.c#L147-L170) uses `\n` line endings without `\r`.

If their player gets broken by files only having a `\n`, I think that's a pretty major bug.

Issue #898, opened 2019-11-12
Add github-action for submitting PR to helics-conda and HELICSBuilder


kdheepak:
There's only a couple of lines that need to be changed in [helics-conda](https://github.com/GMLC-TDC/helics-conda) and [HELICSBuilder](https://github.com/GMLC-TDC/HELICSBuilder) every time we make a release. It'd be great if we can make a PR to those repos when a release is made here.

nightlark:
Can you add me as a collaborator to this repository with the maintain permission setting? Hoping that will be enough to setup and update secrets for GitHub Actions in this repository.

PR #1112, opened 2020-03-03
Send helics-version-update event to helics-packaging after new releases

nightlark:
### Summary
If merged this pull request will send an event after all HELICS release binaries are built to helics-packaging telling it that a new version has been released.

The action that handles the event in `helics-packaging` updates the version file, but doesn't open a PR yet -- it shows a [git diff](https://github.com/GMLC-TDC/helics-packaging/runs/483558701?check_suite_focus=true#step:4:10) of the change that would be in a PR.

### Proposed changes
- Send a `helics-version-update` event to `helics-packaging` after the the release build is finished

### DIFF
--- a/.github/workflows/release-build.yml
+++ b/.github/workflows/release-build.yml
@@ -7,9 +7,9 @@ on:
     types: published
 
 jobs:
-################################
+#####################################
 # Create all submodules archive
-################################
+#####################################
   create-all-submodule-archive:
     name: Create all submodule archive
     runs-on: ubuntu-latest
@@ -33,9 +33,9 @@ jobs:
         name: all-submodules-archive
         path: artifact
         
-################################
+#####################################
 # Build MSVC archives
-################################
+#####################################
   build-windows-msvc:
     runs-on: ${{ matrix.os }}
     strategy:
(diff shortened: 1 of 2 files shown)


.github/workflows/release-build.yml
```diff
@@ -7,9 +7,9 @@ on:
     types: published
 
 jobs:
-################################
+#####################################
```
phlptp: Is there a reason all these are longer?

nightlark: Just to match the length of the longest header name

.github/workflows/release-build.yml
```diff
@@ -291,3 +291,21 @@ jobs:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         UPLOAD_URL: ${{ github.event.release.upload_url }}
       run: ./.github/workflows/upload-release-asset.sh "artifacts/Helics-$(git rev-parse --abbrev-ref "${GITHUB_REF}")-SHA-256.txt"
+
```
phlptp: Is there a corresponding receiver in the other repos?

nightlark: The git diff link in the PR description goes to one of them.

phlptp:
This is related to #898 ?

nightlark:
Indirectly - helics-packaging will be set up to send version update events to any other repositories that need updates.

PR #1123, opened 2020-03-06
Add 'helics-release-build' dispatch event trigger

nightlark:
### Summary
If merged this pull request will add an alternate way to trigger building release artifacts, either all of the assets or a subset based on asset "type", and with an option to set which branch/tag they should be built for.

It can't limit by os/arch yet because the matrix information isn't available for use in conditionals at the job level.

### Proposed changes
- Trigger release builds for 'helics-release-build' events -- recognizes `everything`, `archive`, `msvc`, `installer*`, and `sharedlib*` package names.

### DIFF
--- a/.github/workflows/release-build.yml
+++ b/.github/workflows/release-build.yml
@@ -1,6 +1,11 @@
 name: Build release artifacts
 
 on:
+  # Handles a 'helics-release-build' dispatch event type
+  # The event payload should contain these fields:
+  # packages - a list of which packages to build
+  # commit (optional) - the commit-ish to do a release build with
+  repository_dispatch:
   schedule:
     - cron: '15 09 * * *' # Run at in the early hours of the morning (UTC)
   release:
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -38,6 +38,7 @@ Increased code coverage and additional bug fixes.  The error propagation in HELI
 -  `localError`, and `GlobalError` function calls the Federate API and in the C++ and sharedLibrary.  
 -  `helics_terminate_on_error` flag to escalate what would be a local error into a global one that will halt the co-simulation.  This flag can be specified through the flag to federates or to brokers and cores through a command line option `--terminate_on_error` 
 -   A 32-bit Windows zip install archive for releases
+-   Support for a 'helics-release-build' event trigger to the release build GitHub Actions workflow
```
phlptp: I am guessing whichever PR is merged second is going to have a conflict on this part of the Changelog

.github/workflows/release-build.yml
```diff
@@ -125,7 +144,9 @@ jobs:
     steps:
     - name: Checkout event ref
       uses: actions/checkout@v2
-      if: github.event.action == 'published'
+      with:
+        ref: ${{ github.event.client_payload.commit }}
+      if: github.event.action == 'published' || github.event.action == 'helics-release-build'
```
phlptp: This one doesn't have conditions like some of the others,  just want to make sure that is intentional?

CHANGELOG.md
nightlark: Yep... maybe it could get moved to the changed section and reworded.

.github/workflows/release-build.yml
nightlark: It's above the matrix section -- https://github.com/GMLC-TDC/HELICS/pull/1123/files/fd29bb2185f602ccb6567c6d084aaa80002486dc#diff-01667485dced5bf57d8837edbffc6507R117

nightlark: https://github.com/GMLC-TDC/HELICS/blob/fd29bb2185f602ccb6567c6d084aaa80002486dc/.github/workflows/release-build.yml#L117

nightlark: The checkout steps were changed so that it would use this one for published/event builds, and the fixed 'develop' branch checkout for the scheduled nightly build.

PR #1125, opened 2020-03-06
Update generated HELICS apps manpages

nightlark:
### Summary
If merged this pull request will update the generated manpages for the HELICS apps

### Proposed changes
- Update generated HELICS apps manpages with the new options


nightlark:
The dates always changing is a bit annoying for detecting changes. Maybe when a PR to master is opened?

PR #1126, opened 2020-03-06
Final Version and doc update to roadmap and readme and date.

phlptp:
### Summary

If merged this pull request will update the date roadmap, and add a new section to the readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -150,8 +150,6 @@ A list of optional component that are not included in HELICS but are optionally 
 
 ## Publications
 
-### General HELICS
-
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -154,6 +154,11 @@ A list of optional component that are not included in HELICS but are optionally
```
nightlark: Should we get rid of the `### General HELICS` heading and just list the papers? Doesn't seem like it makes sense to categorize them unless there are a lot of papers.

phlptp: I made the change to get rid of that header and your other suggestion

PR #1129, opened 2020-03-07
Develop

phlptp:
### Summary

If merged this pull request will update master to v2.4.1 for release


phlptp:
The Drone CI build and the docker hub are failing due to a change in the automatic line endings, which is making the merge with the PR error when they are loading it it up.    The MSYS build is failing due to a server failure on the msys2 repositories.  The codecov errors we are still trying to figure out the exact setup we want for the actual checks, but the code coverage is still improving, so that one we will figure out eventually.    In any case there isn't much we can do to make these tests pass until the PR is merged which should fix the drone and docker checks.

PR #1130, opened 2020-03-07
Fix macOS Python interface build

nightlark:
### Summary
If merged this pull request will fix a missed `_helics` target rename in the Python interface CMake file that breaks building the Python interface on macOS with swig. A Python 2 interface is unbuildable on macOS with the 2.4.1 release.

### Proposed changes
- Switch the missed instance of `_helics` to `${SWIG_MODULE_helics_REAL_NAME}`


kdheepak:
Should we make a new 2.4.2?

nightlark:
We had some discussion on gitter about it -- since it only really affects macOS users that need to use the Python 2.7 interface (generated interface files should work fine for almost all 3.x users), in ~2 weeks we can do a 2.4.2 release with a few other things.

kdheepak:
Sounds good to me.

PR #1169, opened 2020-04-01
Add spack install to list of package managers

nightlark:
### Summary
If merged this pull request will add spack to the list of package managers for installing HELICS.

### Proposed changes
- Add spack to the various docs listing package managers that can be used to install HELICS
- Add pip to the install options in the README

### DIFF
--- a/docs/installation/index.md
+++ b/docs/installation/index.md
@@ -37,6 +37,8 @@ Use Spack (requires Spack develop branch or versions released after v0.14.1) on 
 spack install helics
 ```
 
+For more information on supported options (e.g. using a custom HELICS build with MPI support), see the [package managers](https://helics.readthedocs.io/en/latest/installation/package_manager.html) page for more details, or the documentation for your package manager.
+
 ### Using an installer for your operating system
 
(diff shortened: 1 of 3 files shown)


docs/installation/index.md
```diff
@@ -29,6 +29,14 @@ pip install helics
 pip install helics-apps
 ```
 
+OR
+
+Use Spack (requires Spack develop branch or versions released after v0.14.1) on HPC systems (Linux/macOS) to install C/C++ HELICS components and apps:
+
+```bash
+spack install helics
+```
```
phlptp, line 38: I believe in the spack configuration you had some variants that would work as well,  You might want to explicitly state for example how to install with spack including MPI or something to that effect.

phlptp: or at least point to where you get more details in the next file

nightlark: The read the docs page on package managers talks about it. Maybe the top of quick start should say any of these package managers may support other options, see that page for details and the manuals for your respective package managers?

PR #1172, opened 2020-04-02
Use sourceline key for Travis CI apt sources

nightlark:
### Summary
If merged this pull request will specify Travis APT sources using the sourceline key in the Travis config file. Hopefully it resolves the disallowed sources error that Travis is giving.

### Proposed changes
- Use sourceline key in Travis yaml file for setting apt sources

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -111,7 +111,7 @@ jobs:
         apt:
           sources:
             - sourceline: "ppa:ubuntu-toolchain-r/test"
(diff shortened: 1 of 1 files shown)


phlptp:
Are the changes from this useful any longer?  or was it a bug in travis that got fixed or something of that nature.

nightlark:
I haven't been able to find news from Travis about the issue. There was a post on their forums so at least a few other people were affected, but no official response saying it was a bug that's been fixed.

PR #1176, opened 2020-04-06
Set BOOST_ROOT to Boost 1.72.0 on Windows Azure CI

nightlark:
### Summary
If merged this pull request will fix the failed Boost find on Windows Azure Pipelines builds. It seems Azure Pipelines and GitHub Actions are using the same VM images, and Boost was removed from the PATH environment variable (or BOOST_ROOT is just no longer set).

### Proposed changes
- Set BOOST_ROOT to the Boost 1.72.0 install on Azure pipelines Windows images
- Disables Java interface builds on Windows

### DIFF
--- a/.ci/azure-pipelines.yml
+++ b/.ci/azure-pipelines.yml
@@ -51,11 +51,11 @@ jobs:
 #    displayName: 'Configure Python interface'
 #    condition: and(ne( variables['imageName'], 'vs2015-win2012r2' ), eq( variables['langArch'], 'x64' ))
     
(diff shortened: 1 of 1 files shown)


nightlark:
Re-enabling the Java tests will be part of cleaning up the Java interface swig/CMake scripts.

phlptp:
Do these changes have anything to do with eachother or just convenient to put them together.

nightlark:
Pretty sure the Java failure is unrelated, so just convenient combining them.

PR #1183, opened 2020-04-09
Cpplint fixes

phlptp:
### Summary

If merged this pull request will be the first in a series of fixes for cpplint

### Proposed changes

- fix todo format
- fix runtime/int
- fix license
- fix header_guard issues


src/helics/cpp98/CPPLINT.cfg
```diff
@@ -0,0 +1,24 @@
+set noparent
```
nightlark, line 1: If this line gets removed it might be possible to just use `filter=-build/header_guard # this doesn't match the HELICS style for the C shared library` in this file.

phlptp: that makes things a little simpler

PR #1186, opened 2020-04-10
Run clang-tidy on PRs

nightlark:
### Summary
If merged this pull request will add clang-tidy to the static analyzers run on PRs.

This PR check run shows the warnings output from a PR that modifies 3 files:
https://github.com/GMLC-TDC/HELICS/pull/1187/checks?check_run_id=577183846

Running clang-tidy on all source files takes a long time, probably around an hour.

### Proposed changes
- Run clang-tidy on changed files in a PR


phlptp:
Would it make sense to put in some checks that just skip the test if the number of files is large, like >20, and what is this going to do with things like the formatting PR's and PR for the interface files?

phlptp:
which clang-tidy file does this use.  The .clang-tidy in the root?

phlptp:
I think we will have to clean up the clang-tidy file somewhat to reduce ignored warnings

phlptp:
I would expect eventually we could have this error if any issues are found?

nightlark:
> Would it make sense to put in some checks that just skip the test if the number of files is large, like >20, and what is this going to do with things like the formatting PR's and PR for the interface files?

That's a good idea. It might not run at all for interface files, since the compile commands database generated doesn't have any of the interfaces enabled. I could probably make it skip running on PRs whose names start with "Automated" which would avoid running on the formatting PRs.

> which clang-tidy file does this use. The .clang-tidy in the root?

It uses the same file as clang-tidy typically uses (should match the clang-tidy targets in the CMake). Probably starting with the closest parent?
```
When the value is empty, clang-tidy will
attempt to find a file named .clang-tidy for
each source file in its parent directories.
```

> I think we will have to clean up the clang-tidy file somewhat to reduce ignored warnings

> I would expect eventually we could have this error if any issues are found?

Does clang-tidy just give warnings, or does it ever give an error? If clang-tidy ever exits with a non-zero exit code then this will error. If clang-tidy only gives warnings, after some of them are cleaned up we could make it error if the clang-tidy output has any lines containing "warning:".

phlptp:
It does generate some errors for certain critical things by default.  I will play with it more once this is merged and we start the (long) process of cleaning things up.

phlptp:
Why is the clang-tidy PR check failing?

phlptp:
> Why is the clang-tidy PR check failing?

looks like you are working on that.

nightlark:
It looks like grep not finding any changed source files (exit code 0 if grep gets no matches) was making the clang-tidy job fail.

The automated source formatting PRs aren't a problem -- because they're created by the @github-actions bot account, they don't trigger action workflow runs.

.github/workflows/static-analyzers.yml
```diff
@@ -13,3 +13,39 @@ jobs:
...
+    if: github.event_name == 'pull_request'
+    runs-on: ubuntu-latest
+    container: helics/buildenv:clang-extra-tools
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Run clang-tidy on changed files
+      shell: bash
+      run: |
+        FILES_URL="$(jq -r '.pull_request._links.self.href' "$GITHUB_EVENT_PATH")/files"
```
phlptp, line 27: would it be appropriate to put this in a separate script?  looks like it might be getting a bit long for having in a yml file or are there aspects of it that require being in the yml file

PR #1204, opened 2020-04-16
Cpplint IWYU

phlptp:
### Summary

If merged this pull request will resolve include_what_you_use checks for cpplint.

### DIFF
--- a/src/helics/common/zmqContextManager.cpp
+++ b/src/helics/common/zmqContextManager.cpp
@@ -16,6 +16,7 @@ SPDX-License-Identifier: BSD-3-Clause
  * LLNS Copyright End
  */
 
+#include <utility>
 #include "zmqContextManager.h"
 
 #include "cppzmq/zmq.hpp"
--- a/src/helics/common/zmqContextManager.cpp
+++ b/src/helics/common/zmqContextManager.cpp
@@ -19,7 +19,8 @@ SPDX-License-Identifier: BSD-3-Clause
 #include "zmqContextManager.h"
 
 #include "cppzmq/zmq.hpp"
-//#include <iostream>
+//#include <utility>
+#include <iostream>
 #include <map>
 #include <mutex>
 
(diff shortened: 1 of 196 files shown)


src/helics/common/zmqContextManager.cpp
```diff
@@ -19,7 +19,10 @@ SPDX-License-Identifier: BSD-3-Clause
 #include "zmqContextManager.h"
 
 #include "cppzmq/zmq.hpp"
-//#include <iostream>
+
+#include <utility>
+//#include <utility>
```
nightlark, line 24:
```suggestion
```

src/helics/network/zmq/ZmqComms.cpp
```diff
@@ -15,8 +15,13 @@ SPDX-License-Identifier: BSD-3-Clause
 #include "../networkDefaults.hpp"
 #include "ZmqCommsCommon.h"
 #include "ZmqRequestSets.h"
-//#include <csignal>
+
+#include <utility>
+//#include <utility>
```
nightlark, line 20:
```suggestion
```

src/helics/network/zmq/ZmqCommsSS.cpp
```diff
@@ -14,9 +14,12 @@ SPDX-License-Identifier: BSD-3-Clause
 #include "../networkDefaults.hpp"
 #include "ZmqCommsCommon.h"
 #include "ZmqRequestSets.h"
-//#include <csignal>
+//#include <map>
```
nightlark, line 17: Why is this commented out?

phlptp: my guess is at one point we were using map objects and it is just left over

nightlark: It looks like this file still uses std::map, I couldn’t tell if something got messed with where new headers got inserted, since csignal is now uncommented.

phlptp: that might be how I inserted it, since the code that did that was picking off the first `#include <` and injecting it.  which probably meant that I didn't take into account comment characters.

phlptp: I fixed in the latest commit

PR #1207, opened 2020-04-17
Network data config

phlptp:
### Summary

If merged this pull request will allow json configuration of broker/core parameter and network parameters including specifying a particular subsection of the file.  

### Proposed changes

- update CLI11 with some new features
- tweak the jsonConfig to match 
let the  networkBrokerData and brokerBase Command line use that file.


phlptp:
This PR is the second in a series to get the multi-broker operational.  The first task has been to figure out how to handle configuration of the multibroker in a sane fashion. That requires allowing the core/comms configuration to be handled through files, which required some modifications to CLI11 and the structure to allow it to work with JSON files and specific sections of JSON and TOML files.  Which is what the last two PR's have been about.  

Eventually the changes will work through to the CLI11 master, but for now we are operating on a custom fork again.

PR #1221, opened 2020-04-23
build configurations

phlptp:
### Summary

If merged this pull request will add a specialized build configuration for Raspberry pi and some documentation about it.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -437,6 +437,17 @@ option(ENABLE_ZMQ_CORE "Enable ZeroMQ networking library" ON)
 mark_as_advanced(ENABLE_MPI_CORE ENABLE_ZMQ_CORE)
 
 # -------------------------------------------------------------
+# Some specialized configurations
+# -------------------------------------------------------------
+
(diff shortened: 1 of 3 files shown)


phlptp:
I haven't tried on other distros,  from what I have been reading if we were to compile on adroid this flag would also be needed, so it might be more general for common ARM processors.  

The idea was to embed the specialized configurations that can't be easily detected in CMake into a single variable that was a little more readable.    How I got it to build initially was modifying the CMAKE_CXX_FLAG,  but that seems like a little specialized knowledge to require that.    If we require a separate script then we have to supply and maintain that script rather than having it built in to the cmake configuration already.    I suppose there are are pros and cons.  and am certainly open to other options,  mainly just wanted to capture the requirements to build with a raspberry pi in some way that is easy to access and doesn't require remembering some obscure build flag.

phlptp:
I still want to try with some other PI distributions at some point but I haven't got around to that yet.

phlptp:
I think if we figure out some better way of handling this situation(an others like it ) in the future was can update.

PR #1223, opened 2020-04-23
Deprecate error method in federate

phlptp:
### Summary

If merged this pull request will deprecate the error method in the Application_api federate object.  this function is duplicative of localError and `localError` or `globalError` should be used for better control over the error process.

### DIFF
--- a/tests/helics/application_api/FederateTests.cpp
+++ b/tests/helics/application_api/FederateTests.cpp
@@ -758,7 +758,7 @@ TEST(federate_tests, forceErrorExec)
     auto Fed1 = std::make_shared<helics::Federate>("fed1", fi);
 
     Fed1->enterExecutingMode();
-    Fed1->error(9827);
+    Fed1->localError(9827);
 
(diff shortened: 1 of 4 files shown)


phlptp:
This is a duplicate of #1222 with a different branch name so some tests don't fail due to the word error in the branch name

tests/helics/application_api/FederateTests.cpp
```diff
@@ -810,7 +810,7 @@ TEST(federate_tests, forceErrorPendingTimeAsync)
 
     Fed1->enterExecutingMode();
     Fed1->requestTimeAsync(2.0);
-    Fed1->error(9827);
+    Fed1->globalError(9827);
```
nightlark, line 813: Was making this test use `globalError` intentional?

phlptp: sort of.  It doesn't impact the test results and I wanted the function used for code coverage purposes.  This was an easy way to do that.    local or Global could be used interchangeably here.

PR #1235, opened 2020-04-25
Java interface build refactor

nightlark:
### Summary
If merged this pull request will add standalone build support for the Java interface.

### Proposed changes
- Adds standalone build support for the Java interface
- Moves CMake project for the build jar step into a subfolder (avoids cache errors from CMake)
- Cleans up the extra files installed in the java folder on platforms supporting rpaths


interfaces/java/CMakeLists.txt
```diff
@@ -7,143 +7,289 @@
...
+
+    endif()
+
+    # Add the local directory to the rpath
+    if(NOT APPLE)
+        list(APPEND CMAKE_INSTALL_RPATH $ORIGIN)
+        list(APPEND CMAKE_INSTALL_RPATH
+            "\$ORIGIN/../${CMAKE_INSTALL_LIBDIR}"
+        )
+    else()
```
phlptp, line 46: what does this do on Windows?

phlptp:
Do we have some tests for the new code paths in cmake?

interfaces/java/CMakeLists.txt
nightlark: Nothing, Windows doesn't support rpaths.

nightlark:
Travis has tests covering the Swig and no swig paths, Appveyor and the nightly run of the release builds execute the Windows specific file copying (already caught an issue with the refactor).

phlptp:
We should probably have a set of tests that build the interfaces to test the standalone paths if we do not already.

phlptp:
@nightlark  Are you planning on doing anything more to these PR's?

phlptp:
I guess in one of them the changelog should be updated to reflect the change.

nightlark:
No other changes for this release... Ideally with CI test plan changes, one of the builds would create a HELICS installation as part of a build stage (testing packaging?), then a subsequent stage would have one or more jobs that test each of the standalone interface builds using the install from the previous stage (with + without swig). Maybe it would also be a good idea to make the interface tests work without building the rest of HELICS.

Getting matlab running on a CI service could be part of setting up the standalone interface build tests.

PR #1236, opened 2020-04-25
Matlab interface build refactor

nightlark:
### Summary
If merged this pull request will add standalone build support for the Matlab interface

### Proposed changes
- Adds standalone build support for the Java interface

### DIFF
--- a/interfaces/matlab/CMakeLists.txt
+++ b/interfaces/matlab/CMakeLists.txt
@@ -7,229 +7,371 @@
 # SPDX-License-Identifier: BSD-3-Clause
 # ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-# Generate SWIG wrapper for MATLAB
-if(HELICS_SWIG_GENERATE_INTERFACE_FILES_ONLY)
-    include(${CMAKE_CURRENT_SOURCE_DIR}/matlabSwigGenerateOnly.cmake)
-else()
-    find_package(Matlab COMPONENTS MAIN_PROGRAM)
+cmake_minimum_required(VERSION 3.4)
+project(helicsMATLAB)
 
-    if(HELICS_ENABLE_SWIG AND SWIG_EXECUTABLE)
-        # check to make sure the swig executable supports matlab
-        execute_process(
-            COMMAND ${SWIG_EXECUTABLE} -help
-            OUTPUT_VARIABLE SWIG_HELP_OUTPUT
+get_directory_property(hasParent PARENT_DIRECTORY)
+
+if(NOT hasParent)
+    message(STATUS "Standalone MATLAB interface build")
+
+    if(CMAKE_VERSION VERSION_LESS 3.16)
+        cmake_policy(VERSION ${CMAKE_MAJOR_VERSION}.${CMAKE_MINOR_VERSION})
+    else()
+        cmake_policy(VERSION 3.16)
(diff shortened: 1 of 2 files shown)


interfaces/matlab/CMakeLists.txt
```diff
@@ -7,229 +7,371 @@
 # SPDX-License-Identifier: BSD-3-Clause
 # ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-# Generate SWIG wrapper for MATLAB
-if(HELICS_SWIG_GENERATE_INTERFACE_FILES_ONLY)
-    include(${CMAKE_CURRENT_SOURCE_DIR}/matlabSwigGenerateOnly.cmake)
-else()
-    find_package(Matlab COMPONENTS MAIN_PROGRAM)
+cmake_minimum_required(VERSION 3.4)
```
phlptp, line 10: don't we normally use 3.5?

interfaces/matlab/CMakeLists.txt
```diff
@@ -7,229 +7,371 @@
...
+    string(FIND "${SWIG_HELP_OUTPUT}" "-matlab" MATLAB_HELP_FOUND)
+    if(${MATLAB_HELP_FOUND} LESS 0)
+        set(MATLAB_SWIG_NOT_AVAILABLE 1)
+        message(
+            WARNING
+                " SWIG VERSION does not support Matlab, reverting to build only"
         )
     endif()
 
+    # Find helics.i swig interface file
```
phlptp, line 83: is this the same code we use in the other interface modules?  if so it something that could be extracted to a separate file or would that make things more complicated?

nightlark: It would be convenient to have it in a separate file, but is more complicated. It needs to be accessible for building as a subproject of HELICS and when building standalone (ideally without the assumption that it is part of the HELICS source tree).

interfaces/matlab/CMakeLists.txt
nightlark: 3.4 is what we use in most places. The benchmarks have a line setting it to 3.5.1 and package_config_tests set it to 3.6.0, everywhere else uses 3.4 (other than 3rd party).

phlptp:
I was thinking of GridDyn we use 3.5 there.


---
Sent from Workspace ONE Boxer<https://whatisworkspaceone.com/boxer>

On April 25, 2020 at 6:36:19 PM PDT, Ryan Mast <notifications@github.com> wrote:

@nightlark commented on this pull request.

________________________________

In interfaces/matlab/CMakeLists.txt<https://github.com/GMLC-TDC/HELICS/pull/1236#discussion_r415185975>:

> @@ -7,229 +7,371 @@
 # ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

-# Generate SWIG wrapper for MATLAB
-if(HELICS_SWIG_GENERATE_INTERFACE_FILES_ONLY)
-    include(${CMAKE_CURRENT_SOURCE_DIR}/matlabSwigGenerateOnly.cmake)
-else()
-    find_package(Matlab COMPONENTS MAIN_PROGRAM)
+cmake_minimum_required(VERSION 3.4)


3.4 is what we use in most places. The benchmarks have a line setting it to 3.5.1 and package_config_tests set it to 3.6.0, everywhere else uses 3.4 (other than 3rd party).

Issue #821, opened 2019-09-21
xSDK compatibility


phlptp:
For better compatibility with HPC systems and smoother interaction with other scientific software this might not be a bad goal to aspire to.

[xSDK](https://figshare.com/articles/xSDK_Community_Package_Policies/4495136) compatibility

I don't think we are that far off but there are definitely a few things we don't meet yet.  
I think compatibility might be a goal for HELICS 3.0.  and some discussion might be warranted on whether we pursue this officially or unofficially.

phlptp:
For anyone interested here is a webinar on the topic

https://www.exascaleproject.org/event/software-policies/

phlptp:
XSDK compatibility checklist
- [ ] M1 Required CMAKE options
  - [ ] USE_XSDK_DEFAULTS
  - [ ] XSDK_ENABLE_<language>
  - [ ] Machine readable provenance
- [ ] M2 test suite
- [ ] M3 no direct use of MPI_COMM_WORLD
- [ ] M4 Portability
- [ ] M5 Reliable contact information
- [ ] M6 Respect system resources & settings
- [ ] M7 Permissible license
- [ ] M8 Full version info available at runtime
  - [ ] runtime configuration options listing
  - [ ] run time version and build info for dev builds
- [ ] M9 use proper namespacing
- [ ] M10 Use version control
- [ ] M11 No hard wired print
  - [ ] option to turn off all printing
  - [ ] all logging to a single thread
- [ ] M12 external dependencies must be replaceable
- [ ] M13 Honor install prefix
- [ ] M14 64 bit pointers by default
- [ ] M15 Compatibility is sustainable
- [ ] M16 production quality installation process 
  - [ ] Spack

Recommended 
- [ ] R1 Public source code
- [ ] R2 Use Debugging tools, valgrind, etc
- [ ] R3 Error reporting is consistent
- [ ] R4 Release system resources as soon as possible
- [ ] R5 Export an ordered list of dependencies
- [ ] R6 Document versions of dependencies

nightlark:
I think we might meet requirement `M3 no direct use of MPI_COMM_WORLD`, we either use the a user provided communicator or a duplicate of MPI_COMM_WORLD.

PR #1255, opened 2020-04-29
add build flags call

phlptp:
### Summary

If merged this pull request will add a buildFlags and compierVersion call to the C API and some constexpr values to the helicsVersion Header.


phlptp:
Addresses M8 in xSDK compatibility Issue #821

PR #1258, opened 2020-04-29
Try gcc10

phlptp:
### Summary

If merged this pull request will update the circleCI build to include a nightly gcc 10 build,  and change the tsan build to use gcc instead of clang which isn't work right for some reason, or it is doing something odd.  

### Proposed changes

-


phlptp:
@nightlark  the gcc10 build had some warnings come up on the netif code.   https://circleci.com/gh/GMLC-TDC/HELICS/12330.    Not sure we will be running with that flag very often but if it doesn't cause any issues might be worth fixing at some point.

PR #1273, opened 2020-05-06
update clang-format,

phlptp:
### Summary

If merged this pull request will update clang-format to include  2 spaces before trailing comment, and align after open bracket.


phlptp:
@kdheepak  what do you think, worth making the change?

kdheepak:
Yeah, I think I like it a lot actually. It seems to improve readability consistently across the board. I think the only drawback is that if you change the function name the git diff is going to be as many lines as there are arguments, but seems like a worthy tradeoff.

kdheepak:
What are the odd cases you mention @nightlark?

nightlark:
https://github.com/GMLC-TDC/HELICS/pull/1274/files#diff-a43d5d06807e86b45e25c1a9c6321f2bR34-R35 looks kinda odd with both settings.

There was a case similar to https://github.com/GMLC-TDC/HELICS/pull/1274/files#diff-ade3c524b059270cd542ac5434e3ce9cR31-R33 (except using a curly brace, I think) where it would have been nice having the closing parenthesis/brackets on separate lines. The previous setting didn't handle this well either.

Issue #1078, opened 2020-02-06
automated formatting of cmake files


phlptp:
We already do automated formatting of the C++ and hpp files.  It would be nice if we could do the same with the cmake files to keep everything clean and consistent.

We have cmake format but it currently depends on someone running that regularly and pushing it.  automating it would save some time.

nightlark:
I think the formatting should be limited to indentation, unlike with clang-format, cmake_format can make changes that break things.

Or an alternative to cmake_format that format things while ensuring it still works the same way.

phlptp:
Other than that one issue with the long strings, what other problems have we encountered.  I have run the cmake-format a number of times on most of the cmake code, and apart from the one issue we encountered haven't had any issues.

nightlark:
I think it's just the long strings -- I don't really trust it to not break things if it's allowed to change where line breaks within a command happen.

phlptp:
I would expect a level of code review and CI tests to catch most if not all of the potential issues.

nightlark:
It looks like setting line_width to some large number would probably keep it from happening and still format functions with lots of arguments in a vertical layout -- that would make it work with the 3 instances currently using `# cmake-format: off` without breaking a long string or a string arg that ends up too close to the end of the line cut-off.

Do we want it to create PRs like the clang-format action does? Would we want to have a combined formatting action that opens a single formatting PR with two separate commits?

PR #1275, opened 2020-05-06
add cmake-format to the pre-commit checks and updates

phlptp:
### Summary

If merged this pull request will add cmake-format to the pre-config list.  FIx issue #1078

### Proposed changes

- add cmake-format pre-commit to the formatting checks


nightlark:
It looks like it's better than before at not splitting quoted arguments across lines. It seems to mess up comments in several cases.

- commented out code blocks
- `######` section type comments get a space added after the first `#` (a few commented out files also got this
- merging commented lines that were intended to be separate; separate sentences on different lines, or commented out blocks of code (which may be better to remove entirely)

phlptp:
Most of the ####### blocks that we used have were turned into # ----------------------- or something to that effect a while ago, so we could do that to the remaining ones to be consistent.  I don't really think the extra ' ' is that much of an issue.  

The reflow of comments on the code sections is a little more problematic, but the reality is those should probably be removed anyway right now so not likely that big of an issue either.

examples/BrokerServerTestCase/CMakeLists.txt
```diff
@@ -14,13 +14,10 @@ target_link_libraries(brokerServerTestCase PUBLIC HELICS::apps)
...
-    #    NAME brokerServerTestCase
-    #    COMMAND
-    #        ${PROJECT_SOURCE_DIR}/scripts/test_example.sh 10
-    #        $<TARGET_FILE:nonlings_fed1_cpp> $<TARGET_FILE:nonlings_fed2_cpp>
-    #)
-    #set_property(TEST nonlings_example PROPERTY LABELS Continuous)
+    # add_test(
+    # NAME brokerServerTestCase COMMAND ${PROJECT_SOURCE_DIR}/scripts/test_example.sh 10
+    # $<TARGET_FILE:nonlings_fed1_cpp> $<TARGET_FILE:nonlings_fed2_cpp> ) set_property(TEST
+    # nonlings_example PROPERTY LABELS Continuous)
```
phlptp, line 20: @nightlark  is this whole if clause something we need?

nightlark: I think we can get rid of it.

phlptp:
I made one more tweak that cleaned up a few cases a little better, and got rid of that commented out section.

benchmarks/helics/CMakeLists.txt
```diff
@@ -34,17 +34,17 @@ set(HELICS_MULTINODE_BENCHMARKS
     TimingLeafFederate
 )
 
-if (NOT DEFINED HELICS_BENCHMARK_SHIFT_FACTOR)
+if(NOT DEFINED HELICS_BENCHMARK_SHIFT_FACTOR)
     include(ProcessorCount)
-    ProcessorCount(N)
-    if (N LESS 5)
-       set (HELICS_BENCHMARK_SHIFT_FACTOR -1)
+    processorcount(N)
```
nightlark, line 39: I guess this is okay since CMake commands are case insensitive.

PR #1287, opened 2020-05-08
add markdownlint check

phlptp:
### Summary

If merged this pull request will add markdownlint check and fix issues


nightlark:
I'm not sure about the replacing single quotes with double quotes in the yaml files. Some of those are shell scripts that could be using single quotes to avoid variable substitution.

phlptp:
Are there specific files that are likely to be an issue? It would be pretty easy to exclude specific files from the formatting.

nightlark:
I'm looking at the yaml files for configuring CI builds -- `.ci/`, `.circleci/`, `.travis.yml`, and `.github/`. I don't know that replacing single quotes with double quotes for values in a yaml file is entirely correct in general, since the single quotes could be getting used for a string literal without escape characters.

phlptp:
@nightlark  finally got it to pass the markdownlint tests.  I think you can review the whole thing now.

PR #1301, opened 2020-05-11
merge the formatting actions to all use pre-commit

phlptp:
### Summary

If merged this pull request will merge the clang-formatting action into the pre-commit action


phlptp:
once this gets merged to master, We can probably remove the buildenv clang-format image since we won't be using that anymore.

PR #1302, opened 2020-05-11
add shell check and shell format to pre-commit

phlptp:
### Summary

If merged this pull request will add shell check and shfmt to pre-commit


phlptp:
@nightlark  I am going to let you fix this, since I am not totally sure what is going on with all the scripts

PR #1317, opened 2020-05-13
Use assign error

phlptp:
### Summary

If merged this pull request will refactor the shared_library to use an assignError function which simplifies the check and the code.  also fix a warning about variable name shadowing.

### DIFF
--- a/src/helics/shared_api_library/internal/api_objects.h
+++ b/src/helics/shared_api_library/internal/api_objects.h
@@ -155,6 +155,15 @@ class QueryObject {
         }                                                                                                                                  \
     } while (false)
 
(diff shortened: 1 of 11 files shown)


src/helics/shared_api_library/internal/api_objects.h
```diff
@@ -155,17 +155,23 @@ class QueryObject {
         }                                                                                                                                  \
     } while (false)
 
+/** assign and error string and code to an error object if it exists*/
```
nightlark, line 158:
```suggestion
/** assign an error string and code to an error object if it exists*/
```

PR #1332, opened 2020-05-16
Require cxx17

phlptp:
### Summary

If merged this pull request will update the minimum builds to require C++17 and CMAKE 3.10


phlptp:
This needs an updated container library to be merged first before it will work on windows.  Which is done but some tests weren't passing so I am rebuilding some of the builders to work on that.

phlptp:
Travis and appveyor aren't going to pass anytime soon,  we need to figure out what to do with appveyor and travis needs to be updated with the correct builds.  I think we should leave at least one/two regular builds on travis with a few more nightly.  

Appveyor I was toying with trying to get a cygwin build going.    

But in any case they aren't going to be working on this PR.  

The 32 bit windows build is interesting,  Had to disable the webserver, since apparently there is a bug in visual studio that is triggered in certain cases by the code in beast with certain numbers of processors and varying versions of visual studio.   Doesn't do it on my system, just on azure.  So the webserver build is disabled for 32 bit MSVC 2017,  it works fine on 2019.    Will probably need to make a note of that in the webserver documentation at some point.

PR #1341, opened 2020-05-18
remove mpark::variant to use std::variant

phlptp:
### Summary

If merged this pull request will remove the use of mpark::variant in favor of std::variant


phlptp:
There will be a couple more of these type of things to remove any, optional, and string view to use the standard library versions

PR #1349, opened 2020-05-19
Json cpp to upstream

phlptp:
### Summary

If merged this pull request will switch the jsoncpp library to the jsoncpp repo instead of a fork

### Proposed changes

- switch jsoncpp to upstream repo
- update utilities library to latest master
- start switching appveyor to a cygwin build
- a number of fixes related to the other changes and getting cygwin build to work


appveyor.yml
```diff
@@ -1,31 +1,36 @@
...
+    - '%CYG_ROOT%\setup-x86_64.exe --quiet-mode --no-shortcuts --upgrade-also --packages cmake,cygwin-devel,git,gcc-core,gcc-g++,libboost-devel,libzmq-devel'
+    - '%CYG_ROOT%\bin\cygcheck -dc'
+    - cd %APPVEYOR_BUILD_FOLDER%
+    - git submodule update --init --recursive
 
 before_build:
   - mkdir build
   - cd build
-  - cmake .. -G "Visual Studio 14 2015 Win64" -DBOOST_INSTALL_PATH=C:/libraries/boost_1_63_0 -DHELICS_ENABLE_SWIG=OFF -DHELICS_BUILD_TESTS=ON -DHELICS_BUILD_BENCHMARKS=ON -DHELICS_BUILD_EXAMPLES=ON -DHELICS_BUILD_CXX_SHARED_LIB=ON -DBUILD_PYTHON_INTERFACE=ON -DBUILD_JAVA_INTERFACE=ON
+  - '%CYG_BASH% -lc "cd $APPVEYOR_BUILD_FOLDER\build && cmake ..  -DHELICS_ENABLE_SWIG=OFF -DHELICS_BUILD_TESTS=ON -DHELICS_DISABLE_ASIO=ON"'
```
nightlark, line 29: Probably need backslashes for some of these. If forward slashes are supported, it might easier using them to avoid escaping issues.
```suggestion
  - '%CYG_BASH% -lc "cd $APPVEYOR_BUILD_FOLDER\\build && cmake ..  -DHELICS_ENABLE_SWIG=OFF -DHELICS_BUILD_TESTS=ON -DHELICS_DISABLE_ASIO=ON"'
```

phlptp:
@nightlark  do you have any idea what is wrong with the packaging tests on appveyor

nightlark:
It looks like most of the failed appveyor builds so far have been during the configure step.

phlptp:
The azure windows build are what I am trying to figure out.  the appveyor stuff was just an aside,  don't really care about it at the moment but I don't understand what is going wrong with the azure builds.

CMakeLists.txt
```diff
@@ -100,6 +100,14 @@ if(MSYS
     set(UNIX_LIKE TRUE)
 endif()
 
+# Set the build output paths
+if (CMAKE_PROJECT_NAME STREQUAL PROJECT_NAME)
```
nightlark, line 104: https://github.com/GMLC-TDC/HELICS/blob/develop/interfaces/python/CMakeLists.txt#L13-L15

phlptp: I suppose either would work, both have possible incorrect results, this has the advantage of no additional variable, the other is a little more obvious what it is doing.   I think I switch to this to move away from additional variables but could move back.

nightlark: PARENT_DIRECTORY should only be set if it has been added as a subdirectory of the build. What are situations where it would have incorrect results?

phlptp: well if CMakeLIst.txt were included in a top level directory via include() or add_subdirectory were called without defining a project.    Not totally sure those situations are allowed by CMake or not.

nightlark: It looks like not defining a project is okay, since it uses the path of the CMakeLists.txt/source directory that added it as a subdirectory.

include() does break it... using include for an external dependency seems wrong (and a CMakeLists.txt file), but with the number of results include(*CMakeLists.txt) gives...

phlptp: I think in the case of add_subdirectory with no parent project we still would want HELICS to act as if it were the parent project and set everything accordingly.

phlptp: In nearly all cases either would work.  This is what was recommended by some folks working on the CLI11 library and some folks who have given tutorials on CMake, so is what I have been using for a while.

phlptp:
I think this turned into a lot more involved than I first intended but generally I like the changes.  So I think this is ready for review.

phlptp:
@kdheepak  this ends up changing the output directory of the build, so all the executable and libraries get placed in the same directory.  We might want to consider making the same change on develop.

phlptp:
@kdheepak @nightlark   One big question is whether we should have the develop branch use the same build, as in make all the libraries and binaries go to the same folder?

kdheepak:
I think I'm not following. Will this change what we are doing right now, i.e placing binaries in `bin` and libraries in `lib`? What do you mean by changing the output directory of the build? I'm not seeing the relevant change in this PR that would do this.

phlptp:
This doesn't affect the installation at all.  

What it does change is instead of creating the build executables and libraries in a location relative to the source path it now places them all  build/bin  or build/lib.  This means everything from the archive files, debugging files, executables, shared libs.    It simplifies a few things on windows as we no longer need to copy around the files.  There may be some extra complexity on the interface builds yet that we haven't resolved yet.    

Again won't impact the installation but it is a pretty big change in how the builds are organized.

PR #1351, opened 2020-05-20
Use enum for all options

kdheepak:
### Summary

If merged this pull request will make the C api use enum for all the argument types.

### DIFF
--- a/src/helics/shared_api_library/MessageFederate.h
+++ b/src/helics/shared_api_library/MessageFederate.h
@@ -489,7 +489,7 @@ HELICS_EXPORT int helicsMessageGetMessageID(helics_message_object message);
  *
  * @return The flags associated with a message.
  */
-HELICS_EXPORT helics_bool helicsMessageCheckFlag(helics_message_object message, helics_federate_flags flag);
+HELICS_EXPORT helics_bool helicsMessageCheckFlag(helics_message_object message, int flag);
 
 /**
  * Get the size of the data payload in bytes.
--- a/src/helics/shared_api_library/MessageFederate.h
+++ b/src/helics/shared_api_library/MessageFederate.h
@@ -402,7 +402,7 @@ HELICS_EXPORT void helicsEndpointSetInfo(helics_endpoint end, const char* info, 
  * @param[in,out] err An error object to fill out in case of an error.
  * @endforcpponly
  */
-HELICS_EXPORT void helicsEndpointSetOption(helics_endpoint end, int option, helics_bool value, helics_error* err);
+HELICS_EXPORT void helicsEndpointSetOption(helics_endpoint end, helics_handle_options option, helics_bool value, helics_error* err);
 
 /**
  * Set a handle option on an endpoint.
@@ -410,7 +410,7 @@ HELICS_EXPORT void helicsEndpointSetOption(helics_endpoint end, int option, heli
  * @param end The endpoint to modify.
  * @param option Integer code for the option to set /ref helics_handle_options.
  */
-HELICS_EXPORT helics_bool helicsEndpointGetOption(helics_endpoint end, int option);
+HELICS_EXPORT helics_bool helicsEndpointGetOption(helics_endpoint end, helics_handle_options option);
 
 /**
  * \defgroup Message operation functions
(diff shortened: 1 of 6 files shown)


src/helics/shared_api_library/MessageFederate.h
```diff
@@ -489,7 +489,7 @@ HELICS_EXPORT int helicsMessageGetMessageID(helics_message_object message);
  *
  * @return The flags associated with a message.
  */
-HELICS_EXPORT helics_bool helicsMessageCheckFlag(helics_message_object message, int flag);
+HELICS_EXPORT helics_bool helicsMessageCheckFlag(helics_message_object message, helics_federate_flags flag);
```
kdheepak, line 492: I'm pretty sure this shouldn't be `helics_federate_flags` but not sure what it should be.

src/helics/shared_api_library/helics.h
```diff
@@ -764,7 +764,7 @@ HELICS_EXPORT void
  * @param[in,out] err An error object that will contain an error code and string if any error occurred during the execution of the function.
  * @endforcpponly
  */
-HELICS_EXPORT void helicsFederateInfoSetIntegerProperty(helics_federate_info fi, int intProperty, int propertyValue, helics_error* err);
+HELICS_EXPORT void helicsFederateInfoSetIntegerProperty(helics_federate_info fi, helics_properties intProperty, int propertyValue, helics_error* err);
```
kdheepak, line 767: Is `helics_properties` a subset of what can be set over here? This should be reviewed.

src/helics/shared_api_library/MessageFederate.h
phlptp: The message has a 16 bit flag field some of which are purposely left user defined, so there is a set which HELICS uses but others which it doesn't, so if we used an enum it would need to define 16 enumerations from 0-15

src/helics/shared_api_library/helics.h
```diff
@@ -632,7 +632,7 @@ HELICS_EXPORT void helicsFederateInfoSetBrokerInitString(helics_federate_info fi
  * @param[in,out] err An error object that will contain an error code and string if any error occurred during the execution of the function.
  * @endforcpponly
  */
-HELICS_EXPORT void helicsFederateInfoSetCoreType(helics_federate_info fi, int coretype, helics_error* err);
+HELICS_EXPORT void helicsFederateInfoSetCoreType(helics_federate_info fi, helics_core_type coretype, helics_error* err);
```
phlptp, line 635: in HELICS 3 (at some point) we will allow user defined core types in which case helics can't know about all possibilities

src/helics/shared_api_library/helics.h
phlptp: The biggest issue is that if we are (eventually) to allow user defined core-types, then we need to allow user defined properties on those core types.

kdheepak:
Are the remaining changes okay? This PR is useful to have when wrapping in languages like Julia and Nim, since we can get compile time checks.

src/helics/shared_api_library/helics.h
nightlark: Is `helics_core_type` something that could become a typedef, with a separate enum (`helics_builtin_core_types`) for any that come with HELICS?

phlptp:
The endpoint, input, publication, filter, set/get options I think I could be ok with.  Those aren't likely to be user customized.  The other ones are all subject to user defined federates and core types, which means the set isn't bounded very well.

nightlark:
It looks like these changes require updating the cpp98 wrapper.

phlptp:
What do we want to do with this.  Can't merge it into helics 3 as we have locked the API now.  There would be a lot of issues we would need to work out to contemplate using enums for the properties.  So not sure where to go with this right now

kdheepak:
This is probably the best (only?) time to make this breaking change though. It'll also make it easier to generate interfaces in other languages. But I don't think this change is critical. It just avoids bugs when using the C API directly. I think all these changes are handled in Python, which is our most used interface. It would be nice to have for SWIG to generate better interfaces in C#, Java, etc

phlptp:
I think we are still planning on a v4.0 which would likely be early 2023, so will get at least one more chance in the HELICS+ project.

PR #1352, opened 2020-05-20
(Modified) Watts-Strogatz Benchmark

nightlark:
### Summary
If merged this pull request will add a benchmark that implements a modified Watts-Strogatz network for communication. The key difference from the [Watts-Strogatz model described here](https://en.wikipedia.org/wiki/Watts%E2%80%93Strogatz_model) is that the links constructed by individual federates are directional.

From trying some different default parameters, increasing the initial message count created by each federate can make the benchmark hang when using the udp core as it gets near ~50 messages. Also, a large number of initial messages (100) resulted in the lower degrees of connectivity (1 link) being slower than higher degrees of connectivity (I think this is something that one of the message send benchmarks already exercises).

### Proposed changes
- add a modified Watts-Strogatz benchmark
- some clean-up in the PHOLD federate (remove todo and methods implemented in the base class)

### DIFF
--- a/benchmarks/helics/PholdFederate.hpp
+++ b/benchmarks/helics/PholdFederate.hpp
@@ -46,8 +46,6 @@ class PholdFederate: public BenchmarkFederate {
     std::uniform_int_distribution<unsigned int> rand_uniform_int;
 
   public:
-    // TODO(@nightlark): output cluster name in a summary file, along with node count and feds per
-    // node
     PholdFederate(): BenchmarkFederate("PHOLD") {}
 
     // functions for setting parameters
@@ -57,16 +55,6 @@ class PholdFederate: public BenchmarkFederate {
     void setInitialEventCount(unsigned int count) { initEvCount_ = count; }
     void setLocalProbability(double p) { localProbability_ = p; }
     void setLookahead(double v) { lookahead_ = v; }
-
-    // functions for setting callbacks
-    void setBeforeFinalizeCallback(std::function<void()> cb = nullptr)
-    {
-        callBeforeFinalize = std::move(cb);
-    }
-    void setAfterFinalizeCallback(std::function<void()> cb = nullptr)
-    {
-        callAfterFinalize = std::move(cb);
-    }
 
     std::string getName() override { return "phold_" + std::to_string(index); }
 
--- a/benchmarks/helics/WattsStrogatzFederate.hpp
+++ b/benchmarks/helics/WattsStrogatzFederate.hpp
@@ -50,10 +50,10 @@ class WattsStrogatzFederate: public BenchmarkFederate {
     WattsStrogatzFederate(): BenchmarkFederate("WattsStrogatzFederate") {}
 
     // functions for setting parameters
-    void setGenerateRandomSeed(bool b) { generateRandomSeed = b; };
-    void setRandomSeed(unsigned int s) { seed = s; };
-    void setDegree(int val) { k = val; };
-    void setRewireProbability(double val) { b = val; };
+    void setGenerateRandomSeed(bool b) { generateRandomSeed = b; }
+    void setRandomSeed(unsigned int s) { seed = s; }
+    void setDegree(int val) { k = val; }
+    void setRewireProbability(double val) { b = val; }
 
     std::string getName() override { return getNameForIndex(index); }
     std::string getNameForIndex(int i) { return "watts_" + std::to_string(i); }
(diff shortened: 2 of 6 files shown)


benchmarks/helics/PholdFederate.hpp
```diff
@@ -58,16 +56,6 @@ class PholdFederate: public BenchmarkFederate {
     void setLocalProbability(double p) { localProbability_ = p; }
     void setLookahead(double v) { lookahead_ = v; }
 
-    // functions for setting callbacks
```
phlptp, line 61: Did we not need these anymore?

benchmarks/helics/WattsStrogatzFederate.hpp
```diff
@@ -0,0 +1,206 @@
...
+    // Federates, this simplified version has each node setup the connectiosn on its right side only
+    // (so k=K/2 in this benchmark)
+    int k{1};  // degree
+    double b{0};  // re-wire probability for each of k rightmost neighbors
+
+    // Classes related to the exponential and uniform distribution random number generator
+    bool generateRandomSeed{false};
+    unsigned int seed{0xABad5eed};  // suggestions for seed choice were not having a majority of the
+                                    // bits as 0 is better
+    std::mt19937 rand_gen;
```
phlptp, line 39: how much variation is there depending on the seed.  It looks like in the default case you have the seed the same for each case, which could make the test skewed in some fashion.

benchmarks/helics/WattsStrogatzFederate.hpp
```diff
@@ -0,0 +1,206 @@
...
+            if (edge < 0) {
+                edge += maxIndex;
+            }
+            availableEdges.push_back(edge);
+        }
+
+        // Re-wire links
+        if (!availableEdges.empty()) {
+            // NOLINTNEXTLINE(modernize-loop-convert)
+            for (unsigned int i = 0; i < currentEdges.size(); i++) {
```
phlptp, line 142: why can't this loop be modernized?

benchmarks/helics/WattsStrogatzFederate.hpp
```diff
@@ -0,0 +1,206 @@
...
+                    availableEdges[newLink] = currentEdges[i];
+                    currentEdges[i] = tmp;
+                } else {
+                    // No link rewiring for this link
+                }
+            }
+        }
+
+        // Confirm one last time that the number of edges matches the degree
+        if (currentEdges.size() != static_cast<unsigned int>(k)) {
```
phlptp, line 163: shouldn't this static_cast be to size_t?

benchmarks/helics/PholdFederate.hpp
nightlark: They're handled by the parent class.

benchmarks/helics/WattsStrogatzFederate.hpp
nightlark: It's swapping values in the arrays.

benchmarks/helics/WattsStrogatzFederate.hpp
```diff
@@ -0,0 +1,206 @@
...
+                if (rand_rewire(rand_gen) < b) {
+                    // Criteria for the new link:
+                    // 1. Avoid self-loops
+                    // 2. Avoid duplication with links that exist at this point in the algorithm
+
+                    // Pick a new link from the available edges
+                    auto newLink = rand_available_link(rand_gen);
+
+                    // Swap current link with available edge
+                    auto tmp = availableEdges[newLink];
```
phlptp, line 153: can you use std::swap here?

benchmarks/helics/WattsStrogatzFederate.hpp
```diff
@@ -0,0 +1,206 @@
...
+        std::string txstring(100, '1');
+        for (int i = 0; i < initialMessageCount; i++) {
+            helics::Time evTime = fed->getCurrentTime() + deltaTime;
+            ept->send(links[rand_transmit_link(rand_gen)], txstring, evTime);
+        }
+    }
+
+    void doMainLoop() override
+    {
+        std::string txstring(100, '1');
```
phlptp, line 190: is this variable being used?

benchmarks/helics/wattsStrogatzBenchmarks.cpp
```diff
@@ -0,0 +1,180 @@
...
+#include <benchmark/benchmark.h>
+#include <chrono>
+#include <fstream>
+#include <gmlc/concurrency/Barrier.hpp>
+#include <iostream>
+#include <thread>
+
+using helics::core_type;
+
+static void WattsStrogatzArguments(benchmark::internal::Benchmark* b)
```
phlptp, line 22: A comment on what is doing would be helpful

phlptp:
should this be added to the run_all_benchmarks or not?

benchmarks/helics/WattsStrogatzFederate.hpp
nightlark: It looks like std::swap isn't changing the data in the arrays.

benchmarks/helics/WattsStrogatzFederate.hpp
nightlark: The variation as a result of other factors in test runs seemed to be larger than variation due to changing the seed.

nightlark:
For run all benchmarks -- it could be interesting to see, but it probably becomes more interesting when you have multiple nodes and directly communication routes between federates instead of relying on the broker for sending everything.

PR #1361, opened 2020-05-22
Tidy fixes2

phlptp:
### Summary

If merged this pull request will merge in recent clang-tidy fixes and cmake changes from the helics3 branch with the develop line

### Proposed changes

- clang-tidy fixes
- use CMAKE_RUNTIME_OUTPUT_LOCATION variable for storing all build output to the same location.


phlptp:
Not intentionally.  I did a merge so some may have snuck in.  If you see some change related to that let me know and I will revert it for this PR.

phlptp:
is github having some issues, or did this break something?

tests/CMakeLists.txt
```diff
@@ -36,12 +36,8 @@ if(hasParent)
     endif(NOT HELICS_SWIG_GENERATE_INTERFACE_FILES_ONLY)
 
 else()
-    cmake_minimum_required(VERSION 3.4)
-    if(CMAKE_VERSION VERSION_LESS 3.17)
-        cmake_policy(VERSION ${CMAKE_MAJOR_VERSION}.${CMAKE_MINOR_VERSION})
-    else()
-        cmake_policy(VERSION 3.17)
-    endif()
+    cmake_minimum_required(VERSION 3.10..3.17)
```
nightlark, line 39: Here's another CMake version related change.

phlptp:
@nightlark  any idea why the automatic PR generation is failing?

nightlark:
It could be related to this: https://www.githubstatus.com/

nightlark:
It looks like API requests are having issues, which could be why it's returning 403 forbidden.

nightlark:
Interesting -- the tests expect the helics.py file from swig and the compiled python extension _helics.so to be in `${CMAKE_BINARY_DIR}/interfaces/python`, which I guess are getting separated now with helics.py in `${CMAKE_BINARY_DIR}/interfaces/python` and `_helics.so` in the new bin/lib folder.

phlptp:
I was wondering about that a little, but I don't think we have an operational python test in the HELICS 3 branch right now.    What can we do about it.

nightlark:
We either need to copy the compiled language interface libraries (+ helicsSharedLib because rpaths probably aren't right) into the `interface/<lang>` subfolder, or copy the language wrapper (.py, .m, etc) into the bin/lib folder that has the compiled interface libraries (will be different on Windows and Linux/macOS).

phlptp:
we were not copying the libHelicsSharedLib into the folder before correct?

nightlark:
We were only copying libhelicsSharedLib on Windows before. On Linux/macOS the build rpath for the Python extension worked for finding libhelicsSharedLib in the build tree, but now the build rpath is set to expect libhelicsSharedLib in the lib folder instead (so both need to get copied).

phlptp:
@nightlark  I think this is a ready now

interfaces/python/CMakeLists.txt
```diff
@@ -277,17 +290,7 @@ install(FILES ${CMAKE_CURRENT_BINARY_DIR}/helics.py DESTINATION python COMPONENT
 if(WIN32)
     # Realistically for bundling into a Python wheel, the dll should be given a unique name to avoid
     # conflicts with other packages using a different version of the dll with the same name
-    add_custom_command(
```
nightlark, line 280: I think the copy for helicsSharedLib (+ libzmq if available) should remain for Windows builds.

phlptp: do you think we need to do the same for the other interface builds as well?

nightlark: Probably, because Windows doesn't have rpaths for finding libraries in other directories.

phlptp: Was hoping we could get rid of that but I guess not.  May have more tweaks when we start to test the matlab interface

Issue #1360, opened 2020-05-22
global action for generating PR's


phlptp:
The PR generation code we use is useful to many other libraries.  It should be extracted out and made into an action repo that can easily be used elsewhere and simplify the action scripts in HELICS and other repos.

nightlark:
The new action is at https://github.com/gmlc-tdc/create-pr-action

Some documentation (README + metadata) should be done to improve it and demonstrate use.

PR #1388, opened 2020-05-31
Switch local PR creation action to gmlc-tdc/create-pr-action

nightlark:
### Summary
If merged this pull request will move the local PR creation action to gmlc-tdc/create-pr-action.

#1386 and #1387 were test PRs opened by the moved action.

### Proposed changes
- Remove the local PR creation action
- Switch formatting workflow to use gmlc-tdc/create-pr-action@v0.1
- Switch swig-gen workflow to use gmlc-tdc/create-pr-action@v0.1


.github/workflows/swig-interface-gen.yml
```diff
@@ -43,3 +43,4 @@ jobs:
         GIT_EMAIL: "HELICS-bot@users.noreply.github.com"
         GIT_NAME: "HELICS-bot"
         BRANCH_PREFIX: "swig-gen/"
+        DISABLE_STAGING: true
```
phlptp, line 46: Do we do the stages in the script that generates the files?

phlptp:
You should probable add a readme to the new repo.  
I think automated is spelled incorrectly in the script.

phlptp:
addresses #1360

.github/workflows/swig-interface-gen.yml
nightlark: I added a bit to the action so that it stages changed files by default (should be the more common use case). The swig generating action disables staging because the step before stages just changed files in the `interfaces/` subfolder (maybe more conservative than needed).

nightlark:
> You should probable add a readme to the new repo.

Yea, also need to make sure the action.yml metadata about the inputs it takes is alright.

> I think automated is spelled incorrectly in the script.

Looks like it should be `automated` instead of `automatted`?

PR #1419, opened 2020-06-08
Allow setting Python build output folders when not run as a HELICS subproject

nightlark:
### Summary

If merged this pull request will allow users to set the Python interface build output folders when it is not built as a subproject of the HELICS build. Fixes the issue that prevented the Linux Python packages from getting built successfully.

### Proposed changes

- Only force setting the Python interface build output folders when building as a HELICS subproject


interfaces/python/CMakeLists.txt
```diff
@@ -231,14 +231,17 @@ if(HELICS_ENABLE_SWIG AND SWIG_EXECUTABLE)
...
     set_target_properties(${SWIG_MODULE_helics_REAL_NAME} PROPERTIES FOLDER interfaces)
-    set_target_properties(
-        ${SWIG_MODULE_helics_REAL_NAME} PROPERTIES RUNTIME_OUTPUT_DIRECTORY
-                                                   ${CMAKE_CURRENT_BINARY_DIR}
-    )
-    set_target_properties(
-        ${SWIG_MODULE_helics_REAL_NAME} PROPERTIES LIBRARY_OUTPUT_DIRECTORY
-                                                   ${CMAKE_CURRENT_BINARY_DIR}
-    )
+
```
phlptp, line 234: One question is what behavior we want if a user manually sets the CMAKE_RUNTIME_OUTPUT_DIR /CMAKE_LIBRARY_OUTPUT_DIR?  In that case do we need to detect it and not change this Even if this is not a subproject?

nightlark: Pretty sure the reason we set this was to make sure the python tests would keep working, which requires building the interface as a HELICS subproject. I think the only time a user would set them is if they are using as a subproject of a non-CMake build system that requires the output runtime and library files to be built in a specific place... if they're doing that then they are likely to want the C++ or C libraries rather than the Python interface.

So I don't know, its a fairly obscure case. In the top level CMakeLists.txt it should probably avoid overwriting them if a user has set them.

Issue #1374, opened 2020-05-27
Cereal and serialization


phlptp:
Right now we use Cereal as a serialization backend for common types.    I am slightly concerned by the lack of activity on that project in recent years.    

So a point of contemplation is whether we want to use cereal in HELICS 3.0, or mimic the minimum functionality we need directly and drop all of cereal.  Or use something else entirely.

The types we serialize are doubles, int64, strings, vectors of doubles, complex, vector of complex, and namedPoint (string+double).  The C++ supports more conversions using cereal but to my knowledge no one has used that and it isn't supported in any other languages, so I don't consider it a critical feature.  It would also allow a bit more flexibility and likely performance in handing the data serialization/deserialization.

trevorhardy:
Given I don't know much about this, don't understand the magnitude of work, and won't be the one doing it, I'll keep my comments general. 

My inclination is to do our own implementation. If we can find another library that we have reason to believe will be around for a while I would be in favor of using that but I'm fearful of switching to another library that also begins to fade away in a few years (hypothetically). We don't want to be in a position with limited funding and have unexpected complications due to the disappearance of support for a critical piece of infrastructure.

nightlark:
It seems the amount of active development in cereal is similar to boost serialization; neither has much happening.

If we do our own serialization for the limited subset of types, the two things I think we'd need to be careful of are little/big-endian conversion when needed, and that it works for sending data 32 and 64 bit systems. We aren't really using cereal for that much, so I think doing our own with this small set of types could be okay. It looks like we're doing our own serialization for some data structures like ActionMessages, so replacing cereal with something else for serializing values isn't a big change and wouldn't add much maintenance burden.

If we decide to serialize more complex data structures (user defined?), picking an existing library with implementations in multiple languages would probably be better than trying to write and maintain our own. MessagePack, FlatBuffers, Cap'n Proto, and Protocol Buffers would be some of the options to look at that seem to have pretty good adoption and are a bit more efficient than passing around JSON.

phlptp:
For development effort,  it is non-trivial, but also not an enormous effort, so there is a lot we need to be careful about but we do have some experience with it.

It would mean some loss of capabilities in the C++ API,  which might not be that big of deal since that capability is not present in the C or language API's.    

We probably should have a discussion about where we want to go with more complex structured values before finalizing a decision though so we have some idea where we might go with that first.

phlptp:
[...]

PR #1421, opened 2020-06-10
More string view

phlptp:
### Summary

If merged this pull request will pull out more of the mostly unused functions in the C++ interface.  With the aim by HELICS 3 to make the C++98 and the C++ interface mostly equivalent in operation.  A number of the API calls in HELICS 2 were somewhat trials to see what if any got used so many of those are being removed and will be deprecated in HELICS 2.6.  Also migrate more of the interfaces to use string_view


phlptp:
This is starting to setup for #1374.  Probably will take a couple steps after this one before that can be done.  But it is a start.

Issue #479, opened 2018-10-10
global time barrier


phlptp:
I am starting down the path of relaxing some of the interface registration requirements in HELICS 2.0.    The technical functionality was pretty much in place so the last step was removing the checks that prevented it.    But then the issue turns into how exactly this will interact with a simulation.  In a tightly coupled simulation where you are simply adding an interface that interacts with an already existing dependency there really shouldn't be much of an issue.   However in loosely coupled systems running in parallel this will introduce some potential indeterminate behavior since all the time control is distributed.   The most common use cases I can think of for this, even that probably doesn't matter.    But for some it might.

One solution I am considering is the introduction of a global time barrier capability,  which will tell all federates to block at or before a particular time until the barrier is cleared.  

The question is are there other use cases for barriers of this or similar nature that should be considered.  
The barrier would prevent federates from granting at or after the requested time.  

Also what should be done in the case of a time that is already past?

phlptp:
The federates themselves really wouldn't have to be aware of the time barrier.  The way it would function is a time block on the time coordinator which is part of the core layer.  This will prevent granting any time equal to or after the block.  I would anticipate the barriers being an extension of the existing block mechanic that is part of the filter operations.  So unless a federate is requesting the time barrier I don't anticipate a need for it to be aware of it, and it wouldn't impact the actual times being granted just when they are able to do so.

trevorhardy:
I think we've discussed this in the past @phlptp, but a global time barrier is essential (probably?) for controlling a federation for debugging purposes. If this feature existed now it would be possible to fairly quickly create demonstration or prototype versions of federation debugging co-sims; this would be very cool, if you ask me.

phlptp:
The time block has been added,  next steps will include making a little more adaptable and controllable for debugging

PR #1433, opened 2020-06-19
Global time block

phlptp:
### Summary

If merged this pull request will add a global time barrier for debugging.
 
### Proposed changes

- address #479 
- add a global time barrier and an ability to update it.  
- add some test for the global time barrier
- add time barrier functions to the C api
- add time barrier functions to the C++98 API


phlptp:
I will add some additional docs in a separate pull request as an FYI

PR #1435, opened 2020-06-21
add the frozen library and get rid of some hacky code for the actionM…

phlptp:
### Summary

If merged this pull request will add the frozen library to HELICS 3 and start to use it to replace a few hacks that were in place before.  

### Proposed changes

- add frozen lib
- use in ActionMessage
- remove a deprecated function from the Broker.hpp interface


interfaces/python/CMakeLists.txt.bak
```diff
@@ -0,0 +1,268 @@
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
nightlark, line 1: This file looks like it was accidentally added?

src/helics/core/ActionMessage.cpp
```diff
@@ -573,13 +582,15 @@ std::unique_ptr<Message> createMessageFromCommand(ActionMessage&& cmd)
     return msg;
 }
 
+
 static constexpr char unknownStr[] = "unknown";
 
 // done in this screwy way because this can be called after things have started to be deconstructed
 // so static consts can cause seg faults someday will change to frozen::  once we can use all of
 // C++14
```
nightlark, line 591: Does this PR resolve this comment?

interfaces/python/CMakeLists.txt.bak
phlptp: I will remove it.

src/helics/core/ActionMessage.cpp
phlptp: yes I will remove it

src/helics/core/ActionMessage.cpp
```diff
@@ -579,152 +587,131 @@ static constexpr char unknownStr[] = "unknown";
...
```
nightlark, line 588: The comment re: switching to frozen is still here.

PR #1438, opened 2020-06-22
All queries JSON

phlptp:
### Summary

If merged this pull request will convert all queries to respond with valid json strings

### Proposed changes

- string vectors are json string vector as are numerical vectors
- invalid queries respond with http response code and message.

### DIFF
--- a/src/helics/application_api/Federate.cpp
+++ b/src/helics/application_api/Federate.cpp
@@ -1037,7 +1037,7 @@ std::string Federate::query(const std::string& target, const std::string& queryS
         if (coreObject) {
             res = coreObject->query(target, queryStr);
         } else {
-            res = "#disconnected";
+            res = "{\"error\":{\"code\":410\n\"message\":\"Federate is disconnected\"\n}";
         }
     }
     return res;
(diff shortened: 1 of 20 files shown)


src/helics/application_api/Federate.cpp
```diff
@@ -1002,12 +1002,16 @@ std::string Federate::query(const std::string& queryStr)
 {
     std::string res;
     if (queryStr == "name") {
-        res = getName();
+        res.push_back('"');
+        res.append(getName());
+        res.push_back('"');
```
nightlark, line 1007: The places using this pattern aren't guaranteed to be valid JSON, especially if a user is controlling the value that gets placed in between the quotes.

I think a utility function that takes care of making sure these characters are escaped in the string getting appended:
```
Backspace - \b.
Form feed is - \f.
Newline - \n.
Carriage return - \r.
Tab is replaced with \t.
Double quote -  \"
Backslash - \\
```

Of those, I think backslash, tab, and double quote are the most likely ones that a user might try to use.

phlptp: wondering if I should run it through the JSON stringify and use jsoncpp as the escape mechanism to deal with anything odd.

nightlark: That might be easier to make sure it's a valid JSON string. At some point we should check if jsoncpp does anything other than escape characters for strings.

PR #1448, opened 2020-07-01
Add option to install benchmark executables

nightlark:
### Summary

If merged this pull request will add an option for installing the benchmark binaries. Tested the change with spack and it provides a potentially easy way to get the HELICS benchmarks built on any systems that support spack.

### Proposed changes

- Install HELICS benchmark binaries when they are built
- Include benchmarks component in installers when built with the benchmarks enabled

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -980,6 +980,7 @@ if(HELICS_ENABLE_PACKAGE_BUILD)
         java
         octave
         csharp
+        benchmarks
         cereal
     )
     if(WIN32)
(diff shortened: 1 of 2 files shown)


phlptp:
The cEcho_benchmarks are not being installed as presently modified?  is this on purpose or an oversight?

nightlark:
Something I wasn't sure about. I don't want to install benchmarks from every language interface (it should be quick to pull things like a python file separately as needed), but since the C echo benchmark uses the HELICS build system, it would take some time and be a bit of a hassle to build separately. It should probably get installed too.

Issue #1368, opened 2020-05-26
All string translations are JSON?


phlptp:
One possible change for HELICS 3 would be making all queries and string conversions to something else be valid json objects.    The biggest change is probably quoting strings and string vectors,  but it would allow more obvious translations to external languages and be more contextually easy to explain.  
many existing queries already return json, so having them all return valid json may simplify some things.  

The most annoying on might be complex vectors as I don't think Json supports complex numbers directly.  

@nightlark @trevorhardy @kdheepak @eberleim

trevorhardy:
I like the idea of queries always returns structured data (JSONs in this case). 

And regarding the complex numbers, it looks like Python has a pretty easy way to convert strings to complex values so we might be able to squeeze it in that way.

nightlark:
Queries returning JSON seems better than coming up with our own format, and there are standard libraries in other languages that make working with JSON easy -- though we should aim to make it conform to  the JSON standard (RFC 7159) to avoid issues with non-standard additions like comments.

What would string conversions to something else affect? Something like a `complexFromString` function that takes a string as input?

phlptp:
I am guessing the changes would be 
-  queries returning single strings need to be quoted
-  queries returning string vectors will be modified to meet json standard
-  vector to string would likely be modified slightly
- complex to string ?
- complex vector to string? 

The last two might require some thought as there are likely a couple different ways that could be done.  

As far as supporting comments the only place I think we really need it is in the config file support

nightlark:
Okay. Having the returned string be something (JSON-friendly?) that can be turned into valid JSON by adding `[]` or quotes around it seems more natural than needing to remove them.

kdheepak:
I'm supportive of having json be returned as part of the queries. Right now custom parsing is required to deal with this. If we are returning structured data, we might as well use a standard format.

PR #1452, opened 2020-07-06
make the vector/complex number conversions to fit JSON standard.

phlptp:
### Summary

If merged this pull request will use JSON strings as the vector and complex vector conversion operations.  

### Proposed changes

- update vector to string operations
- update complex vector to string operations
- update named point to string operations
- update complex number to string operation now `[real, imag]` instead of using the i/j notation, though the ability to read that notation remains in place.

### DIFF
--- a/ThirdParty/frozen/bits/pmh.h
+++ b/ThirdParty/frozen/bits/pmh.h
@@ -96,13 +96,15 @@ pmh_buckets<M> constexpr make_pmh_buckets(const carray<Item, N> & items,
   result_t result{};
   // Continue until all items are placed without exceeding bucket_max
   while (1) {
-    for (auto & b : result.buckets) {
-      b.clear();
-    }
     result.seed = prg();
     for (std::size_t i = 0; i < N; ++i) {
       auto & bucket = result.buckets[hash(key(items[i]), static_cast<size_t>(result.seed)) % M];
-      if (bucket.size() >= result_t::bucket_max) { continue; }
+      if (bucket.size() >= result_t::bucket_max) {
+        continue;
+        for (auto & b : result.buckets) {
+          b.clear();
+        }
+      }
       bucket.push_back(i);
     }
     return result;
--- a/src/helics/application_api/helicsTypes.cpp
+++ b/src/helics/application_api/helicsTypes.cpp
@@ -104,7 +104,7 @@ std::string helicsComplexString(std::complex<double> val)
     return helicsComplexString(val.real(), val.imag());
 }
 /** map of an assortment of type string that can be converted to a known type*/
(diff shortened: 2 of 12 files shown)


phlptp:
Address issue #1368

ThirdParty/frozen/bits/pmh.h
```diff
@@ -96,13 +96,15 @@ pmh_buckets<M> constexpr make_pmh_buckets(const carray<Item, N> & items,
...
-    }
     result.seed = prg();
     for (std::size_t i = 0; i < N; ++i) {
       auto & bucket = result.buckets[hash(key(items[i]), static_cast<size_t>(result.seed)) % M];
-      if (bucket.size() >= result_t::bucket_max) { continue; }
+      if (bucket.size() >= result_t::bucket_max) {
+        continue;
+        for (auto & b : result.buckets) {
+          b.clear();
+        }
```
nightlark, line 106: Isn't the `continue` going to skip the for loop? It also looks like if this is here than clearing all of the buckets shouldn't be happening multiple times.

src/helics/application_api/helicsTypes.cpp
```diff
@@ -178,20 +163,67 @@ static const std::unordered_map<std::string, data_type> typeMap{
     {"", data_type::helics_any},
     {"all", data_type::helics_any}};
 
-data_type getTypeFromString(const std::string& typeName)
+static const std::unordered_map<std::string, data_type> demangle_names{
```
nightlark, line 166: The manged/demangled names are only used within a federate and never get sent to another federate, right?

phlptp: The only place this is used, is when a user enters a type, via a string, This should detect if it is one of the primary types.   I should double check that it actually maps it to a common string though.  If not that might be something we should do in HELICS 3.

ThirdParty/frozen/bits/pmh.h
phlptp: Hmm,  that would be a bug in the frozen library.   I will investigate that further.

src/helics/application_api/helicsTypes.cpp
phlptp: It is definitely possible to change the string as it comes in to some stable type, but I am a little hesitant to do that right now, since it could mean that if the user queries the type they put in, they would get something else back.    I am not sure we want that.

phlptp: I added a clean function for the types, but it will only change them if they are matched through the demangle map,  if they match in the static map, they will stay the same.

phlptp: It will do this for input and publications

Issue #1445, opened 2020-06-25
CMAKE error while installing master in Linux


priyatm62:
- **I'm submitting a ...**
  - [ x] bug

### Bug

1. **What is the current behavior?**
While installing, cmake and make are successful. During make install, the following error comes (username replaced) -
CMAKE error: File install cannot find 
"/home/<user>/HelicsInstall/lib/libhelics-shared.so.2.5.2".

When I browse to the installation location, there is a file with the following name in Camel case and not dashes,
/home/<user>/HelicsInstall/lib/libhelicsSharedLib.so.2.5.2

1. **What is the expected behavior? What is the motivation / use case for changing the behavior?**

1. **What are the steps to reproduce this bug? Please provide a minimal working example of the bug if possible.**
 mkdir ${HELICS_INSTALL}
 cd ${HELICS_INSTALL}
 cmake {HELICS_SOURCE} -DCMAKE_INSTALL_PREFIX=${HELICS_INSTALL} -DBUILD_PYTHON_INTERFACE=ON -DCMAKE_BUILD_TYPE=Release -DHELICS_BUILD_CXX_SHARED_LIB=ON -DHELICS_BUILD_TESTS=OFF -DHELICS_BUILD_EXAMPLES=OFF
make -j 4
make install

1. **Please tell us about your environment:**
Helics master cloned on 06/24

- Operating System: [ Linux]
- Language Extension: [ Python ]

- **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, gitter, etc)

priyatm62:
Same error in develop as well

kdheepak:
Can you try it without this flag `-DHELICS_BUILD_CXX_SHARED_LIB=ON`?

priyatm62:
I am getting this error during make install - 
CMake Error at src/helics/apps/cmake_install.cmake:47
file INSTALL cannot find "/home/<username>/HelicsInstall/bin/helics_player"

And that binary is not there in bin. There is helics-config, helics_app, helics_broker, helics_broker_server and helics_recorder.

I am using the latest develop branch

phlptp:
what platform are you using.  Specific linux variant, compiler, etc?  We do have some tests that pass that do this so I think it is something platform specific, but I am confused as to what might be going on.

nightlark:
[...]

PR #1459, opened 2020-07-07
CMake error if user sets the install prefix to the build folder

nightlark:
### Summary

If merged this pull request will make CMake exit with an error if the install prefix is set to the build folder. Setting install prefix to the build folder is the likely cause of #1445, due to build files conflicting with install files when the install path is set to the build folder.

### Proposed changes

- Exit CMake configure with an error if a user sets the install prefix to the build folder

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -100,8 +100,7 @@ if(CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT AND CMAKE_PROJECT_NAME STREQUAL P
     endif(WIN32)
 endif()
 
-# Check to make sure the install prefix isn't the build folder, if it is, build errors will
-# happen
+# Check to make sure the install prefix isn't the build folder, if it is, build errors will happen
 get_filename_component(tmp_install_prefix ${CMAKE_INSTALL_PREFIX} REALPATH)
 get_filename_component(tmp_proj_bindir ${PROJECT_BINARY_DIR} REALPATH)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -100,6 +100,19 @@ if(CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT AND CMAKE_PROJECT_NAME STREQUAL P
...
+get_filename_component(tmp_proj_bindir ${PROJECT_BINARY_DIR} REALPATH)
+# Windows paths are case insensitive
+if(WIN32)
+    string(TOLOWER "${tmp_install_prefix}" tmp_install_prefix)
+    string(TOLOWER "${tmp_proj_bindir}" tmp_proj_bindir)
+endif()
+if(tmp_install_prefix STREQUAL tmp_proj_bindir)
+    message(FATAL_ERROR "CMAKE_INSTALL_PREFIX must not be set to the build folder")
+endif()
+
```
phlptp, line 114: The `REALPATH` here is to resolve, symbolic links and relative paths in either variable?

nightlark: Yes, leaving a symlink or relative path in one but not the other could make it look like two paths are different when they are the same.

Hopefully this catches most of the cases. macOS is a special, since it can have case-sensitive or case-insensitive depending the filesystem the drive was formatted with; I didn't find a way to detect what the drive was formatted with in CMake.

PR #1461, opened 2020-07-09
Benchmark build workflow

nightlark:
### Summary

If merged this pull request will add a workflow to build a package with the benchmarks for Windows/macOS/Linux that can be triggered by an API call (or manually once the workflow is in the master branch). Once it makes its way into master it will also run every 2nd day of the week (3/4 times/week, something like Sunday, Tuesday, Thursday, Saturday or Monday, Wednesday, Friday).

### Proposed changes

- Add a workflow for building packages with the benchmark binaries

### DIFF
--- a/.github/actions/benchmark-build/benchmark-Windows.sh
+++ b/.github/actions/benchmark-build/benchmark-Windows.sh
@@ -1,0 +1,17 @@
+#!/bin/bash
+# This uses bash variable substitution in a few places
+# 1. replacing x86 with Win32 (setting the Python version uses x86)
+# 2. getting the cmake directory for running cpack with an absolute path (chocolatey has an unfortunately named alias)
+
+echo "Building ${CPACK_GEN} installer with ${BUILD_GEN} for ${BUILD_ARCH}"
+choco install -y swig
+BOOST_ROOT="${BOOST_ROOT_1_72_0}"
+export BOOST_ROOT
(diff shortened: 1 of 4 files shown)


.github/actions/benchmark-build/benchmark-Windows.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+# This uses bash variable substitution in a few places
+# 1. replacing x86 with Win32 (setting the Python version uses x86)
+# 2. getting the cmake directory for running cpack with an absolute path (chocolatey has an unfortunately named alias)
+
+echo "Building ${CPACK_GEN} installer with ${BUILD_GEN} for ${BUILD_ARCH}"
+choco install -y swig
```
phlptp, line 7: why do we need to install `swig` for this?

.github/actions/benchmark-build/benchmark-macOS.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+# This uses bash variable substitution in a few places
+# 1. getting the cmake directory for running cpack with an absolute path (chocolatey has an unfortunately named alias)
+
+brew install swig boost
```
phlptp, line 5: Can probably also remove swig from here as well, unless this is used for other things as well.

PR #1464, opened 2020-07-13
Helics3 converters

phlptp:
### Summary

If merged this pull request will add serializers and deserializers for the primary helics data types

### Proposed changes

- to/from binary converters for common data types 
- tests of the converters

### DIFF
--- a/src/helics/application_api/ValueConverter.cpp
+++ b/src/helics/application_api/ValueConverter.cpp
@@ -29,22 +29,92 @@ template class ValueConverter<std::vector<std::complex<double>>>;
 template class ValueConverter<NamedPoint>;
 
 namespace detail {
-    size_t getBinaryLength(double val) { return 9; }
-    static constexpr unsigned char doubleCode = 0xB0;
+    namespace checks {
+        /*! This code in checks namespace exerpted from cereal portable binary archives
+        code is modified slightly to fit name conventions and make use of C++17
 
-    static constexpr unsigned char intCode = 0x50;
-    static constexpr unsigned char complexCode = 0x12;
-    static constexpr unsigned char stringCode = 0x0E;
-    static constexpr unsigned char vectorCode = 0x6C;
-    static constexpr unsigned char npCode = 0xAE;
-    static constexpr unsigned char cvCode = 0x62;
-    static constexpr unsigned char customCode = 0xF4;
+        \file binary.hpp
+    \brief Binary input and output archives */
+        /*
+          Copyright (c) 2014, Randolph Voorhies, Shane Grant
+          All rights reserved.
 
-    void convertToBinary(char* data, double val)
+          Redistribution and use in source and binary forms, with or without
+          modification, are permitted provided that the following conditions are met:
+              * Redistributions of source code must retain the above copyright
+                notice, this list of conditions and the following disclaimer.
+              * Redistributions in binary form must reproduce the above copyright
+                notice, this list of conditions and the following disclaimer in the
+                documentation and/or other materials provided with the distribution.
+              * Neither the name of cereal nor the
+                names of its contributors may be used to endorse or promote products
+                derived from this software without specific prior written permission.
+
+          THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+          ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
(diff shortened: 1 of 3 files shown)


phlptp:
@nightlark  any idea why the formatting PR generation is failing.  I am not clear on what the error means.

nightlark:
The formatting doesn't seem to be showing anything needs fixing?

nightlark:
It seems like sometimes the GitHub API requests can be flaky for unknown reasons. Happened last week where one commit failed to open a PR, but the next one did.

phlptp:
This is another step along the way.  This just add the converters and tests.  The next step will be to actually use them with the converter templates.  

I also would like some way to test the endianness conversion, but not sure what the best way to doing that would be.

phlptp:
I fixed the formatting locally, but some of the previous commits never generated a PR when the formatting was incorrect.

phlptp:
I think I am going to try to use boost::endian library for some additional tests in the next PR on this sequence to verify the operation of the endian conversions.  

Then the one after that will try to use the new serialization for actual data transfer.

src/helics/application_api/ValueConverter.cpp
```diff
@@ -29,22 +29,320 @@ template class ValueConverter<std::vector<std::complex<double>>>;
...
+                    (static_cast<size_t>(data[2]) << 8U) + static_cast<size_t>(data[3]);
+            case data_type::helics_named_point:
+                return (static_cast<size_t>(data[9]) << 16U) +
+                    (static_cast<size_t>(data[10]) << 8U) + static_cast<size_t>(data[11]);
+        }
+    }
+
+    data_type detectType(const std::byte* data, size_t size)
+    {
+        switch (size) {
```
nightlark, line 219: Is there a reason for putting the code mask at the end? Always having the first byte be the code byte would make it easier to detect the type and could enable packing different value types into a single data blob.

phlptp: I think the biggest consideration is keeping the serialization of double and int64 to <=15 bytes, so it fits in the short string optimization size of all platforms and there are no additional allocations for those types.  Additionally complex <=22 bytes.  As that I know is a performance hit if it doesn't fit in a SSO for the std::string.  
(maybe we want to consider a custom container with different Small size optimization?)  

The string pointer is always aligned (that is not the only use case but a very common one)  so having aligned access to copy makes it a single instruction to serialize that part of it.  

I am not sure what the cost of unaligned access is.

nightlark: Actually, the opposite order (footer instead of header) could get rid of the need to add padding by having the data elements first, followed by 4/8 bytes size if needed, and the type of data/endianness at the end. If combining multiple fields into a single packet, null bytes (null type code) could optionally be added as padding to align copies for the subsequent fields, with the downside being that deserializing fields in a class/struct would need to happen in reverse.

I haven't found much on the performance of memcpy for aligned/unaligned on x86. It's mostly embedded systems stuff that appears (https://www.embedded.com/optimizing-memcpy-improves-speed/) on ARM, which probably behaves differently from x86.

phlptp:
We get this error on the formatting PR
```
[pre-commit/update-helics3_converters cde200b] Automated formatting of repo files
 1 file changed, 9 insertions(+), 12 deletions(-)
To https://github.com/GMLC-TDC/HELICS
 ! [remote rejected] pre-commit/update-helics3_converters -> pre-commit/update-helics3_converters (shallow update not allowed)
error: failed to push some refs to '***github.com/GMLC-TDC/HELICS'
```

@nightlark  any ideas?

phlptp:
I can fix the formatting easy enough but I am wondering why the PR generation is failing

nightlark:
I think the problem is that the checkout action is doing a shallow clone now instead of fetching all history.

phlptp:
how do we fix that?

nightlark:
I have a PR that does a full clone that could be rebased after this is merged.

nightlark:
Actually, I'll just rebase it now.

phlptp:
I think I converted all the serializers to have a consistent 8 byte header.  4 bytes of which are just 0 at the moment so we have some flexibility for the future if needed.

nightlark:
The size should probably shift to bytes 4-7 and leave 3 bytes for extra header stuff after the byte code. I don't think we should be artificially limiting the maximum data size to ~8mb instead of the maximum size on 32-bit systems, if someone wants to try to send more. Someone is bound to try sending a large array and wonder why only part of it is getting sent. Little-endian CPUs are common enough that the size could probably just be copied as is and bytes swapped based on the endian bit in the header in the rare case of a big endian federate.

PR #1488, opened 2020-07-20
Small buffer

phlptp:
### Summary

If merged this pull request will add a SmallBuffer implementation to the Helics/common library

### Proposed changes

- SmallBuffer Implementation
- tests for Small Buffer

NOTE:  this has not been integrated into usage, that will be the next PR

### DIFF
--- a/src/helics/common/SmallBuffer.cpp
+++ b/src/helics/common/SmallBuffer.cpp
@@ -55,7 +55,7 @@ SmallBuffer::SmallBuffer(const void* data, size_t size): heap(buffer.data())
 
 SmallBuffer& SmallBuffer::operator=(const SmallBuffer& sb)
 {
-    if (this==&sb) {
+    if (this == &sb) {
         return *this;
     }
     resize(sb.size());
@@ -65,7 +65,7 @@ SmallBuffer& SmallBuffer::operator=(const SmallBuffer& sb)
 
 SmallBuffer& SmallBuffer::operator=(std::string_view val)
 {
-    if (reinterpret_cast<const std::byte *>(val.data())==heap) {
+    if (reinterpret_cast<const std::byte*>(val.data()) == heap) {
         bufferSize = val.size();
         return *this;
     }
@@ -90,7 +90,6 @@ SmallBuffer& SmallBuffer::operator=(SmallBuffer&& sb) noexcept
             }
             delete[] heap;
         }
-        
     }
     if (sb.usingAllocatedBuffer) {
         heap = sb.heap;
@@ -112,7 +111,8 @@ SmallBuffer& SmallBuffer::operator=(SmallBuffer&& sb) noexcept
     return *this;
 }
 
-std::byte* SmallBuffer::release() {
+std::byte* SmallBuffer::release()
+{
     if (!usingAllocatedBuffer) {
         return nullptr;
     }
@@ -129,7 +129,7 @@ void SmallBuffer::moveAssign(void* data, std::size_t size, std::size_t capacity)
 {
     auto* newHeap = reinterpret_cast<std::byte*>(data);
     if (usingAllocatedBuffer && !nonOwning) {
-        if (newHeap!=heap) {
+        if (newHeap != heap) {
             delete[] heap;
         }
     }
@@ -145,7 +145,7 @@ void SmallBuffer::spanAssign(void* data, std::size_t size, std::size_t capacity)
     auto* newHeap = reinterpret_cast<std::byte*>(data);
     if (usingAllocatedBuffer && !nonOwning) {
         if (newHeap == heap) {
-            //if the heaps are the same the only thing to change is the size and capacity
+            // if the heaps are the same the only thing to change is the size and capacity
             // don't change the other characteristics
             bufferSize = size;
             bufferCapacity = capacity;
@@ -162,6 +162,9 @@ void SmallBuffer::spanAssign(void* data, std::size_t size, std::size_t capacity)
 
 void SmallBuffer::assign(const void* start, const void* end)
 {
+    if (start > end) {
+        throw(std::invalid_argument("invalid range specified, end pointer before start pointer"));
+    }
     const auto* st1 = reinterpret_cast<const std::byte*>(start);
     const auto* end1 = reinterpret_cast<const std::byte*>(end);
     resize(end1 - st1);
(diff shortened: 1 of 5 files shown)


phlptp:
@nightlark  I think this is ready for review.  
@eberleim Just wanted to start getting you involved in some core library code reviews.

src/helics/common/SmallBuffer.cpp
```diff
@@ -0,0 +1,260 @@
...
+        delete[] heap;
+    }
+    heap = newHeap;
+    bufferCapacity = capacity;
+    bufferSize = size;
+    nonOwning = true;
+    usingAllocatedBuffer = true;
+}
+
+void SmallBuffer::assign(const void* start, const void* end)
```
nightlark, line 163: Does a check to make sure end >= start and throw an exception need to get added? It looks like std::vector's assign method throws a bad_alloc exception in that scenario, but I don't know if that's something in their assign method implementation, or elsewhere in the standard library that results in that behavior.

phlptp: I added the check, end==start isn't a problem but end < start could be.  So I am throwing a std::invalid_argument in that case.

nightlark: So end>=start is okay, with end==start being the "empty" case.

tests/helics/common/SmallBufferTests.cpp
```diff
@@ -0,0 +1,478 @@
...
+}
+
+TEST(small_buffer_tests, assign)
+{
+    SmallBuffer sb1;
+    std::string t1(3634, 'g');
+    sb1.assign(t1.data(), t1.data() + t1.size());
+    EXPECT_EQ(t1.size(), sb1.size());
+    EXPECT_EQ(sb1[3333], std::byte{'g'});
+}
```
eberleim, line 351: What happens with an assignment of size zero? 
```
    sb1.assign(t1.data(), t1.data());
    EXPECT_EQ(sb1.size(), 0);
```
Would be a valid test here, but I do not believe it would be a meaningful buffer, and I believe attempts to access the buffer object would result in a segfault.

src/helics/common/SmallBuffer.hpp
```diff
@@ -0,0 +1,93 @@
...
+    SmallBuffer(std::string_view val);
+    SmallBuffer(const void* data, size_t size);
+    ~SmallBuffer();
+    SmallBuffer& operator=(const SmallBuffer& sb);
+    SmallBuffer& operator=(SmallBuffer&& sb) noexcept;
+    SmallBuffer& operator=(std::string_view val);
+    /** return a pointer to the data location*/
+    std::byte* data() const { return heap; }
+    /** get the start of the data*/
+    std::byte* begin() { return heap; }
```
eberleim, line 31: `begin()` and `end()` may be more function if return type is `std::iterator<std::input_iterator_tag, std::Byte>` rather than a direct pointer, as this will enable range-based iteration across SmallBuffer.

src/helics/common/SmallBuffer.cpp
```diff
@@ -0,0 +1,267 @@
...
+        heap = sb2.heap;
+        bufferCapacity = sb2.bufferCapacity;
+        usingAllocatedBuffer = true;
+        nonOwning = sb2.nonOwning;
+        sb2.usingAllocatedBuffer = false;
+        sb2.nonOwning = false;
+        sb2.heap = buffer.data();
+        sb2.bufferCapacity = 64;
+
+        memcpy(sb2.heap, buffer.data(), bufferSize);
```
eberleim, line 231:
```suggestion
        std::memcpy(sb2.heap, buffer.data(), bufferSize);
```

src/helics/common/SmallBuffer.cpp
```diff
@@ -0,0 +1,267 @@
...
+        sb2.heap = heap;
+        sb2.bufferCapacity = bufferCapacity;
+        sb2.usingAllocatedBuffer = true;
+        sb2.nonOwning = nonOwning;
+        usingAllocatedBuffer = false;
+        nonOwning = false;
+        heap = buffer.data();
+        bufferCapacity = 64;
+
+        memcpy(heap, sb2.buffer.data(), sb2.size());
```
eberleim, line 219:
```suggestion
         std::memcpy(heap, sb2.buffer.data(), sb2.size());
```

src/helics/common/SmallBuffer.hpp
```diff
@@ -0,0 +1,93 @@
...
+    /** end iterator*/
+    std::byte* end() { return heap + bufferSize; }
+    /** get a const iterator*/
+    const std::byte* begin() const { return heap; }
+    /** get a const end iterator*/
+    const std::byte* end() const { return heap + bufferSize; }
+    /** get an element*/
+    std::byte operator[](size_t index) const { return heap[index]; }
```
eberleim, line 39: As mentioned elsewhere in this review, this operator does no validity checks, allowing unbounded access to elements outside the array. While it adds overhead, adding a safe array accessor would be preferable for safety.

src/helics/common/SmallBuffer.cpp
eberleim: In "empty" case described by @nightlark is there a desired behavior if the `operator[]` is used to access a member of the size-zero heap? At present, it is a segfault.

src/helics/common/SmallBuffer.hpp
phlptp: I think I would defer to std::vector here.  I added an `at` function which does the bounds checking and will throw a std::out_of_range exception.  Where as [] does no checking.

tests/helics/common/SmallBufferTests.cpp
phlptp: as far as I can this test works fine.  If done initially the contained buffer is active,  otherwise it doesn't change the existing buffer.  So unless someone passed in a void buffer this will always result in a valid buffer and not produce a seg fault.

src/helics/common/SmallBuffer.hpp
phlptp: I am pretty sure pointers are considered iterators for range based for loops.  and some of the tests use a range based for loop in the test.    So my inclination is to leave this as a pointer since then begin and end can be used in the assign of another Small buffer, and I don't want to make the assign a template right now.

src/helics/common/SmallBuffer.cpp
phlptp: I don't think it would be a segfault.  at size 0 it would still be using the internal buffer so would have a minimum capacity of 64, which means anything 0-63 and likely a few more wouldn't be  segfault

tests/helics/common/SmallBufferTests.cpp
eberleim: You are correct, I was attempting to access the buffer index in an invalid way, and caused the segfault myself.

PR #1494, opened 2020-07-22
Spdlog

phlptp:
### Summary

If merged this pull request will integrate spdlog and switch over logging services to spdlog

### Proposed changes

- include spdlog
- update fmt
- use spdlog for logging
- remove old logger and usages


phlptp:
@nightlark  I think this is ready to go

src/helics/apps/Recorder.cpp
```diff
@@ -24,12 +23,22 @@ SPDX-License-Identifier: BSD-3-Clause
...
 
+/** encode the string in base64 if needed otherwise just return the string*/
+static std::string encode(const std::string& str2encode)
+{
+    return std::string("b64[") +
+        gmlc::utilities::base64_encode(reinterpret_cast<const unsigned char*>(str2encode.c_str()),
+                                       static_cast<int>(str2encode.size())) +
+        ']';
+}
+
```
nightlark, line 41: This function seems like it should eventually get moved to `gmlc/utilities/base64.h`

src/helics/apps/Recorder.hpp
```diff
@@ -54,9 +54,9 @@ namespace apps {
         Recorder(const std::string& name, CoreApp& core, const FederateInfo& fi);
         /**constructor taking a file with the required information
     @param name the name of the app
-    @param file a file defining the federate information in JSon or text
+    @param jsonString a file or jsonString defining the federate information in JSON
```
nightlark, line 57:
```suggestion
    @param jsonString a file or JSON string defining the federate information in JSON
```

src/helics/apps/Recorder.cpp
phlptp: probably, though I am not sure b64[ is fully standard,  I should look into that more.

nightlark: There isn't really a standard for marking base64 data in JSON, other than just using the base64 encoded data and knowing that the field is base64 encoded.. which it looks like we have an encoding field already that gets set `message["encoding"] = "base64";` that could get used.

PR #1504, opened 2020-07-31
Benchmark scripts

nightlark:
### Summary

If merged this pull request will add the latest version of the scripts used to run benchmarks on clusters using slurm.


phlptp:
Are these the scripts you used on the pi cluster?

phlptp:
Is there any docs on how to use these scripts?

PR #1506, opened 2020-08-01
Bug fixes

phlptp:
### Summary

If merged this pull request will fix a bug with negative numerical strings. 
Add a few missing calls in the C++98 API
Update fmt lib

### DIFF
--- a/src/helics/application_api/helicsTypes.cpp
+++ b/src/helics/application_api/helicsTypes.cpp
@@ -345,14 +345,14 @@ NamedPoint helicsGetNamedPoint(const std::string& val)
     if (loc == std::string::npos) {
         auto fb = val.find_first_of('[');
         if (fb != std::string::npos) {
-            return {val, std::nan("0")};
+            return {val, getDoubleFromString(val)};
         }
         auto V = helicsGetComplex(val);
         if (V.real() <= invalidDouble) {
             return {val, std::nan("0")};
         }
         if (V.imag() == 0) {
-            return {"value", std::abs(V)};
+            return {"value", V.real()};
         }
         return {val, V.real()};
     }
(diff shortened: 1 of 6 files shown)


src/helics/application_api/helicsTypes.cpp
```diff
@@ -345,14 +345,14 @@ NamedPoint helicsGetNamedPoint(const std::string& val)
...
         }
         auto V = helicsGetComplex(val);
         if (V.real() <= invalidDouble) {
             return {val, std::nan("0")};
         }
         if (V.imag() == 0) {
-            return {"value", std::abs(V)};
+            return {"value", V.real()};
         }
         return {val, V.real()};
```
nightlark, line 357: Why is it replacing the name with "value"? It's otherwise returning something that should be identical to the return if `V.imag() != 0`.

phlptp: I have debated what to do here.  A NamedPoint is a string and a double.   If you just send a numerical string,  what do you put in the string?   Right now if there is no loss of information, then I just put "value",  if there is a loss of information I put the original string.  I went back and forth on whether to put "value" there as well and just have it be a truncation for consistency.  

Open to suggestions on something better or which policy we should adopt.

nightlark: I think the conclusion I've reached is that it seems weird, because converting from a complex number to a double is losing information, and the conversion to a NamedPoint probably isn't something that they should be doing when working with complex numbers; either way won't fit every use case, so they should probably be handling the real and imaginary parts of a complex number on their own based on what their use case.

phlptp: I agree this would be an odd conversion.

Issue #1462, opened 2020-07-09
ValueFederate::registeValueInterfacesJson() get publication units


xcosmos6:
Hi, 
I am using the helics cpp98 and calling the helicscpp::ValueFederate(const std::string& configString) constructor to get an instance of the valueFedeate object. In my json config file, I have publications JSON array and I have "unit" for each of the pub elements. However, after the config file is parsed, all the units are empty string. After I step through the code, I found that function ValueFederate::registeValueInterfacesJson() is called, which is in the application_api:
`auto units = getOrDefault(sub, "units", emptyStr);`
Here "units" is used instead of "unit". I am not sure if this is a bug or intentional, but after I change the key in my config file to "units", the correct units show up instead of empty string.
Thank you very much.

phlptp:
json now supports both "units" and "unit" as a specify for including units

PR #1512, opened 2020-08-04
Unit eq units

phlptp:
### Summary

If merged this pull request will allow the use of "unit"  in addition to "units"  and fix the fact that most config files were incorrect with regard to the units since they had "unit"


phlptp:
Address Issue #1462

phlptp:
previously it matched "units"  so I think that needs to the priority (in case both are given for some reason),  
If we want to make some changes in helics3 then I think we can do that.  but given there were examples with "unit"  and "units" I think both need to be supported for now.    I am not sure which I prefer.

PR #1513, opened 2020-08-05
add dumplog, and force_logging_flush flag options for the federate.

phlptp:
### Summary

If merged this pull request will add flagOption for specifying dumplog and force_logging_flush to capture all actionMessages and force a logging file to flush every message.


tests/helics/application_api/LoggingTests.cpp
```diff
@@ -257,3 +257,32 @@ TEST(logging_tests, check_log_message_levels_high)
...
+                           [&mlog](int level,
+                                   const std::string& /*unused*/,
+                                   const std::string& message) {
+                               mlog.lock()->emplace_back(level, message);
+                           });
+
+    Fed->enterExecutingMode();
+
+    Fed->setFlagOption(helics_flag_dumplog);
+    Fed->setFlagOption(helics_flag_dumplog, false);
```
nightlark, line 280: What is setting the flag twice testing?

phlptp: It is turning it on and off during the run, so I only get a limited number of messages in the log file.  There is one that depending on the timing of the different threads could come in between the two but in most cases the log callback will only have two messages.  The initial log message and the dumplog message.  If it has fewer the dump log didn't work at all, if it has more than 3 the turn off didn't work.

nightlark: Okay. Should probably have a comment mentioning that reason for turning it off right after turning it on.

PR #1531, opened 2020-08-14
Setup CI jobs on Azure pipelines

nightlark:
### Summary

Setup commit and some daily jobs on Azure Pipelines

If merged this pull request will add Linux and macOS jobs to runs on Azure Pipelines, and some scheduled daily jobs for checking (possibly less common) variants.

### Proposed changes

- Add Linux and macOS jobs to Azure Pipelines
- Make some aspects of the CI build/test scripts use-able on services other than Travis CI

### DIFF
--- a/.ci/azure-pipelines-daily.yml
+++ b/.ci/azure-pipelines-daily.yml
@@ -1,9 +1,9 @@
 schedules:
-- cron: "0 10 * * *"
-  displayName: Daily late night PST (GMT-8)
-  branches:
-    include:
-      - helics3
+  - cron: '0 10 * * *'
+    displayName: Daily late night PST (GMT-8)
+    branches:
+      include:
+        - helics3
 
 jobs:
   - job: Daily Linux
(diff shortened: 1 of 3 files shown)


.ci/azure-pipelines-daily.yml
```diff
@@ -0,0 +1,41 @@
...
+    branches:
+      include:
+        - helics3
+
+jobs:
+  - job: Daily Linux
+    strategy:
+      matrix:
+        all-tests:
+          containerImage: 'helics/buildenv:ubuntu18.04-default-builder'
```
phlptp, line 13: what compiler does the default builder use?

.ci/azure-pipelines-daily.yml
```diff
@@ -0,0 +1,41 @@
...
+    timeoutInMinutes: 90
+
+    steps:
+      - bash: |
+          source scripts/setup-helics-ci-options.sh
+          mkdir -p build && cd build
+          ../scripts/ci-build.sh
+        env:
+          MAKEFLAGS: '-j 4'
+          CMAKE_GENERATOR: 'Unix Makefiles'
```
phlptp, line 31: should we have a test somewhere with the ninja generator?  Not that this needs to be.

nightlark: The builds on Drone CI use Ninja, but being able to have some of the tests use different generators in the future was one of the reasons for selecting the generator using an environment variable.

.ci/azure-pipelines-daily.yml
nightlark: gcc7 - this is the build using the default version of all Ubuntu 18.04 packages.

Issue #1526, opened 2020-08-13
Default cpp98 Message Constructor creates internal null helics_message_object


afisher1:
**Describe the bug**
The Default Message class constructor in the cpp98/Endpoint.hpp creates the class with the internal helics_message_object being set with a null pointer. This causes calls to set the attributes of the internal helics_message_object to fail such as data(const string &string), returning with an error stating the message object is not valid. Currently the only way to use the Message class fully is by first creating a helics_message_object using C API and passing that to the constructor. I suggest adding a couple constructors that take a helicscpp::Federate or a helicscpp::Endpoint such that under the hood helicsFederateCreateMessageObject() and helicsEndpointCreateMessageObject() can be used to initialize Message class's helics_message_object member.

PR #1533, opened 2020-08-14
add some Message constructors and tests in the C++98 API

phlptp:
### Summary

If merged this pull request will add some creation functions and tests for the Message object in the C++98 API.

### DIFF
--- a/src/helics/cpp98/Endpoint.hpp
+++ b/src/helics/cpp98/Endpoint.hpp
@@ -15,12 +15,17 @@ SPDX-License-Identifier: BSD-3-Clause
 #include <vector>
 
 namespace helicscpp {
+class Endpoint;
+class Federate;
 
 class Message {
   public:
     /** default constructor*/
     Message() HELICS_NOTHROW: mo(HELICS_NULL_POINTER) {}
-
+    /** create a message associated with a federate*/
+    explicit Message(const Federate& fed);
+    /** create a message associated with a federate*/
+    explicit Message(const Endpoint& ept);
     /** construct from a helics_message object*/
     explicit Message(helics_message_object hmo) HELICS_NOTHROW: mo(hmo) {}
 
@@ -96,7 +101,7 @@ class Message {
         helicsMessageSetOriginalSource(mo, osrc.c_str(), hThrowOnError());
         return *this;
     }
-    /** get the originali message destination if a filtered altered it*/
+    /** get the original message destination if a filtered altered it*/
     const char* originalDestination() const { return helicsMessageGetOriginalDestination(mo); }
     /** set the original destination field*/
     Message& originalDestination(const std::string& odest)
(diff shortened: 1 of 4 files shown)


src/helics/cpp98/Endpoint.hpp
```diff
@@ -96,7 +101,7 @@ class Message {
         helicsMessageSetOriginalSource(mo, osrc.c_str(), hThrowOnError());
         return *this;
     }
-    /** get the originali message destination if a filtered altered it*/
+    /** get the original message destination if a filtered altered it*/
```
nightlark, line 104:
```suggestion
    /** get the original message destination if a filter altered it*/
```

src/helics/cpp98/Endpoint.hpp
```diff
@@ -15,12 +15,17 @@ SPDX-License-Identifier: BSD-3-Clause
...
+class Federate;
 
 class Message {
   public:
     /** default constructor*/
     Message() HELICS_NOTHROW: mo(HELICS_NULL_POINTER) {}
-
+    /** create a message associated with a federate*/
+    explicit Message(const Federate& fed);
+    /** create a message associated with a federate*/
```
nightlark, line 27:
```suggestion
    /** create a message associated with an endpoint*/
```

src/helics/cpp98/Endpoint.hpp
```diff
@@ -177,7 +182,11 @@ class Message {
         mo = HELICS_NULL_POINTER;
         return mreturn;
     }
+    /** generate a new message in a federate*/
+    Message& newMessageObject(const Federate& fed);
 
+     /** generate a new message in a federate*/
```
nightlark, line 188:
```suggestion
     /** generate a new message in an endpoint*/
```

src/helics/cpp98/MessageFederate.hpp
```diff
@@ -110,5 +110,21 @@ class MessageFederate: public virtual Federate {
   private:
     std::vector<helics_endpoint> local_endpoints;
 };
+//this code needs the definition of federate before it can de defined
+inline Message::Message(const Federate& fed):
```
nightlark, line 114: Would it make sense to also have a version that takes a `Federate*` so users can pass a `fed` pointer instead of `*fed`?

phlptp: I thought about this, but Federate and Endpoint are basically wrappers around a void *,  so for the tests I wrap that in a shared_ptr so an overload that took a pointer wouldn't really help anything.  And in general Value Classes for these objects are preferred over pointers.  So I would lean towards not having a pointer overload.

phlptp:
Link #1526 @afisher1 when you get a chance take a look at this to see if the modifications are what you were thinking.

PR #1553, opened 2020-08-25
Update NEW_ROUTE command in MpiComms to work with SmallBuffer

nightlark:
### Summary

If merged this pull request will update handling of the NEW_ROUTE command in MpiComms to work with the SmallBuffer changes.

### Proposed changes

- Convert SmallBuffer payload to a string for NEW_ROUTE commands in MpiComms


nightlark:
There's something weird with the clang5 build where pytest is definitely installed, but it isn't finding it for running the tests.

PR #1557, opened 2020-08-26
Handle empty install prefix without an error

nightlark:
### Summary

If merged this pull request will quote the variables used as part of the `install prefix == build dir` to handle an empty install prefix without triggering an error.

### Proposed changes

- Quote variables used in getting install prefix and build dir components to avoid errors if either happens to be empty

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -101,8 +101,8 @@ if(CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT AND CMAKE_PROJECT_NAME STREQUAL P
 endif()
 
 # Check to make sure the install prefix isn't the build folder, if it is, build errors will happen
-get_filename_component(tmp_install_prefix ${CMAKE_INSTALL_PREFIX} REALPATH)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -101,8 +101,8 @@ if(CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT AND CMAKE_PROJECT_NAME STREQUAL P
 endif()
 
 # Check to make sure the install prefix isn't the build folder, if it is, build errors will happen
-get_filename_component(tmp_install_prefix ${CMAKE_INSTALL_PREFIX} REALPATH)
-get_filename_component(tmp_proj_bindir ${PROJECT_BINARY_DIR} REALPATH)
+get_filename_component(tmp_install_prefix "${CMAKE_INSTALL_PREFIX}" REALPATH)
```
phlptp, line 104: Should we have a specific warning in this case?  Seems like in most cases if the install prefix is empty that represents some problem with the command line or something of that nature.

nightlark: I think that's true. In almost all cases its probably that they used an empty environment variable either from a typo or that they forgot to set.

PR #1569, opened 2020-09-14
Command interface

phlptp:
### Summary

If merged this pull request will add a command interface to HELICS.  A command interface is an asynchronous message passing interface.  command can be sent from anywhere and are not dependent on time synchronization inside helics.
Command message pass on the priority channel like queries.    This is a building block with limited established capabilities thus far.

future work will include additional capabilities.

### DIFF
--- a/docs/user-guide/commandInterface.md
+++ b/docs/user-guide/commandInterface.md
@@ -1,8 +1,7 @@
 # Commands
 
-The command interface for HELICS was introduced in HELICS 3.0.  It is an asynchronous communication mechanism to send commands or other information to other components.
+The command interface for HELICS was introduced in HELICS 3.0. It is an asynchronous communication mechanism to send commands or other information to other components.
 Cores and Brokers will respond to a small subset of commands known by HELICS, but federates have a command interface to allow retrieval of commands for interpretation by a federate.
-
 
 The general function appears like
 
@@ -10,7 +9,7 @@ The general function appears like
 void sendCommand(const std::string& target, const std::string& commandStr)
 ```
 
-the same function is available for federates, cores, and brokers.  
+the same function is available for federates, cores, and brokers.
 
 ## Targets
 
@@ -32,8 +31,7 @@ A target is specified, and can be one of the following. A federate named one of 
 
 ## Command String
 
-The `commandStr` is a generic string, so can be anything the can be contained in a string object.  It is expected that most commandStr will have a json format, though a few simple ones are just plain strings.
-
+The `commandStr` is a generic string, so can be anything the can be contained in a string object. It is expected that most commandStr will have a json format, though a few simple ones are just plain strings.
 
 ### HELICS supported commands
 
@@ -53,8 +51,8 @@ The following queries are defined for federates. Federates may specify a callbac
 ```
 
 ### Future
-How this will get used is somewhat up in the air yet.  It is expected that future commands to the objects will help support debugging and other diagnostics but beyond that it is expected to evolve considerably.
 
+How this will get used is somewhat up in the air yet. It is expected that future commands to the objects will help support debugging and other diagnostics but beyond that it is expected to evolve considerably.
 
 ## Usage Notes
 
@@ -63,7 +61,7 @@ Commands that must traverse the network travel along priority paths.
 ### Application API
 
 There are two basic calls in the application API as part of a [federate object](../doxygen/classhelics_1_1Federate.html)
-To retreive a command addressed to a federate there are two commands
+To retrieve a command addressed to a federate there are two commands
 
 ```cpp
 std::pair<std::string,std::string>     getCommand();
@@ -71,9 +69,10 @@ std::pair<std::string,std::string>     waitCommand();
 ```
 
 The first will return immediately but the strings will be empty if there is no command available.
-The second is a blocking call and will only return if a command is available.  
+The second is a blocking call and will only return if a command is available.
 
-Equivalent calls in the C API are 
+Equivalent calls in the C API are
+
 ```c
 const char *helicsFederateGetCommand(helics_federate fed, helics_error *err);
 const char *helicsFederateWaitCommand(helics_federate fed, helics_error *err);
--- a/docs/user-guide/commandInterface.md
+++ b/docs/user-guide/commandInterface.md
@@ -1,6 +1,6 @@
 # Commands
 
-The command interface for HELICS was introduced in HELICS 3.0. It is an asynchronous communication mechanism to send commands or other information to other components.
+The command interface for HELICS was introduced in HELICS 3. It is an asynchronous communication mechanism to send commands or other information to other components.
 Cores and Brokers will respond to a small subset of commands known by HELICS, but federates have a command interface to allow retrieval of commands for interpretation by a federate.
 
 The general function appears like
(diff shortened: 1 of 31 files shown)


phlptp:
@kdheepak  @nightlark can you take a look at the doc error in this,  not sure if it was something we changed or something we need to update for readthedocs

nightlark:
It might be a problem with syntax highlighting for the example code block in the documentation. What happens if the `cpp` language marking gets removed?

phlptp:
I think I fixed the doc issue in the PR #1572

src/helics/core/FederateState.cpp
```diff
@@ -1671,6 +1671,65 @@ void FederateState::logMessage(int level,
...
+}
+
+std::pair<std::string, std::string> FederateState::waitCommand()
+{
+    auto val = commandQueue.pop();
+    if (val.first == "notify") {
+        if (parent_ != nullptr) {
+            parent_->sendCommand(val.second, "notify_response", name);
+        }
+        val = commandQueue.pop();
```
nightlark, line 1728: Why are there multiple calls to pop? Shouldn't the first call already remove the value for the next command from the queue?

docs/user-guide/commandInterface.md
```diff
@@ -0,0 +1,82 @@
...
+
+```eval_rst
++------------------------------------------+---------------------------------------------------------------------------------------+
+| target                                   | Description                                                                           |
++==========================================+=======================================================================================+
+| ``broker``                               | The first broker encountered in the hierarchy from the caller                         |
++------------------------------------------+---------------------------------------------------------------------------------------+
+| ``root``, ``federation``                 | The root broker of the federation                                                     |
++------------------------------------------+---------------------------------------------------------------------------------------+
+| ``core``                                 | The core of a federation, this is not a valid target if called from a broker          |
```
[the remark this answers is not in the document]
phlptp: yes

src/helics/core/FederateState.cpp
phlptp: Notify is a command that shouldn't be sent to the federate,  but generates an echo like response to the source.  The command that gets sent back up to the federate is the next one, hence the two pops in that path.

docs/user-guide/commandInterface.md
```diff
@@ -0,0 +1,82 @@
...
```
phlptp, line 26:
```suggestion
| ``core``                                 | The core of a federate, this is not a valid target if called from a broker          |
```

phlptp:
rebased on helics3 branch

src/helics/application_api/Federate.cpp
```diff
@@ -1100,6 +1100,36 @@ void Federate::setGlobal(const std::string& valueName, const std::string& value)
     }
 }
 
+void Federate::sendCommand(const std::string& target, const std::string& commandStr)
+{
+    if (coreObject) {
+        coreObject->sendCommand(target, commandStr, getName());
+    } else {
+        throw(InvalidFunctionCall(
+            " command cannot be called on uninitialized federate or after disconnect call"));
```
nightlark, line 1109:
```suggestion
            "command cannot be called on uninitialized federate or after disconnect call"));
```

src/helics/application_api/Federate.cpp
```diff
@@ -1100,6 +1100,36 @@ void Federate::setGlobal(const std::string& valueName, const std::string& value)
...
+    }
+}
+
+std::pair<std::string, std::string> Federate::getCommand()
+{
+    if (coreObject) {
+        return coreObject->getCommand(fedID);
+    } else {
+        throw(InvalidFunctionCall(
+            " command cannot be called on uninitialized federate or after disconnect call"));
```
nightlark, line 1119:
```suggestion
            "command cannot be called on uninitialized federate or after disconnect call"));
```

src/helics/application_api/Federate.cpp
```diff
@@ -1100,6 +1100,36 @@ void Federate::setGlobal(const std::string& valueName, const std::string& value)
...
+    }
+}
+
+std::pair<std::string, std::string> Federate::waitCommand()
+{
+    if (coreObject) {
+        return coreObject->waitCommand(fedID);
+    } else {
+        throw(InvalidFunctionCall(
+            " command cannot be called on uninitialized federate or after disconnect call"));
```
nightlark, line 1129:
```suggestion
            "command cannot be called on uninitialized federate or after disconnect call"));
```

src/helics/application_api/Federate.hpp
```diff
@@ -356,6 +358,29 @@ class HELICS_CXX_EXPORT Federate {
     */
     void setGlobal(const std::string& valueName, const std::string& value);
 
+    /** send a command to another core or federate
+  @param target  the target of the command can be "federation", "federate", "broker", "core", or a
+  specific name of a federate, core, or broker
+  @param commandStr a string with the command instructions see other documentation for specific
```
nightlark, line 364:
```suggestion
  @param commandStr a string with the command instructions, see other documentation for specific
```

src/helics/application_api/Federate.hpp
```diff
@@ -356,6 +358,29 @@ class HELICS_CXX_EXPORT Federate {
...
+  properties to command, can be defined by a federate
+  */
+    void sendCommand(const std::string& target, const std::string& queryStr);
+
+    /** get any commands for the Federate
+ @param target  the target of the command can be "federation", "federate", "broker", "core", or a
+ specific name of a federate, core, or broker
+ @return a pair of strings <command,source> with the command instructions for the federate the
```
nightlark, line 372:
```suggestion
 @return a pair of strings <command,source> with the command instructions for the federate; the
```

src/helics/application_api/Federate.hpp
```diff
@@ -356,6 +358,29 @@ class HELICS_CXX_EXPORT Federate {
...
+ command string will be empty if no command is given
+ */
+    std::pair<std::string, std::string> getCommand();
+
+    /** get a commands for the Federate, call will block until a command is received
+@param target  the target of the command can be "federation", "federate", "broker", "core", or a
+specific name of a federate, core, or broker
+@return a pair of strings <command,source> with the command instructions for the federate
```
nightlark, line 380:
```suggestion
    /** get a command for the Federate, if there is none the call will block until a command is received
@param target  the target of the command can be "federation", "federate", "broker", "core", or a
specific name of a federate, core, or broker
@return a pair of strings <command,source> with the command instructions for the federate
```

src/helics/application_api/Federate.hpp
```diff
@@ -356,6 +358,29 @@ class HELICS_CXX_EXPORT Federate {
...
 
+    /** send a command to another core or federate
+  @param target  the target of the command can be "federation", "federate", "broker", "core", or a
+  specific name of a federate, core, or broker
+  @param commandStr a string with the command instructions see other documentation for specific
+  properties to command, can be defined by a federate
+  */
+    void sendCommand(const std::string& target, const std::string& queryStr);
+
+    /** get any commands for the Federate
```
nightlark, line 369:
```suggestion
    /** get a command for the Federate
```

tests/helics/system_tests/commandInterfaceTests.cpp
```diff
@@ -0,0 +1,326 @@
...
+    // register the publications
+    vFed1->registerGlobalPublication<double>("pub1");
+
+    vFed2->registerSubscription("pub1");
+    helics::BrokerApp brk1(brokers[0]);
+    helics::BrokerApp brk2(brokers[1]);
+    helics::BrokerApp brk3(brokers[2]);
+    brk1.sendCommand(vFed2->getName(), "test");
+    brk2.sendCommand(vFed2->getName(), "test");
+    brk3.sendCommand(vFed2->getName(), "test");
```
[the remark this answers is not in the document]
phlptp: They would be in the order you send them from a particular origin point.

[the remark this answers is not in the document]
phlptp: otherwise nothing is guaranteed between different sources.  they could come in any order

src/helics/application_api/Federate.hpp
```diff
@@ -356,6 +358,30 @@ class HELICS_CXX_EXPORT Federate {
     */
     void setGlobal(const std::string& valueName, const std::string& value);
 
+    /** send a command to another core or federate
+  @param target  the target of the command can be "federation", "federate", "broker", "core", or a
+  specific name of a federate, core, or broker
+  @param commandStr a string with the command instructions, see other documentation for specific
+  properties to command, can be defined by a federate
+  */
+    void sendCommand(const std::string& target, const std::string& queryStr);
```
nightlark, line 367:
```suggestion
    void sendCommand(const std::string& target, const std::string& commandStr);
```

nightlark (commented):
Header file used queryStr; matching function in the cpp file used commandStr.

nightlark (approved):

src/helics/shared_api_library/FederateExport.cpp
```diff
@@ -1224,3 +1224,52 @@ void helicsFederateLogLevelMessage(helics_federate fed, int loglevel, const char
...
+    auto* fedObj = getFed(fed, err);
+    if (fedObj == nullptr) {
+        return;
+    }
+    fedObj->sendCommand(AS_STRING(target), AS_STRING(command));
+}
+
+const char* helicsFederateGetCommand(helics_federate fed, helics_error* err)
+{
+    auto* fedObj = helics::getFedObject(fed, err);
```
eberleim, line 1239: Just confirming because I do not have the code open at the moment:
getFedObject returns an access pointer to an object persistent in memory elsewhere, correct? (verifying that the use of `c_str()` below will be an access into a persistent data point and not a temporary).

eberleim (approved):
Some comments, but none observing issues I consider significant enough to withhold approval.

src/helics/core/CommonCore.cpp
```diff
@@ -2358,6 +2376,31 @@ void CommonCore::initializeMapBuilder(const std::string& request,
     }
 }
 
+void CommonCore::processCommandInstruction(ActionMessage& command)
+{
+    auto cmd = command.payload.to_string();
+    if (cmd == "terminate") {
```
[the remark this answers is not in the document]
phlptp: A couple more are likely particularly in regard to debugging or profiling.  Probably not that many though. and it is annoying to make a switch statement work with strings.

src/helics/shared_api_library/FederateExport.cpp
phlptp: Yeah the actual command and source get copied to a value in the FederateObject,  So it is stored until the next command is retrieved.

PR #1572, opened 2020-09-14
Change library names

phlptp:
### Summary

If merged this pull request will change the names of the shared libraries to libhelics and libhelicscpp
and remove the python interface and the helics-config app


phlptp:
@nightlark @kdheepak It would be nice to get a pre-release of HELICS 3 out this week, can you take a look at the changes to the library names, and see if there is anywhere else we need to change to account for this modification

nightlark (approved):

Issue #1380, opened 2020-05-29
helicsPublicationPublishString() intermittently fails on TCP core on Linux


vbarsuk:
**Describe the bug**
On intense traffic helicsPublicationPublishString() can fail on TCP core on Linux.

python: /........./helics/helics_2.5/HELICS-src/src/helics/network/tcp/TcpHelperClasses.cpp:282: size_t helics::tcp::TcpConnection::send(const string&): Assertion `sz == dataString.size()' failed.

on debug run sz value was 30, dataString.size() was 4222
With Linux it is possible to get a successful partial write into TCP socket in async mode.

**What is the expected behavior?**
 send() should not fail on assert() with partial write

**To Reproduce**
Keep  publishing of large strings

**Environment (please complete the following information):**
 
Happened on helics 2.5 but I believe many other versions can be affected.
Environment: CentOS Linux release 7.4.1708 (Core)
```
$ helics_app --version
```

**Additional context and information**
I believe it is a exclusive behavior of TCP sockets on Linux. I never saw such a behavior on Windows or any other system.
I was analyzing the source code when found the issue and I was able to reproduce the issue under the load.

In TcpHelperClasses.cpp in send() , mere assert() is not sufficient.

phlptp:
So is the issue that the return from the TCP send call can on occasion return a value less then requested?  In which case did it not send it send a partial packet and we should send the rest or treat that as a failure and try again?    

I agree on the assertion, the question is what should be done in that case.   check partial success and try again on the part that wasn't sent?

vbarsuk:
Yes.
With Windows and other unix systems send() either sends nothing or everything. According to specs it still can send a partial buffer.
I saw this behavior on Linux only.
I would keep sending the rest of the buffer in the loop till it is completely sent.
I will work on the solution if it is ok with you.

Slava

phlptp:
If you have a good understanding of how to fix it a PR would be welcome.

phlptp:
@vbarsuk I know you were working on a PR,  what is the status and any chance it would be ready by the end of the week or early next at the latest.  Otherwise would get it into 2.6.1 release.

PR #1600, opened 2020-09-25
Retry partial TCP buffer sends

vbarsuk:
### Summary

If merged this pull request will address  the bug #1380 

### Proposed changes

resend the non sent part of the buffer up to 5 times

### DIFF
--- a/src/helics/network/tcp/TcpHelperClasses.cpp
+++ b/src/helics/network/tcp/TcpHelperClasses.cpp
@@ -257,13 +257,27 @@ namespace tcp {
                 return 0;
             }
         }
-        auto sz = socket_.send(asio::buffer(buffer, dataLength));
-        assert(sz == dataLength);
-        return sz;
+	
+	size_t	sz,sent_size;
+	int p=0,count=0;
+	sent_size=dataLength;
+        while (count++  < 5 && (sz = socket_.send(asio::buffer((void *)((char *)buffer+p), sent_size))) != sent_size) {
+	   sent_size-=sz; p+=sz;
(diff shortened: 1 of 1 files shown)


phlptp:
Looks like most of the linter checks are either formatting or stuff that was already preexisting.  The formatting will get taken care of once merged and the other things we can deal with later.  

@vbarsuk Is there a test case we could devise that would trigger the new code?

src/helics/network/tcp/TcpHelperClasses.cpp
```diff
@@ -257,13 +257,27 @@ namespace tcp {
...
-        return sz;
+	
+	size_t	sz,sent_size;
+	int p=0,count=0;
+	sent_size=dataLength;
+        while (count++  < 5 && (sz = socket_.send(asio::buffer((void *)((char *)buffer+p), sent_size))) != sent_size) {
+	   sent_size-=sz; p+=sz;
+	//   std::cerr << "DEBUG partial buffer sent" << std::endl;
+	}
+	if(count < 5 ) return dataLength;
```
phlptp, line 268: This should be count<=5 if I am not mistaken?   and any particular reason for 5 being chosen here?

vbarsuk: From what I saw while debugging, one  extra resend works in most cases, sometimes it requires two. The max number of sends it set to  4 (sorry, not 5) to eliminate some false positives,  so it should be sufficient for now I believe. If it would require more than 4, then most likely it will be going for a very large number of resends. It may be beneficial to add a very short sleep between resends, but it may negatively affect the performance.

phlptp (approved):

PR #1602, opened 2020-09-28
update the default libzmq for build

phlptp:
Update default libzmq to 4.3.3

### DIFF
--- a/config/cmake/addlibzmq.cmake
+++ b/config/cmake/addlibzmq.cmake
@@ -19,11 +19,12 @@ else()
     set(zmq_shared_build ON)
 endif()
 
-if(MINGW)
-    set(${PROJECT_NAME}_LIBZMQ_VERSION v4.3.1)
-else()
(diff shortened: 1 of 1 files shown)


config/cmake/addlibzmq.cmake
```diff
@@ -124,6 +125,17 @@ set(WITH_PERF_TOOL
     OFF
     CACHE INTERNAL ""
 )
+
+set(WITH_LIBSODIUM_STATIC
+    OFF
+    CACHE INTERNAL ""
+)
```
nightlark, line 132: Okay. `ENABLE_CURVE` is also off so this shouldn't have the TweetNacl mutex.hpp error that we were seeing before.

nightlark (approved):

config/cmake/addlibzmq.cmake
phlptp: This was a new option to enable a static libsodium,  it is off by default,  I just didn't want to see it the CMAKE options for HELICS.

PR #1607, opened 2020-09-29
Strict config check

phlptp:
### Summary

If merged this pull request will add a strict config check option on by default to throw errors if there are any invalid values in a parts of a config file.  It also adds some getter functions for the FederateInfo object which will be expanded on at a later time and exposed to the C API.  

### Proposed changes

- add strict_config_check option for some config file options
- send the warnings/errors through the logging system instead of std::cerr for the above config file errors
- add a getter for the FederateInfo object
- remove some unneeded portions of the libzmq builder from the previous PR


trevorhardy:
I'm not seeing anything new in this code that would catch the bug I was seeing (changing the filter type to "reroute3", an invalid type). Is that not part of this?

src/helics/application_api/Federate.cpp
```diff
@@ -834,12 +839,30 @@ void Federate::registerFilterInterfacesJson(const std::string& jsonString)
...
+                } else {
+                    logMessage(helics_log_level_warning,
+                               "input and output types may only be specified for custom filters");
+                }
                 continue;
             }
             if (!useTypes) {
                 if (opType == filter_types::unrecognized) {
-                    std::cerr << "unrecognized filter operation:" << operation << '\n';
+                    if (strictConfigChecking) {
```
phlptp, line 855: @trevorhardy this is the code that would catch the invalid filter type and generate an error

trevorhardy (approved):

src/helics/application_api/Federate.cpp
```diff
@@ -834,12 +839,30 @@ void Federate::registerFilterInterfacesJson(const std::string& jsonString)
...
+                        logMessage(helics_log_level_error,
+                                   fmt::format("unrecognized filter operation:{}", operation));
+
+                        throw(InvalidParameter(
+                            "input and output types may only be specified for custom filters"));
+                    } else {
+                        logMessage(
+                            helics_log_level_warning,
+                            "input and output types may only be specified for custom filters");
```
nightlark, line 864: Are the other log messages here correct?

src/helics/application_api/Federate.cpp
```diff
@@ -921,12 +965,30 @@ void Federate::registerFilterInterfacesToml(const std::string& tomlString)
...
+                    }
```
nightlark, line 991: Also these log messages don't all match.

src/helics/application_api/Federate.cpp
```diff
@@ -857,8 +880,18 @@ void Federate::registerFilterInterfacesJson(const std::string& jsonString)
                 if (props.isArray()) {
                     for (const auto& prop : props) {
                         if ((!prop.isMember("name")) || (!prop.isMember("value"))) {
-                            std::cerr
-                                << "properties must be specified with \"name\" and \"value\" fields\n";
+                            if (strictConfigChecking) {
+                                logMessage(
+                                    helics_log_level_error,
+                                    "properties must be specified with \"name\" and \"value\" fields");
```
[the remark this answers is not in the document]
phlptp: In theory others will be ignored

nightlark (approved):

PR #1624, opened 2020-10-03
Remove Travis Linux/macOS jobs migrated to Azure

nightlark:
### Summary

If merged this pull request will remove the macOS and Linux jobs from Travis that were moved to Azure Pipelines. Adds an empty Linux job that does nothing, because Travis doesn't like it when all the jobs in a config file are conditional.

Build job testing Windows mingw and msys builds using keywords in commit message: https://travis-ci.com/github/GMLC-TDC/HELICS/builds/187930088
- MinGW seems to start compiling, but has build errors
- MSYS build is having issues; repo.msys2.org connection fails, likely issues with the msys2 repos

### Proposed changes

- Remove Linux jobs from Travis
- Remove macOS jobs from Travis
- Remove Travis Brewfiles
- Add empty Linux job

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -44,6 +44,12 @@ jobs:
   fast_finish: true
 
   include:
+    - os: Linux
+      name: 'Default nothing job'
+      before-install: echo 'Before Install'
+      install: echo 'Install'
+      script: echo 'Script'
+      
     - os: windows
       name: 'MinGW'
       if: (branch = develop AND type IN (cron)) OR (branch = helics3 AND type IN (cron)) OR (branch = master AND type IN (push, pull_request)) OR (branch =~ /(mingw)/) OR (commit_message =~ /(mingw)/)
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -38,130 +38,17 @@ _basic_env:
...
-            - clang-3.6
-            - libzmq3-dev
-      env:
-        - MATRIX_EVAL="COMPILER=clang && CC='clang-3.6' && CXX='clang++-3.6'"
-        - CCACHE_CPP2=yes
-        - USE_SWIG=true
-        - CI_BOOST_VERSION=1.58.0
-        - SKIP_TEST_RUN=true
-        - JOB_OPTION_FLAGS="-DHELICS_DISABLE_BOOST=ON"
+    - os: Linux
```
phlptp, line 47: What is this linux job doing?

nightlark: It’s making so Travis doesn’t say the entire config file is invalid because all the remaining builds only run conditionally (probably a bug that Travis needs to fix).

phlptp (approved):

PR #1626, opened 2020-10-03
Create compress images GitHub Actions workflow

nightlark:
### Summary

If merged this pull request will add a workflow that tries to compress added images. For PRs the compressed images can be committed directly if they originated from the same repository.

A later addition (or change to this PR) could make it comment on PRs coming from forks with the results of compressing images, and open a PR when commits result in images that were able to be compressed.

### Proposed changes

- Add a compress images GitHub actions workflow

### DIFF
--- a/.github/workflows/compress-images.yml
+++ b/.github/workflows/compress-images.yml
@@ -1,0 +1,36 @@
+name: Compress Images
+on:
+  pull_request:
+    paths:
+      - '**.jpg'
+      - '**.jpeg'
+      - '**.png'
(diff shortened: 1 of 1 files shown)


phlptp:
@nightlark  what is happening with this,  anything else needed before merge?

nightlark:
It worked in my fork, so I think it's okay to merge. There were some comments with ideas for additions that would be nice to have  for external contributors working from a branch.

phlptp (approved):

PR #1628, opened 2020-10-03
Update macOS deployment target to 10.14

nightlark:
### Summary

Install macOS 10.14 SDK and set macOS 10.14 as the deployment target for release builds.

If merged this pull request will set `LIBGUARDED_NO_DEFAULT` when the macOS deployment target is < 10.12, so if someone wants to try to mess with the default macOS stdlib to build on older versions of macOS < 10.14 they can try.

### Proposed changes

- Install macOS 10.14 SDK for release builds
- Update macOS deployment target version to 10.14 for release builds
- Set `LIBGUARDED_NO_DEFAULT` when building for macOS SDKs < 10.12
- Sets some workflow_dispatch things for testing


phlptp:
Is this tested in any of the CI builds?

nightlark:
When the helics3 branch becomes develop the workflow that uses this gets tested nightly. I'll need to use the workflow_dispatch  change to manually test this change.

phlptp (approved):

Issue #1564, opened 2020-09-06
Custom query


mzy2240:
Howdy!

How to register custom query callback on federates? It seems to me that the only approach is to use setQueryCallback function inside the FederateState class, however did not find the FedderateState importable in the python wrapper.

phlptp:
Right now this callback is only available on the C++ interface.   The federateState object is only in the Core library so would likely never be available in the python API.    The plan is to make that callback available in the C API shortly and that would then be available in the python API through the [pyhelics](https://github.com/GMLC-TDC/pyhelics) API, as that is much better able to handle python based callbacks.

mzy2240:
Thank you @phlptp 
If the custom callback is not available for now in the python wrapper, how to make a work around if I want to send a time-independent discrete event, like a metadata? I tried to use the endpoints with the function `helicsEndpointSendEventRaw`, however the time argument is not optional which also does not follow the C API reference.

phlptp:
If the metadata is at the start of a co-simulation it is possible to send message data though an endpoint in the initialization state of a co-simulation and use iteration.  

Really though the query interface is the correct thing to use, we just haven't had anyone actually use it yet outside of C++, and given we didn't have a clean way to do the callbacks outside of C++ before the very recent pyHELICS it wasn't a priority. 
   In HELICS 3.0 a control interface would also be available for asynchronous interaction between federates.  

But since someone is asking we will make sure to get it into the 2.6.1 release in a couple weeks.

mzy2240:
Thanks! Today we managed to use the endpoint to send the metadata during the co-simulation, though the timestamp associated with the endpoint message does not make many sense to me, it survives our test and could be used to fulfill our need. Anyway, really looking forward to the new pyhelics with better python support.

BTW, sending a jsonify string does not work really surprised me today. It seems like the comma caused the problem.

phlptp:
I thought of one other option that would be available.  If what you are sending is fixed global information you can use setGlobal to define some data.  then a query can retrieve the information from any federate at any time.

PR #1634, opened 2020-10-06
C api query callback

phlptp:
### Summary

If merged this pull request will add a callback for a federate to respond to custom queries.

### DIFF
--- a/src/helics/application_api/Federate.hpp
+++ b/src/helics/application_api/Federate.hpp
@@ -219,7 +219,7 @@ class HELICS_CXX_EXPORT Federate {
     Time requestTime(Time nextInternalTimeStep);
 
     /** request a time advancement to the next allowed time
-  @return the granted time step*/
+    @return the granted time step*/
     Time requestNextStep() { return requestTime(timeZero); }
 
     /** request a time advancement by a certain amount
@@ -349,12 +349,12 @@ class HELICS_CXX_EXPORT Federate {
 
     /** get the results of an async query
     @details the call will block until the results are returned inquiry of queryCompleted() to check
-   if the results have been returned or not yet
+    if the results have been returned or not yet
 
     @param queryIndex the int value returned from the queryAsync call
-   @return a string with the value requested.  the format of the string will be either a single
-   string a string vector like "[string1; string2]" or JSON The string "#invalid" is returned if the
-   query was not valid
+    @return a string with the value requested.  the format of the string will be either a single
+    string a string vector like "[string1; string2]" or JSON The string "#invalid" is returned if the
+    query was not valid
     */
     std::string queryComplete(query_id_t queryIndex);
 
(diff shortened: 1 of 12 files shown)


phlptp:
Address #1564

src/helics/application_api/Federate.hpp
```diff
@@ -363,6 +363,16 @@ class HELICS_CXX_EXPORT Federate {
...
     bool isQueryCompleted(query_id_t queryIndex) const;
 
+    /** supply a query callback function
+  @details the intention of the query callback is to allow federates to answer particular requests
+  through the query interface this allows other federates to make requests or queries of other
+  federates in an asynchronous fashion.
+  @param queryFunction  a function object that returns a string as a result of a query in the form
+  of const string ref. This callback will be called when a federate received a query that cannot
+  be answered that directed at a particular federate
+  */
```
kdheepak, line 373: Can we add 4 spaces here before the docstring, instead of 2 spaces?

kdheepak (approved):

src/helics/shared_api_library/helicsCallbacks.h
```diff
@@ -98,6 +98,44 @@ HELICS_EXPORT void helicsFilterSetCustomCallback(helics_filter filter,
...
+ * Set the data for a query callback
+ *
+ * @details There are many queries that HELICS understands directly, but it is occasionally useful to have a federate be able to respond
+ * to specific queries with answers specific to a federate
+ *
+ * @param buffer the buffer received in a helicsQueryCallback
+ * @param str pointer to the data to fill the buffer with
+ * @param strSize the size of the string
+ * @forcpponly
+ * @param[in,out] err A pointer to an error object for catching errors.
```
nightlark, line 134: Matching other C api doc comments.

```suggestion
 * Set the data for a query callback.
 *
 * @details There are many queries that HELICS understands directly, but it is occasionally useful to have a federate be able to respond
 * to specific queries with answers specific to a federate.
 *
 * @param buffer The buffer received in a helicsQueryCallback.
 * @param str Pointer to the data to fill the buffer with.
 * @param strSize The size of the string.
 * @forcpponly
 * @param[in,out] err A pointer to an error object for catching errors.
```

src/helics/shared_api_library/helicsCallbacks.h
```diff
@@ -98,6 +98,44 @@ HELICS_EXPORT void helicsFilterSetCustomCallback(helics_filter filter,
...
+ * to specific queries with answers specific to a federate
+ *
+ * @param fed The federate to set the callback for
+ * @param queryAnswer A callback with signature const char *(const char *query, int querySize,int *answerSize, void *userdata);
+ *                 The function arguments are the query string requesting an answer along with its size, the string is not guaranteed to be
+ * null terminated answerSize is an outputParameter intended to filled out by the userCallback and should contain the length of the return
+ * string. the return pointer can be NULL if no answer is given and HELICS will generate the appropriate response.
+ * @param userdata A pointer to user data that is passed to the function when executing.
+ * @forcpponly
+ * @param[in,out] err A pointer to an error object for catching errors.
```
nightlark, line 114: Matching other C api doc comments.

```suggestion
 * Set callback for queries executed against a federate.
 *
 * @details There are many queries that HELICS understands directly, but it is occasionally useful to have a federate be able to respond
 * to specific queries with answers specific to a federate.
 *
 * @param fed The federate to set the callback for.
 * @param queryAnswer A callback with signature const char *(const char *query, int querySize,int *answerSize, void *userdata);
 *                 The function arguments are the query string requesting an answer along with its size, the string is not guaranteed to be
 * null terminated answerSize is an outputParameter intended to filled out by the userCallback and should contain the length of the return
 * string. The return pointer can be NULL if no answer is given and HELICS will generate the appropriate response.
 * @param userdata A pointer to user data that is passed to the function when executing.
 * @forcpponly
 * @param[in,out] err A pointer to an error object for catching errors.
```

src/helics/core/CommonCore.cpp
```diff
@@ -2064,8 +2064,7 @@ std::string CommonCore::federateQuery(const FederateState* fed, const std::strin
         return filteredEndpointQuery(fed);
     }
     if ((queryStr == "queries") || (queryStr == "available_queries")) {
-        return std::string(
-                   "[exists;isinit;state;global_state;version;queries;filtered_endpoints;current_time;") +
+        return std::string("[exists;isinit;state;version;queries;filtered_endpoints;") +
```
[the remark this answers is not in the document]
phlptp: This list is queries about a federate that are answered by a core, so current time in this list would be duplicative as it is also in the federate list

nightlark (approved):

Issue #1552, opened 2020-08-24
`helicsFederateRegisterInterfaces` should throw an error if the publications and inputs are already registered


kdheepak:
**Describe the bug**

Currently, if a user calls `helicsCreateValueFederateFromConfig` and then calls `helicsFederateRegisterInterfaces` on the same json file, HELICS does not error.

**What is the expected behavior?**

It should error and present the user with a clear error message.

**To Reproduce**

See https://github.com/GMLC-TDC/HELICS/issues/1493

**Environment (please complete the following information):**

- Operating System: [all]
- Language Extension: [all]
- Latest release from github
- HELICS version: '2.6.0 (2020-08-20)'

phlptp:
I am not sure they should show an error.  The intention was to allow flag modification of previously created interfaces to add targets, or flag controls.  So it shouldn't be an error to have existing interfaces be present in a json file.

PR #1639, opened 2020-10-08
Add handling for duplicate target specification

phlptp:
### Summary

If merged this pull request will handle a call to registerInterface with the same file as in the create step and deal with duplicate targets appropriately.

### Proposed changes

- fix a bug in the json file handler in the case of duplicate subscriptions
- handle duplicate targets for publications and subscriptions appropriately
- add a warning for duplicate targets for inputs.  
- Fix Issue #1552   not by error but just handling the situation appropriately

### DIFF
--- a/src/helics/application_api/ValueFederateManager.cpp
+++ b/src/helics/application_api/ValueFederateManager.cpp
@@ -126,21 +126,20 @@ void ValueFederateManager::addTarget(const Publication& pub, const std::string& 
 void ValueFederateManager::addTarget(const Input& inp, const std::string& target)
 {
     {
-    auto iTHandle = inputTargets.lock();
-    auto rng = iTHandle->equal_range(inp.handle);
-    for (auto el = rng.first; el != rng.second; ++el) {
-        if (el->second == target) {
-            fed->logWarningMessage(std::string("Duplicate input targets detected for ") +
-                                   inp.actualName + "::" + target);
-            return;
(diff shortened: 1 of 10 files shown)


nightlark:
It looks like the application api tests have started timing out. The previous successful run on the helics3 branch took 290s for the application tests, now they're taking >480s (timeout limit).

phlptp:
@eberleim  I am going to start adding you as a reviewer on most pull requests dealing with the core

nightlark (approved):

kdheepak (approved):

src/helics/application_api/ValueFederateManager.cpp
```diff
@@ -125,9 +125,21 @@ void ValueFederateManager::addTarget(const Publication& pub, const std::string&
...
+            if (el->second == target) {
+                fed->logWarningMessage(std::string("Duplicate input targets detected for ") +
+                                       inp.actualName + "::" + target);
+                return;
+            }
+        }
+    }
+
     coreObject->addSourceTarget(inp.handle, target);
+    auto it = inputTargets.lock()->emplace(inp.handle, target);
```
eberleim, line 141: Are there plans to use the `iterator it` returned by emplace at a later date? Not technically a problem, but the storage of an immediately deleted temporary seemed weird.

eberleim (approved):

Issue #1548, opened 2020-08-20
`helicsGetPropertyIndex` does not work for `FILE_LOG_LEVEL` and `CONSOLE_LOG_LEVEL`


kdheepak:
**Describe the bug**

![Screen Shot 2020-08-20 at 11 23 01 AM](https://user-images.githubusercontent.com/1813121/90804639-bde57900-e2d7-11ea-8fcc-1e23f6488808.png)

**What is the expected behavior?**

`helicsGetPropertyIndex` should work consistently across all properties.

**Environment:**

- Operating System: [ Mac ]
- Language Extension: [ Python ]
- Official release on github
- HELICS version: '2.5.2 (2020-06-14)'

**Additional information**

##### Works

- [x] `h.helicsGetPropertyIndex("DELTA")`
- [x] `h.helicsGetPropertyIndex("PERIOD")`
- [x] `h.helicsGetPropertyIndex("OFFSET")`
- [x] `h.helicsGetPropertyIndex("RT_LAG")`
- [x] `h.helicsGetPropertyIndex("RT_LEAD")`
- [x] `h.helicsGetPropertyIndex("RT_TOLERANCE")`
- [x] `h.helicsGetPropertyIndex("INPUT_DELAY")`
- [x] `h.helicsGetPropertyIndex("OUTPUT_DELAY")`
- [x] `h.helicsGetPropertyIndex("MAX_ITERATIONS")`
- [x] `h.helicsGetPropertyIndex("LOG_LEVEL")`

##### Does not work

- [ ] `h.helicsGetPropertyIndex("FILE_LOG_LEVEL")`
- [ ] `h.helicsGetPropertyIndex("CONSOLE_LOG_LEVEL")`

I also tried adding the `TIME_` and `INT_` prefix, and I get a different set of behaviors.

##### Works

- [x] `h.helicsGetPropertyIndex("TIME_DELTA")`

##### Does not work

- [ ] `h.helicsGetPropertyIndex("TIME_PERIOD")`
- [ ] `h.helicsGetPropertyIndex("TIME_OFFSET")`
- [ ] `h.helicsGetPropertyIndex("TIME_RT_LAG")`
- [ ] `h.helicsGetPropertyIndex("TIME_RT_LEAD")`
- [ ] `h.helicsGetPropertyIndex("TIME_RT_TOLERANCE")`
- [ ] `h.helicsGetPropertyIndex("TIME_INPUT_DELAY")`
[...]

PR #1642, opened 2020-10-09
Issue #1548

phlptp:
### Summary

If merged this pull request will address Issue #1548 

### Proposed changes

- add valid property strings
- add test for strings including the same strings as the enumerations


phlptp:
@kdheepak  @nightlark  There were some issues with the formatting action, that is causing some issues.  This one needs to be merged next to resolve them.  Then we might have to rebase the others.

nightlark (approved):

src/helics/application_api/FederateInfo.cpp
```diff
@@ -37,37 +38,64 @@ FederateInfo::FederateInfo(const std::string& args)
     loadInfoFromArgsIgnoreOutput(args);
 }
 
-static const std::map<std::string, int> propStringsTranslations{
+static const std::unordered_map<std::string, int> propStringsTranslations{
     {"period", helics_property_time_period},
+    {"timeperiod", helics_property_time_period},
+    {"timeperiod", helics_property_time_period},
```
kdheepak, line 44: Aren't these two lines the same?

```suggestion
    {"timeperiod", helics_property_time_period},
```

kdheepak (approved):

src/helics/application_api/FederateInfo.cpp
```diff
@@ -37,37 +38,64 @@ FederateInfo::FederateInfo(const std::string& args)
...
+    {"timeinputdelay", helics_property_time_input_delay},
+    {"timeoutputdelay", helics_property_time_output_delay},
+    {"time_input_delay", helics_property_time_input_delay},
+    {"time_output_delay", helics_property_time_output_delay},
     {"loglevel", helics_property_int_log_level},
     {"log_level", helics_property_int_log_level},
-    {"maxiterations", helics_property_int_max_iterations},
     {"logLevel", helics_property_int_log_level},
+    {"intloglevel", helics_property_int_log_level},
+    {"intlog_level", helics_property_int_log_level},
```
kdheepak, line 82: Why are we adding `intlog_level` and not `int_log_level` when later we are adding `int_file_log_level`. 

I think we shouldn't have any `int{}` or `time{}` variant. Basically, I would like to only support users entering the attribute as we use it in the code but without the `helics_property_time_` or `helics_property_int_` prefix.

I think there should be just one and obvious way to use a property. For example, I think it is confusing for users whether `iterations` and `max_iterations` are the same thing or not

src/helics/application_api/FederateInfo.cpp
phlptp: I think I meant to add an underscore

src/helics/application_api/FederateInfo.cpp
phlptp: The "int" and "time" was what you were testing in the issue this is a based on.  Also in helics3 @trevorhardy  removed a lot of the variants so when I port this over, many will be removed.  I can remove them happily but then it won't technically be resolving the issue?

phlptp: @kdheepak   Do you think we should have the "time" and "int" variants like suggested in #1642 ?

phlptp:
@kdheepak  I am going to merge this with the int and time strings.  We can discuss further if we want to remove them later I am fine either way.  But this is blocking some other stuff right now.

Issue #1550, opened 2020-08-20
`helicsGetFlagIndex` doesn't work for all enumerations


kdheepak:
**Describe the bug**

`helicsGetFlagIndex` doesn't work for the following flags:

```
OBSERVER
UNINTERRUPTIBLE
INTERRUPTIBLE
DELAY_INIT_ENTRY
ENABLE_INIT_ENTRY
IGNORE_TIME_MISMATCH_WARNINGS
```

See following code for reproducible example:

```python
In [1]: import helics as h
In [2]: for f in h.HelicsFederateFlag:
    ...:     try:
    ...:         print(f, h.helicsGetFlagIndex(f.name))
    ...:     except h.HelicsException as e:
    ...:         print(e)
    ...:
[-1] Unknown property index for flag `OBSERVER`
[-1] Unknown property index for flag `UNINTERRUPTIBLE`
[-1] Unknown property index for flag `INTERRUPTIBLE`
HelicsFederateFlag.SOURCE_ONLY HelicsFederateFlag.SOURCE_ONLY
HelicsFederateFlag.ONLY_TRANSMIT_ON_CHANGE HelicsFederateFlag.ONLY_TRANSMIT_ON_CHANGE
HelicsFederateFlag.ONLY_UPDATE_ON_CHANGE HelicsFederateFlag.ONLY_UPDATE_ON_CHANGE
HelicsFederateFlag.WAIT_FOR_CURRENT_TIME_UPDATE HelicsFederateFlag.WAIT_FOR_CURRENT_TIME_UPDATE
HelicsFederateFlag.RESTRICTIVE_TIME_POLICY HelicsFederateFlag.RESTRICTIVE_TIME_POLICY
HelicsFederateFlag.REALTIME HelicsFederateFlag.REALTIME
HelicsFederateFlag.SLOW_RESPONDING HelicsFederateFlag.SLOW_RESPONDING
[-1] Unknown property index for flag `DELAY_INIT_ENTRY`
[-1] Unknown property index for flag `ENABLE_INIT_ENTRY`
[-1] Unknown property index for flag `IGNORE_TIME_MISMATCH_WARNINGS`
HelicsFederateFlag.TERMINATE_ON_ERROR HelicsFederateFlag.TERMINATE_ON_ERROR
```

**What is the expected behavior?**

`helicsGetFlagIndex` should work for all enumeration names in `helics_federate_flags`.

**Environment (please complete the following information):**

- Operating System: [ Mac ]
[...]

PR #1645, opened 2020-10-10
add indexing tests and missing ones for the flag enumerations

phlptp:
### Summary

If merged this pull request will resolve Issue #1550 and allow all the flags to be used via a string


phlptp:
@kdheepak  I updated the enumerations.  I didn't change the actual symbols in the enumerations but did add two additional enumeration names  helics_core_flags and helics_flags , if you can think of a better name happy to listen.

phlptp:
@kdheepak  we need to get this in today, can you review.  It sounds like we have some good discussion on changes for helics3 but I don't want to change the API in a 2.6.1 release.  There is one more PR after this, then the release notes that we need to get through pretty quickly since I will not be able to much for a week.

kdheepak (approved):

PR #1648, opened 2020-10-13
Broker port tests

phlptp:
### Summary

If merged this pull request will fix a potential mechanism to force the system to try to bind to negative port numbers if custom port numbers are used.


eberleim (approved):
This seems to work, and also seems to fix the intermittent connection failures I was experiencing yesterday.

PR #1659, opened 2020-10-14
add debugging flag

phlptp:
### Summary

If merged this pull request will add a flag to indicate a user debugging session may occur.  This turns off some timeouts and enables the slow_responding indicator.  There is nothing new here other than support for the flag to trigger others.


phlptp:
@eberleim 
I think part of the issue you were having with federates disconnecting when debugging is the timeouts.  This PR adds a single flag to turn off some of the timeouts and enable the slow_responding flag.

eberleim (approved):

PR #1665, opened 2020-10-14
Merge workflow_dispatch trigger from master into develop

nightlark:
Merges the commit with the workflow_dispatch trigger into develop. Start gradually getting the change propagated to other branches.


phlptp:
Do we need this merged before we tag a release?

nightlark:
I think we can merge from develop to master for 2.6.1 and then might be able to do a fast-forward merge to update develop.

Issue #1647, opened 2020-10-13
Rename `helicsMessageSetData` and `helicsMessageAppendData` to `helicsMessageSetRawData` and `helicsMessageAppendRawData`


kdheepak:
There's `helicsMessageGetString` and `helicsMessageSetString`:

https://github.com/GMLC-TDC/HELICS/blob/2ef452943cc79d5572c9a860b05e03d96af0d6f6/src/helics/shared_api_library/MessageFederate.h#L474
https://github.com/GMLC-TDC/HELICS/blob/2ef452943cc79d5572c9a860b05e03d96af0d6f6/src/helics/shared_api_library/MessageFederate.h#L663

But there's `helicsMessageGetRawData` and `helicsMessageSetData`:

https://github.com/GMLC-TDC/HELICS/blob/2ef452943cc79d5572c9a860b05e03d96af0d6f6/src/helics/shared_api_library/MessageFederate.h#L519-L520
https://github.com/GMLC-TDC/HELICS/blob/2ef452943cc79d5572c9a860b05e03d96af0d6f6/src/helics/shared_api_library/MessageFederate.h#L675

Alternatively, rename `helicsMessageGetRawData` and `helicsMessageGetRawDataSize` to `helicsMessageGetData` and `helicsMessageGetDataSize`.

PR #1682, opened 2020-11-03
remove a redundant raw from some of the method names

phlptp:
### Summary

If merged this pull request will address Issue #1647.  It removes a redundent raw from a few API calls related to messages

### DIFF
--- a/docs/user-guide/simulator_integration.md
+++ b/docs/user-guide/simulator_integration.md
@@ -211,7 +211,7 @@ h.helicsPublicationPublishChar(pub, string_value)
 For sending a message through an endpoint, that once again looks a little bit different, in this case because - unlike with a publication - a message requires a destination. If a default destination was set when the endpoint was registered (either through the config file or through calling `h.helicsEndpointSetDefaultDestination()`), then an empty string can be passed. Otherwise, the destination must be provided as shown in API call below where dest is the destination and msg is the message to be sent.
 
 ```python
-h.helicsEndpointSendMessageRaw(end, dest, msg)
+h.helicsEndpointSendTo(end, msg, dest)
 ```
 
 #### Request the Next Time Step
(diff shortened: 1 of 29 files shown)


kdheepak (approved):

kdheepak:
What are your thoughts on using `Bytes` instead of `Data`? In my opinion, `Bytes` conveys the type of the payload more clearly than `Raw` or `Data`.

phlptp:
Do you think we should use `Bytes` on the publication function names as well?

phlptp:
I converted all use of `Raw` with Bytes.  I think I am a little ambivalent on the use for the message, but I think it does improve things on the Value side,  so instead of `getRawValue` it is `getBytes`   @kdheepak  let me know what you think

phlptp:
@kdheepak  I changed to using Bytes,  need you to review again see what you think of the changes

docs/user-guide/simulator_integration.md
```diff
@@ -211,7 +211,7 @@ h.helicsPublicationPublishChar(pub, string_value)
 For sending a message through an endpoint, that once again looks a little bit different, in this case because - unlike with a publication - a message requires a destination. If a default destination was set when the endpoint was registered (either through the config file or through calling `h.helicsEndpointSetDefaultDestination()`), then an empty string can be passed. Otherwise, the destination must be provided as shown in API call below where dest is the destination and msg is the message to be sent.
 
 ```python
-h.helicsEndpointSendMessageRaw(end, dest, msg)
+h.helicsEndpointSendTo(end, msg, dest)
```
kdheepak, line 214: Even here, I think `helicsEndpointSendBytesTo` reads better than `helicsEndpointSendTo`.

phlptp: I don't really like it there since there is no alternative,  in the other cases Bytes are used there are other functions for the different types.

kdheepak: I think just by reading the function signature (especially in languages like Python, Julia, etc):

```
h.helicsEndpointSendTo(end, msg, dest)
```

it can be hard to tell whether this takes a String, Bytes or a Message object.

src/helics/cpp98/Endpoint.hpp
```diff
@@ -274,15 +275,15 @@ class Endpoint {
     */
     void sendToAt(const void* data, size_t data_size, const std::string& dest, helics_time time)
```
kdheepak, line 276: For clarification, leaving this as it is right now is fine.

phlptp:
@nightlark waiting on your review now.

nightlark (approved):

Issue #1598, opened 2020-09-25
warning: field 'buffer' is uninitialized when used here [-Wuninitialized]


kdheepak:
I'm compiling from the `v3.0.0-alpha.1` tag and I get the following warning:

```
In file included from HELICS/src/helics/network/tcp/TcpCommsSS.cpp:8:
In file included from HELICS/src/helics/network/tcp/TcpCommsSS.h:9:
In file included from HELICS/src/helics/network/tcp/../NetworkCommsInterface.hpp:9:
In file included from HELICS/src/helics/network/tcp/../CommsInterface.hpp:13:
In file included from HELICS/src/helics/core/ActionMessage.hpp:10:
HELICS/src/helics/core/SmallBuffer.hpp:49:30: warning: field 'buffer' is uninitialized when used here [-Wuninitialized]
    SmallBuffer(U&& u): heap(buffer.data())
                             ^ 
```

This is on macosx:

```
$ clang --version
Apple clang version 11.0.3 (clang-1103.0.32.59)
Target: x86_64-apple-darwin19.6.0
Thread model: posix
InstalledDir: /Library/Developer/CommandLineTools/usr/bin
```

PR #1688, opened 2020-11-05
Fix a few warnings in mac and other builds

phlptp:
### Summary

If merged this pull request will attempt to fix a few code warnings in the different builds for HELICS 3


phlptp:
Address #1598 among other things

nightlark (approved):

nightlark:
Removes Python swig interface things from CMake -- not really an api change, but a big difference from v2 if anyone was building the swig Python interface.

phlptp:
I think we had removed it earlier.  This just got rid of the CMAKE variable and the relics in the test yaml files.

PR #1727, opened 2020-12-10
Header generation

phlptp:
### Summary

If merged this pull request will generate a single header for the C shared library.   and an ancillary one for api generation.
The PR also adjust the various tests and projects that use the library to use the new header.


kdheepak (approved):

kdheepak:
This is going to be a breaking change, so we should make sure it is documented and communicated clearly before release.

nightlark (approved):

PR #1739, opened 2021-01-27
Pin Linux release build Docker base image version to hbb 2.1.0

nightlark:
### Summary

If merged this pull request will pin the base docker image builds to hbb 2.1.0, one of the last ones using CentOS 6, and ensures a checkout step runs regardless of what triggered the workflow run. Workflow file change needs to get made to the master branch, pinning the docker image will need to make its way to the develop/helics3 branches. This PR will fix the nightly builds once merged into master+develop.

Latest test run: https://github.com/GMLC-TDC/HELICS/actions/runs/516313478

### Proposed changes

* Pin base docker image to HBB 2.1.0
* Run event ref checkout step for anything that isn't a scheduled build. If the ref (from client_payload) is set to nothing, then it defaults to checking out the current branch for the event.

### DIFF
--- a/.github/actions/linux-release-builder/Dockerfile
+++ b/.github/actions/linux-release-builder/Dockerfile
@@ -1,4 +1,4 @@
-FROM phusion/holy-build-box-64:2.2.0
+FROM phusion/holy-build-box-64:2.1.0
 
 # Install a copy of git and the Boost headers
 # curl -L follows redirects and -k ignores SSL certificate warnings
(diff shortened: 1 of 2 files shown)


nightlark:
Formatting check failure is expected; the develop branch fixes the formatting workflow to use environment files instead of the old insecure way of setting `PY` using `::set-env` to stdout.

phlptp (approved):

PR #1757, opened 2021-03-17
Install Boost 1.74 on Windows Azure Pipelines builds

nightlark:
### Summary


phlptp (approved):

phlptp:
I presume we need to do the same for HELICS 3?

nightlark:
Same thing with HELICS 3 (along with the other CI related changes) and the release package build workflow.

PR #1790, opened 2021-03-22
Install Boost for Windows GitHub Actions Workflows

nightlark:
### Summary

If merged this pull request will install the Boost 1.74 headers needed to build HELICS for the GitHub Actions workflows running on Windows -- release builds and benchmark installers.

This addresses the same issue of Boost getting removed from the virtual environments as Azure Pipelines had.

Benchmarks run: https://github.com/GMLC-TDC/HELICS/actions/runs/680686297
Releases run: https://github.com/GMLC-TDC/HELICS/actions/runs/680686725

### Proposed changes

- Install Boost 1.74 for Windows release build packaging and benchmark installer GitHub Actions workflows
- Comment and minor clean-up of the scripts used for the Windows GHA workflows

### DIFF
--- a/.github/actions/benchmark-build/benchmark-Windows.sh
+++ b/.github/actions/benchmark-build/benchmark-Windows.sh
@@ -12,13 +12,14 @@ cpack_dir="${cpack_dir%/cmake}"
 # Install Boost
 BOOST_ROOT="/c/boost"
 BOOST_URL="https://sourceforge.net/projects/boost/files/boost/1.74.0/boost_1_74_0.tar.bz2/download"
-cd $(mktemp -d)
-curl --location --output "download.tar.bz2" "$BOOST_URL"
-tar xfj "download.tar.bz2"
-mkdir -p "$BOOST_ROOT"
-cp -r boost_*/* "$BOOST_ROOT"
+(
+    cd $(mktemp -d) || exit
+    curl --location --output "download.tar.bz2" "$BOOST_URL"
+    tar xfj "download.tar.bz2"
+    mkdir -p "$BOOST_ROOT"
+    cp -r boost_*/* "$BOOST_ROOT"
+) || exit
 export BOOST_ROOT
-cd -
 
 # Build
 mkdir build && cd build || exit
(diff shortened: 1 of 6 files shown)


phlptp:
Is this related to the azure stuff?

.github/actions/benchmark-build/benchmark-Windows.sh
```diff
@@ -4,13 +4,27 @@
 # 2. getting the cmake directory for running cpack with an absolute path (chocolatey has an unfortunately named alias)
 
 echo "Building ${CPACK_GEN} installer with ${BUILD_GEN} for ${BUILD_ARCH}"
-BOOST_ROOT="${BOOST_ROOT_1_72_0}"
+
```
phlptp, line 7: This looks pretty similar in all the files.  can we put it a script so it easier to change the next time we update boost?

nightlark:
> Is this related to the azure stuff?

Yes, the virtual environments used for Azure Pipelines seem to be similar (the same?) as the ones for GitHub Actions. The repo for generating the Azure environments (https://github.com/microsoft/azure-pipelines-image-generation) just links to the GitHub Actions vm environment repo now.

.github/actions/benchmark-build/benchmark-Windows.sh
nightlark: It should be safe to assume that the scripts are never run using a symlink?

phlptp: I think my biggest concern is a lot of duplication and having to update the boost version in quite a few spots.  If we can reduce that in some way I think would be good

phlptp (approved):

.github/actions/benchmark-build/benchmark-Windows.sh
nightlark: Boost installation is split into a shared script. The way I'm getting the path won't work if the script using the shared code is symlinked, but being located along a path with a symlinked directory should be okay.

PR #1793, opened 2021-03-23
add the option of ordered queries

phlptp:
which are slower but have more guarantees of order since they travel on the normal priority paths
This is to fix some occasional failures in the app tests related to cloning which I think were caused by the queries getting to the destination before the other messages on occasion since they run on the priority channels.  This could potentially cause some missed endpoints to be loaded for some of the clones.

### DIFF
--- a/src/helics/application_api/CoreApp.cpp
+++ b/src/helics/application_api/CoreApp.cpp
@@ -209,7 +209,7 @@ const std::string& CoreApp::getAddress() const
 
 /** make a query at the core*/
 std::string
-    CoreApp::query(const std::string& target, const std::string& queryStr, query_synch_mode mode)
+    CoreApp::query(const std::string& target, const std::string& queryStr, helics_query_mode mode)
 {
     return (core) ? core->query(target, queryStr, mode) : std::string("#error");
 }
--- a/src/helics/application_api/CoreApp.cpp
+++ b/src/helics/application_api/CoreApp.cpp
@@ -208,7 +208,8 @@ const std::string& CoreApp::getAddress() const
 }
 
 /** make a query at the core*/
-std::string CoreApp::query(const std::string& target, const std::string& queryStr, query_synch_mode mode)
+std::string
+    CoreApp::query(const std::string& target, const std::string& queryStr, query_synch_mode mode)
 {
     return (core) ? core->query(target, queryStr, mode) : std::string("#error");
 }
--- a/src/helics/application_api/CoreApp.cpp
+++ b/src/helics/application_api/CoreApp.cpp
@@ -208,9 +208,9 @@ const std::string& CoreApp::getAddress() const
 }
 
 /** make a query at the core*/
-std::string CoreApp::query(const std::string& target, const std::string& queryStr)
+std::string CoreApp::query(const std::string& target, const std::string& queryStr, query_synch_mode mode)
 {
-    return (core) ? core->query(target, queryStr) : std::string("#error");
+    return (core) ? core->query(target, queryStr, mode) : std::string("#error");
 }
 
 void CoreApp::setGlobal(const std::string& valueName, const std::string& value)
--- a/src/helics/application_api/CoreApp.hpp
+++ b/src/helics/application_api/CoreApp.hpp
@@ -108,8 +108,8 @@ class HELICS_CXX_EXPORT CoreApp {
     @param target the target of the query "federation", "parent", "core","broker" or a specific
     named object
     @param queryStr the query to make
-    @param mode defaults to fast (asynchronous) meaning the query goes into priority channels, ordered
-    (synchronous) means slower queries but has more ordering guarantees
+    @param mode defaults to fast (asynchronous) meaning the query goes into priority channels,
+    ordered (synchronous) means slower queries but has more ordering guarantees
     @return a string containing the query results
     */
     std::string query(const std::string& target,
--- a/src/helics/application_api/CoreApp.hpp
+++ b/src/helics/application_api/CoreApp.hpp
@@ -114,7 +114,7 @@ class HELICS_CXX_EXPORT CoreApp {
     */
     std::string query(const std::string& target,
                       const std::string& queryStr,
-                      query_synch_mode mode = helics_query_mode_fast);
+                      helics_query_mode mode = helics_query_mode_fast);
 
     /** set a federation global value
     @details this overwrites any previous value for this name
--- a/src/helics/application_api/CoreApp.hpp
+++ b/src/helics/application_api/CoreApp.hpp
@@ -108,9 +108,11 @@ class HELICS_CXX_EXPORT CoreApp {
     @param target the target of the query "federation", "parent", "core","broker" or a specific
     named object
     @param queryStr the query to make
+    @param mod default to asynchronous meaning the query goes into priority channels, synchronous means slower queries but has more ordering guarantees
     @return a string containing the query results
     */
-    std::string query(const std::string& target, const std::string& queryStr);
+    std::string
+        query(const std::string& target, const std::string& queryStr, query_synch_mode mode=helics_query_mode_fast);
 
     /** set a federation global value
     @details this overwrites any previous value for this name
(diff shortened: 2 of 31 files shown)


src/helics/application_api/CoreApp.cpp
```diff
@@ -208,9 +208,10 @@ const std::string& CoreApp::getAddress() const
 }
 
 /** make a query at the core*/
-std::string CoreApp::query(const std::string& target, const std::string& queryStr)
+std::string
+    CoreApp::query(const std::string& target, const std::string& queryStr, query_synch_mode mode)
```
nightlark, line 212:
```suggestion
    CoreApp::query(const std::string& target, const std::string& queryStr, query_sync_mode mode)
```

src/helics/helics_enums.h
```diff
@@ -305,6 +305,13 @@ typedef enum {
 
 } helics_filter_type;
 
+/** enumeration of sequencing modes for queries
+fast is the default meaning the query travels along priority channels and takes precedence of over
+existing messages ordered means it follows normal priority patterns and will be ordered along with
+existing messages
+*/
+typedef enum { helics_query_mode_fast = 0, helics_query_mode_ordered = 1 } query_synch_mode;
```
nightlark, line 313:
```suggestion
typedef enum { helics_query_mode_fast = 0, helics_query_mode_ordered = 1 } query_sync_mode;
```

Thoughts on renaming this to `sync` instead of `synch`? I think `sync` is the more common abbreviation.

src/helics/helics_enums.h
```diff
@@ -305,6 +305,13 @@ typedef enum {
 
 } helics_filter_type;
 
+/** enumeration of sequencing modes for queries
+fast is the default meaning the query travels along priority channels and takes precedence of over
+existing messages ordered means it follows normal priority patterns and will be ordered along with
+existing messages
```
nightlark, line 311:
```suggestion
/** enumeration of sequencing modes for queries
fast is the default, meaning the query travels along priority channels and takes precedence of over
existing messages; ordered means it follows normal priority patterns and will be ordered along with
existing messages
```

src/helics/core/Core.hpp
```diff
@@ -718,10 +718,12 @@ class Core {
     target can also be "global" to query a global value stored in the broker
     @param target the specific target of the query
     @param queryStr the actual query
+    @param mode
```
nightlark, line 721:
```suggestion
    @param mode the synchronization mode for the query
```

src/helics/application_api/CoreApp.hpp
```diff
@@ -108,9 +108,13 @@ class HELICS_CXX_EXPORT CoreApp {
     @param target the target of the query "federation", "parent", "core","broker" or a specific
     named object
     @param queryStr the query to make
+    @param mod default to asynchronous meaning the query goes into priority channels, synchronous
+    means slower queries but has more ordering guarantees
```
nightlark, line 112:
```suggestion
    @param mode defaults to fast (asynchronous) meaning the query goes into priority channels, ordered
    (synchronous) means slower queries but has more ordering guarantees
```

src/helics/application_api/Federate.hpp
```diff
@@ -314,19 +314,23 @@ class HELICS_CXX_EXPORT Federate {
...
+                      query_synch_mode mode = helics_query_mode_fast);
 
     /** make a query of the core
     @details this call is blocking until the value is returned which make take some time depending
     on the size of the federation and the specific string being queried
     @param queryStr a string with the query see other documentation for specific properties to
     query, can be defined by the federate if the local federate does not recognize the query it
     sends it on to the federation
+    @param mode asynchronous(default) means the query goes on priority channels, synchronous is
+    slower but has more ordering guarantees
```
nightlark, line 328:
```suggestion
    @param mode fast (asynchronous; default) means the query goes on priority channels, ordered (synchronous) is
    slower but has more ordering guarantees
```

src/helics/application_api/Federate.hpp
```diff
@@ -335,9 +339,13 @@ class HELICS_CXX_EXPORT Federate {
     specific name of a federate, core, or broker
     @param queryStr a string with the query see other documentation for specific properties to
     query, can be defined by the federate
+    @param mode asynchronous(default) means the query goes on priority channels, synchronous is
+    slower but has more ordering guarantees
```
nightlark, line 343: @param mode fast (asynchronous; default) means the query goes on priority channels, ordered (synchronous) is
    slower but has more ordering guarantees

src/helics/core/Broker.hpp
```diff
@@ -95,7 +95,9 @@ class Broker {
       @return a string containing the response to the query.  Query is a blocking call and will not
     return until the query is answered so use with caution
```
nightlark, line 96:
```suggestion
    @param mode fast (asynchronous; default) means the query goes on priority channels, ordered (synchronous) is
    slower but has more ordering guarantees
      @return a string containing the response to the query.  Query is a blocking call and will not
    return until the query is answered so use with caution
```

src/helics/application_api/Federate.hpp
```diff
@@ -314,19 +314,23 @@ class HELICS_CXX_EXPORT Federate {
     or a JSON string stored in the first element of the vector.  The string "#invalid" is returned
```
nightlark, line 314: Doc comment is missing an `@param mode` entry.

nightlark (commented):
I noticed some places are using a `mode` argument with the enum and others use a `synchronous` boolean arg -- is the division roughly user API functions vs internal? Are there future enums types that could be added?

src/helics/helics_enums.h
phlptp: I am renaming it to helics_query_mode,  I was originally running async vs synchronous but as I got further down that made a little less sense,  I am also adding it to the C++98 and C interfaces.

src/helics/core/CommonCore.cpp
```diff
@@ -1992,7 +1998,8 @@ void CommonCore::loadBasicJsonInfo(
 
 void CommonCore::initializeMapBuilder(const std::string& request,
                                       std::uint16_t index,
-                                      bool reset) const
+                                      bool reset,
+                                      bool synchronous) const
```
phlptp, line 2002: I am going to rename the synchronous to something else that makes a little more sense now

src/helics/cpp98/Broker.hpp
```diff
@@ -144,7 +144,8 @@ class Broker {
   specific name of a federate, core, or broker
   @param queryStr a string with the query, see other documentation for specific properties to
   query, can be defined by the federate
-  @param mode the ordering mode to use for the query (fast-priority channels, ordered for normal channels ordered with all other messages)
+  @param mode the ordering mode to use for the query (fast-priority channels, ordered for normal
```
nightlark, line 147:
```suggestion
  @param mode the ordering mode to use for the query (fast for priority channels, ordered for normal
```

src/helics/cpp98/Federate.hpp
```diff
@@ -465,14 +465,21 @@ class Federate {
     specific name of a federate, core, or broker
     @param queryStr a string with the query see other documentation for specific properties to
     query, can be defined by the federate
+    @param mode the ordering mode to use for the query (fast-priority channels, ordered for normal
```
nightlark, line 468:
```suggestion
    @param mode the ordering mode to use for the query (fast for priority channels, ordered for normal
```

src/helics/cpp98/Federate.hpp
```diff
@@ -485,14 +492,20 @@ class Federate {
 
     @param queryStr a string with the query, see other documentation for specific properties to
     query, can be defined by the federate
+    @param mode the ordering mode to use for the query (fast-priority channels, ordered for normal
```
nightlark, line 495:
```suggestion
    @param mode the ordering mode to use for the query (fast for priority channels, ordered for normal
```

src/helics/shared_api_library/helicsExport.cpp
```diff
@@ -979,6 +979,15 @@ void helicsQuerySetQueryString(helics_query query, const char* queryString, heli
     queryObj->query = AS_STRING(queryString);
 }
 
+void helicsQuerySetOrdering(helics_query query, int32_t mode, helics_error* err)
```
nightlark, line 982: For the mode would we want to use the enum typedef instead of `int32_t`?

phlptp: I am not sure I want enum in the definitions for the C interface.  That can lead to type difficulties.

nightlark (approved):

PR #1804, opened 2021-03-25
Event triggered flag

phlptp:
### Summary

If merged this pull request will add an event triggered flag to the timing.  The flag is used by filters to adjust the timing slightly, this will hopefully resolve some lingering issues with timing in some tests on the CI systems.


phlptp:
If this resolves some of the tests I will add full support for the flag

nightlark (approved):

PR #1825, opened 2021-04-07
Broker server tests

phlptp:
### Summary

If merged this pull request will fix broken NO_ZMQ tests and fix some issues with broker_server

### DIFF
--- a/src/helics/apps/helics-broker-server.cpp
+++ b/src/helics/apps/helics-broker-server.cpp
@@ -22,11 +22,11 @@ void terminalFunction(std::vector<std::string> args);
 
 int main(int argc, char* argv[])
 {
-    int ret = 0;
-    bool runterminal = false;
+    int ret{0};
+    bool runterminal{false};
 
     helics::helicsCLI11App cmdLine("helics broker server command line");
-    auto term =
+    auto* term =
         cmdLine
             .add_subcommand("term",
                             "helics_broker_server term will start a broker server "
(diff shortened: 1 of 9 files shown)


src/helics/apps/helics-broker-server.cpp
```diff
@@ -202,11 +218,13 @@ void terminalFunction(std::vector<std::string> args)
...
         ->callback(closeBrokerServer);
-    termProg.add_subcommand("broker", "create a new broker with the given arguments")
+    auto brokersub=termProg.add_subcommand("broker", "create a new broker with the given arguments")
         ->callback(newBroker);
 
+    brokersub->add_option("args", bargs, "arguments for the query");
+
     termProg
-        .add_subcommand("terminate!",
+        .add_subcommand("force_terminate",
```
phlptp, line 227: the `!terminate` is not a valid CLI command name

src/helics/core/BrokerFactory.hpp
```diff
@@ -78,6 +78,14 @@ namespace BrokerFactory {
 @param brokerName the name of the broker
 @return a shared_ptr to the Broker*/
     std::shared_ptr<Broker> findBroker(const std::string& brokerName);
+    /** get the first available broker
+@param brokerName the name of the broker
```
nightlark, line 82:
```suggestion
```

src/helics/core/BrokerFactory.cpp
```diff
@@ -184,7 +184,34 @@ thread which allows the destructor to fire if need be without issue*/
...
+            return brk;
+        }
+        if (brokerName.empty()) {
+            return findBroker();
+        }
+        if (brokerName.front() == '#') {
+            return findBroker(brokerName.substr(1));
+        }
+        try {
+            auto val = std::stoull(brokerName);
```
nightlark, line 198: Users should probably avoid using findBroker if their goal is to get a broker by index. A broker given a numeric name would prevent finding a broker with the matching numeric index, and a search for a broker named "123abc" or "123xyz" would both return a broker at index 123.

phlptp: I am going to change the names of the functions,  but there needed to be a way to get a broker by index from the string to handle some situations in the webserver.

nightlark (approved):

PR #1832, opened 2021-04-09
Try a flush query

phlptp:
to force the registration for the clone and recorder filters to resolve sporadic failure

### Summary

If merged this pull request will add a global_flush query that will flush the system and try to resolve the sporadic failures on the apps.

### DIFF
--- a/src/helics/core/FilterFederate.hpp
+++ b/src/helics/core/FilterFederate.hpp
@@ -23,6 +23,7 @@ SPDX-License-Identifier: BSD-3-Clause
 #include <string>
 #include <utility>
 #include <vector>
+#include <deque>
 
 namespace helics {
 class HandleManager;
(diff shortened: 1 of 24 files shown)


phlptp:
@nightlark, take a look at this it does fix some things but still haven't found the key to 100% resolution on the tests yet.  But I want to get this merged and we need to get a release out

src/helics/core/FilterFederate.hpp
```diff
@@ -112,8 +114,9 @@ class FilterFederate {
     void processFilterInfo(ActionMessage& command);
 
     ActionMessage& processMessage(ActionMessage& command, const BasicHandleInfo* handle);
-
-    void destinationProcessMessage(ActionMessage& command, const BasicHandleInfo* handle);
+    /** process destination filters on the message return true if the original command should be
```
nightlark, line 117:
```suggestion
    /** process destination filters on the message and return true if the original command should be
```

src/helics/core/TimeCoordinator.hpp
```diff
@@ -96,7 +96,8 @@ class TimeCoordinator {
   private:
     std::atomic<int32_t> iteration{0};  //!< iteration counter
     bool disconnected{false};
-
+    bool nonGranting{false};  // specify the that the timeCoordinator should not grant times and
```
nightlark, line 99:
```suggestion
    bool nonGranting{false};  // specify that the timeCoordinator should not grant times and
```

nightlark (approved):

PR #1835, opened 2021-04-15
Add vcpkg manifest file

nightlark:
### Summary

If merged this pull request will enable vcpgk to be used as a dependency manager for HELICS dependencies when building from source.

### Proposed changes

- Adds vcpkg.json manifest file
- Add some info on using vcpkg to manage dependencies to docs on building HELICS from source

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -17,12 +17,10 @@ else()
     cmake_policy(VERSION 3.18)
 endif()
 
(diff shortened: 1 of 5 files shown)


nightlark:
Docker Hub is having problems. Fetching images is returning 503 service unavailable.

phlptp (approved):
Do we want to have a test of this running somewhere at some point?

nightlark:
At some point -- probably as a daily build; using vcpkg to install Boost.Beast for the webserver component takes a while.

PR #1845, opened 2021-04-16
Default HELICS_USE_ZMQ_STATIC_LIBRARY=ON if only the static zmq library is found

nightlark:
### Summary

If merged this pull request will default the `HELICS_USE_ZMQ_STATIC_LIBRARY` to `ON` if only the static zmq library is found. Otherwise, it will continue to default to `OFF`.

### Proposed changes

- Default HELICS_USE_ZMQ_STATIC_LIBRARY=ON if only the static zmq library is found

### DIFF
--- a/config/cmake/addZeroMQ.cmake
+++ b/config/cmake/addZeroMQ.cmake
@@ -113,8 +113,10 @@ else()
                 show_variable(ZeroMQ_LIBRARY FILEPATH "path to the ZeroMQ library" "")
                 show_variable(ZeroMQ_ROOT_DIR PATH "path to the ZeroMQ root directory"
                               "")
-                show_variable(ZeroMQ_STATIC_LIBRARY FILEPATH
-                              "path to the ZeroMQ static library" "")
+                if(${PROJECT_NAME}_USE_ZMQ_STATIC_LIBRARY)
+                    show_variable(ZeroMQ_STATIC_LIBRARY FILEPATH
+                                  "path to the ZeroMQ static library" "")
+                endif()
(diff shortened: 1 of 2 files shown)


config/cmake/addZeroMQ.cmake
```diff
@@ -106,10 +113,8 @@ else()
                 show_variable(ZeroMQ_LIBRARY FILEPATH "path to the ZeroMQ library" "")
                 show_variable(ZeroMQ_ROOT_DIR PATH "path to the ZeroMQ root directory"
                               "")
-                if(${PROJECT_NAME}_USE_ZMQ_STATIC_LIBRARY)
-                    show_variable(ZeroMQ_STATIC_LIBRARY FILEPATH
-                                  "path to the ZeroMQ static library" "")
-                endif()
+                show_variable(ZeroMQ_STATIC_LIBRARY FILEPATH
+                              "path to the ZeroMQ static library" "")
```
phlptp, line 117: what is the reason for the change here?

nightlark: I think it was the idea that if they are manually pointing CMake to a zmq library then either the shared or static library is okay.

Though if a zmq library isn't found on the first try, the option to use static library will continue to default to off, so maybe this should remain hidden unless they turn the use static library option on.

phlptp: I think my preference would be to minimize the number of cmake fields showing in the GUI,  but I will leave it up to you.

phlptp (approved):

PR #1852, opened 2021-04-20
Make some tests and fixes for multiple filters on a single endpoint

phlptp:
### Summary

If merged this pull request will make some more tests and fixes for filter operations, particularly with multiple stacked filters

### Proposed changes

 change name of ordering property enumerations
add multi_filter tests
fixes to make multi filters tests pass


phlptp:
@nightlark,  This fixes some problems.  Apparently not all however.

tests/helics/application_api/FilterTests.cpp
```diff
@@ -1147,6 +1304,125 @@ TEST_F(filter_tests, reroute_separate2_5message)
     broker->waitForDisconnect();
 }
 
+TEST_F(filter_tests, reroute_separate2_5000message_ci_skip)
```
nightlark, line 1301: Does this test take a while to run?

phlptp: not that long,  just didn't see a reason to make it run on the CI tests.  Was trying to get it to fail on my machine and didn't succeed, so this test might be able to be removed if desired.

nightlark (approved):

PR #1864, opened 2021-04-23
Fix docs markdown linting errors

nightlark:
### Summary

If merged this pull request will fix markdown linting errors making the formatting job fail.

### Proposed changes

- Fix code block fences without a language errors
- Fix mismatched bulleted list style errors
- Fix multiple top-level header indentation errors
- Fix multiple headers with the same name


phlptp (approved):

docs/user-guide/examples/advanced_examples/advanced_brokers_hierarchies.md
```diff
@@ -48,16 +48,15 @@ When running across multiple compute nodes, the relevant portion of the helics_c
 
 `broker_hierarchy_runner_A.json`
 
-```
+```json
```
trevorhardy, line 51: This isn't valid JSON; why do we need to add the JSON tag?

trevorhardy (commented):
There are numerous occasions when the "json" tag is added to a fragment of a JSON (that is, even when it is not valid JSON on its own). Is this correct rather than just leaving it a generic code block?

trevorhardy:
Also, there were several places where the Markdown list delimiter was changed from "*" to "-". What is motivating this change; aren't both valid Markdown?

docs/user-guide/examples/advanced_examples/advanced_brokers_hierarchies.md
nightlark: The markdown linter wants all code blocks to have a language given. Since its an excerpt from a JSON file, JSON seems to be a better match than other choices like `shell`.

For reference, the languages supported by GitHub markdown are: https://www.rubycoloredglasses.com/2013/04/languages-supported-by-github-flavored-markdown/

nightlark:
> 
> 
> Also, there were several places where the Markdown list delimiter was changed from "*" to "-". What is motivating this change; aren't both valid Markdown?

The linter gives an [MD004 unordered list style](https://github.com/markdownlint/markdownlint/blob/master/docs/RULES.md#md004---unordered-list-style) error. I think its because the list is being inconsistent in which style it uses; sticking to the same style  for all bullet points in a list seems to ameliorate it.

nightlark:
Here's the failed formatting job with a list of all the markdown linting errors: https://github.com/GMLC-TDC/HELICS/runs/2395676252

The various markdown rules: https://github.com/markdownlint/markdownlint/blob/master/docs/RULES.md#rules

docs/user-guide/examples/advanced_examples/advanced_brokers_hierarchies.md
nightlark: Another idea was that the `...` before and after the JSON fragment could be removed since those are never valid lines in a JSON file. Earlier on the docs say `the relevant portion of the helics_cli runner files would look like this` so readers should be able to tell that it is just a single line from a larger file.

trevorhardy (approved):

docs/user-guide/examples/advanced_examples/advanced_brokers_hierarchies.md
kdheepak: You can use `plaintext`:

````
```plaintext
This code will not be syntax highlighted.
```
````

PR #1866, opened 2021-04-27
Try some additional tests and fixes for sporadic timing bug

phlptp:
### Summary

If merged this pull request will clean up some unused code and put a few more timing restrictions on filters.  This will need to be updated in the future but it is a tiny step in the right direction.


nightlark (commented):
Is removing the shared_library test files intentional?

phlptp:
yeah, they were the old files using boost yet,  they were in there as a place to just copy some tests over, but that for the most part has been done, and I don't think we will be doing any more.  @kdheepak  noted something with the files, and I decided it was time to remove them since I they won't have a use any time soon, and we can just copy over the new ones if needed.

nightlark (approved):

PR #1870, opened 2021-04-29
add a global status query

phlptp:
### Summary

If merged this pull request will add a global status query

### DIFF
--- a/src/helics/core/CoreBroker.cpp
+++ b/src/helics/core/CoreBroker.cpp
@@ -2500,7 +2500,8 @@ enum subqueries : std::uint16_t {
     version_all = 5,
     global_state = 6,
     global_time_debugging = 7,
-    global_flush = 8
+    global_flush = 8,
+    global_status = 9
 };
 
 static const std::map<std::string, std::pair<std::uint16_t, bool>> mapIndex{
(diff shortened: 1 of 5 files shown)


src/helics/core/CoreBroker.cpp
```diff
@@ -2703,6 +2716,43 @@ std::string CoreBroker::generateQueryAnswer(const std::string& request, bool for
...
+        Json::Value v;
+        v["status"] = "error";
+        v["timestep"] = -1;
+        return generateJsonString(v);
+    }
+
+    Time mv{Time::maxVal()};
+    if (!builder.getJValue()["cores"][0].isObject()) {
+        state = "init_requested";
+    }
```
nightlark, line 2739: I'm not seeing where `state` is getting used after it gets used here -- is this needed or should there be a condition wrapping the subsequent for loop that checks state?

src/helics/core/CoreBroker.cpp
```diff
@@ -2703,6 +2716,43 @@ std::string CoreBroker::generateQueryAnswer(const std::string& request, bool for
...
+    for (auto& cr : builder.getJValue()["cores"]) {
+        for (auto fed : cr["federates"]) {
+            auto dv = fed["granted_time"].asDouble();
+            if (dv < mv) {
+                mv = dv;
+            }
+        }
+    }
+    std::string tste = (mv >= timeZero) ? std::string("connected") : std::string("init_requested");
```
nightlark, line 2748: Also not seeing `tste` used anywhere in this function.

nightlark (approved):

PR #1881, opened 2021-05-03
Add filter federate example documentation

trevorhardy:
### Summary

Adds documentation for filter federate example as a part of the User Guide for v3.

Also adds a template for the native filter write-up.


phlptp:
looks like a few formatting issues need to be resolved, then I think we could merge this assuming the actual content is good to go.

nightlark:
Okay, so the workflow for compressing images has a condition for pushing optimized images inverted. https://github.com/GMLC-TDC/HELICS/pull/1889 fixes it, I think if this PR is rebased after that gets merged then only the compressed images will make it into the repository history.

nightlark:
Looks like there are quite a few images that were never optimized. Future PRs with images added should just result in the new images getting compressed.

phlptp:
@trevorhardy  unless you tell me otherwise I am going to merge this, and we can polish it up if needed between the beta and final

trevorhardy:
I won't have time to get to this for a few more days so go ahead.

phlptp (approved):

Issue #1880, opened 2021-05-03
Test fails in helics 2.7.0 but passes in 2.6.1


kdheepak:
The `filtered_endpoints` query seems to be failing for me now.

```python
    def test_filter_test_types_clone_test_connections():

        broker = createBroker(3)
        sFed, fedinfo1 = createMessageFederate(1, "source", 1.0,)
        dFed, fedinfo2 = createMessageFederate(1, "dest", 1.0)
        dcFed, fedinfo3 = createMessageFederate(1, "dest_clone", 1.0)

        p1 = h.helicsFederateRegisterGlobalEndpoint(sFed, "src", "")
        p2 = h.helicsFederateRegisterGlobalEndpoint(dFed, "dest", "")
        p3 = h.helicsFederateRegisterGlobalEndpoint(dcFed, "cm", "")

        f1 = h.helicsFederateRegisterGlobalCloningFilter(dcFed, "filt1")
        h.helicsFilterAddDeliveryEndpoint(f1, "cm")
        cr = h.helicsFederateGetCoreObject(sFed)

        h.helicsCoreAddSourceFilterToEndpoint(cr, "filt1", "src")
        h.helicsCoreAddSourceFilterToEndpoint(cr, "", "src")

        h.helicsFederateEnterExecutingModeAsync(sFed)
        h.helicsFederateEnterExecutingModeAsync(dcFed)
        h.helicsFederateEnterExecutingMode(dFed)
        h.helicsFederateEnterExecutingModeComplete(sFed)
        h.helicsFederateEnterExecutingModeComplete(dcFed)

        q = h.helicsCreateQuery("", "filtered_endpoints")
        filteredEndpoints = h.helicsQueryExecute(q, sFed)
>       assert "srcFilters" in str(filteredEndpoints)
E       assert 'srcFilters' in "{'endpoints': [], 'id': 1879048192, 'name': 'Testsource'}"
E        +  where "{'endpoints': [], 'id': 1879048192, 'name': 'Testsource'}" = str({'endpoints': [], 'id': 1879048192, 'name': 'Testsource'})
```

Am I using this incorrectly?

kdheepak:
I'm trying to replicate this test:

https://github.com/GMLC-TDC/HELICS/blob/c4f8bcdbd0b0e25a78bd03f0c352f1cce0bce1d0/tests/helics/shared_library/FilterTests.cpp#L586-L626

kdheepak:
[...]

PR #1882, opened 2021-05-04
Interface tags

phlptp:
### Summary

If merged this pull request will add a tags interface to HELICS 3 supporting user defined key value pairs associated with interfaces.  

### Proposed changes

-  Fix an issue with endpoint subscription data
-  add tags
- add tests for tags
- add interface query which can retrieve tag information
- add misc test based on #1880

### DIFF
--- a/src/helics/core/BasicHandleInfo.hpp
+++ b/src/helics/core/BasicHandleInfo.hpp
@@ -64,11 +64,11 @@ class BasicHandleInfo {
     InterfaceHandle getInterfaceHandle() const { return handle.handle; }
     /** extract a global federate id */
     GlobalFederateId getFederateId() const { return handle.fed_id; }
-    /** set the user level information field*/
+    /** set a tag (key-value pair)*/
     void setTag(const std::string& tag, const std::string& value);
     /** search for a tag by name*/
     const std::string& getTag(const std::string& tag) const;
-    /** get a tag (key, value) pair by index*/
+    /** get a tag (key-value pair) by index*/
     const std::pair<std::string, std::string>& getTagByIndex(size_t index) const
     {
         return tags[index];
@@ -77,6 +77,6 @@ class BasicHandleInfo {
     auto tagCount() const { return tags.size(); }
 
   private:
-    std::vector<std::pair<std::string, std::string>> tags;  //!< storage for a user info string
+    std::vector<std::pair<std::string, std::string>> tags;  //!< storage for user defined tags
 };
 }  // namespace helics
--- a/src/helics/core/BasicHandleInfo.hpp
+++ b/src/helics/core/BasicHandleInfo.hpp
@@ -10,8 +10,8 @@ SPDX-License-Identifier: BSD-3-Clause
 #include "flagOperations.hpp"
 
 #include <string>
-#include <vector>
 #include <utility>
+#include <vector>
 
 namespace helics {
 
@@ -69,10 +69,14 @@ class BasicHandleInfo {
     /** search for a tag by name*/
     const std::string& getTag(const std::string& tag) const;
     /** get a tag (key, value) pair by index*/
-    const std::pair<std::string, std::string>& getTagByIndex(size_t index) const { return tags[index]; }
+    const std::pair<std::string, std::string>& getTagByIndex(size_t index) const
+    {
+        return tags[index];
+    }
     /** get the number of tags associated with an interface*/
     auto tagCount() const { return tags.size(); }
+
   private:
-    std::vector<std::pair<std::string,std::string>> tags;  //!< storage for a user info string
+    std::vector<std::pair<std::string, std::string>> tags;  //!< storage for a user info string
 };
 }  // namespace helics
--- a/src/helics/core/BasicHandleInfo.hpp
+++ b/src/helics/core/BasicHandleInfo.hpp
@@ -10,6 +10,8 @@ SPDX-License-Identifier: BSD-3-Clause
 #include "flagOperations.hpp"
 
 #include <string>
+#include <vector>
+#include <utility>
 
 namespace helics {
 
@@ -56,7 +58,6 @@ class BasicHandleInfo {
     const std::string key;  //!< the name of the handle
     const std::string type;  //!< the type of data used by the handle
     const std::string units;  //!< the units associated with the handle
-    std::string interface_info;  //!< storage for a user info string
     const std::string& type_in;  //!< the input type of a filter
     const std::string& type_out;  //!< the output type of a filter
     /** get the interface handle information */
@@ -64,6 +65,14 @@ class BasicHandleInfo {
     /** extract a global federate id */
     GlobalFederateId getFederateId() const { return handle.fed_id; }
     /** set the user level information field*/
-    void setInfoField(std::string& info) { interface_info = info; }
+    void setTag(const std::string& tag, const std::string& value);
+    /** search for a tag by name*/
+    const std::string& getTag(const std::string& tag) const;
+    /** get a tag (key, value) pair by index*/
+    const std::pair<std::string, std::string>& getTagByIndex(size_t index) const { return tags[index]; }
+    /** get the number of tags associated with an interface*/
+    auto tagCount() const { return tags.size(); }
+  private:
+    std::vector<std::pair<std::string,std::string>> tags;  //!< storage for a user info string
 };
 }  // namespace helics
--- a/src/helics/core/CommonCore.hpp
+++ b/src/helics/core/CommonCore.hpp
@@ -258,11 +258,9 @@ class CommonCore: public Core, public BrokerBase {
      * may need a helper class of some sort*/
     virtual void processDisconnect(bool skipUnregister = false) override final;
 
-    /** set the `info` tag for the interface; `setInterfaceInfo(handle, "infostr")` is equivalent to
-     * `setTag(handle, "info", "infostr")`*/
+    /** set the local information field of the interface*/
     virtual void setInterfaceInfo(InterfaceHandle handle, std::string info) override final;
-    /** gets the `info` tag for the interface; `getInterfaceInfo(handle)` is equivalent to
-     * `getTag(handle, "info")`*/
+    /** get the local information field of the interface*/
     virtual const std::string& getInterfaceInfo(InterfaceHandle handle) const override final;
 
     virtual void setTag(InterfaceHandle handle,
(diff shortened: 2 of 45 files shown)


src/helics/shared_api_library/backup/helics.h
```diff
@@ -2995,6 +3016,27 @@ HELICS_EXPORT const char* helicsPublicationGetInfo(HelicsPublication pub);
  */
 HELICS_EXPORT void helicsPublicationSetInfo(HelicsPublication pub, const char* info, HelicsError* err);
 
+/**
+ * Get the data in a specified tag of a publication
```
nightlark, line 3020:
```suggestion
 * Get the data in a specified tag of a publication.
```

src/helics/shared_api_library/backup/helics.h
```diff
@@ -2975,6 +2975,27 @@ HELICS_EXPORT const char* helicsInputGetInfo(HelicsInput inp);
...
+ * @return A string with the tag data.
+ */
+HELICS_EXPORT const char* helicsInputGetTag(HelicsInput inp, const char* tagname);
+
+/**
+ * Set the data in a specific tag for an input.
+ *
+ * @param inp The input object to query.
+ * @param tagname The string to set.
+ * @param tagvalue the string value to associate with a tag
```
nightlark, line 2992:
```suggestion
 * @param tagvalue The string value to associate with a tag.
```

src/helics/shared_api_library/backup/helics.h
```diff
@@ -2975,6 +2975,27 @@ HELICS_EXPORT const char* helicsInputGetInfo(HelicsInput inp);
  */
 HELICS_EXPORT void helicsInputSetInfo(HelicsInput inp, const char* info, HelicsError* err);
 
+/**
+ * Get the data in a specified tag of an input
```
nightlark, line 2979:
```suggestion
 * Get the data in a specified tag of an input.
```

src/helics/core/BasicHandleInfo.hpp
```diff
@@ -56,14 +58,25 @@ class BasicHandleInfo {
...
+    /** get a tag (key, value) pair by index*/
+    const std::pair<std::string, std::string>& getTagByIndex(size_t index) const
+    {
+        return tags[index];
+    }
+    /** get the number of tags associated with an interface*/
+    auto tagCount() const { return tags.size(); }
+
+  private:
+    std::vector<std::pair<std::string, std::string>> tags;  //!< storage for a user info string
```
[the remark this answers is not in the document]
phlptp: I started with std::map,  could move back to it, but the likely use case is a fairly low number of tags 0 being common, maybe 4 in a couple cases.    so having a map in the object seemed a bit heavy for what it was doing.    One issue with making the json map is the need to loop over them to see what tags a particular interface has.  And there is no restriction on having json compatible field names.    So a vector seemed a bit more natural for the operations that are needed internally and is likely to be faster though speed is not likely to be a big issue here.

phlptp:
I thought about removing the setInfo operations but one difference is the info is just available locally and the tags can be queried.   It is just convenient to store them in the same structure since there are already 2 set of the BasicHandleInfo objects in a core one for operations purely in the core thread and one for threadsafe local operations.

src/helics/core/BasicHandleInfo.hpp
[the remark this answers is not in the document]
nightlark: The empty map and per item overheads seem like a good reason to keep using an std::vector.

If we're using a json parser that follows the standard then the restrictions on json compatible field names are identical to valid string values; any unicode codepoint except `"` or `\` or control characters (which must be escaped). It looks like toml quoted keys are similar.

For the serialized version, serializing it to a json map (jsoncpp `getMemberNames()`) and possibly an inline toml table (toml11 key-value iterator/foreach) would make it nicer for users editing the config files (less brackets and less verbose), and using query responses in other languages (i.e. javascript `response["tags"]["tagname"]` vs writing a custom array search function). The test files would become:

json:
```json
"tags": {
    "period":  "0.5" ,
    "description": "a test publication"
}
```

toml:
```toml
tags = {"period"="0.5", "description"="a test publication"}
```

nightlark:
> 
> 
> I thought about removing the setInfo operations but one difference is the info is just available locally and the tags can be queried. It is just convenient to store them in the same structure since there are already 2 set of the BasicHandleInfo objects in a core one for operations purely in the core thread and one for threadsafe local operations.

When the "info" tag is set using setInfo it can't be queried? Is the setInfo only useable from the BHI object used for threadsafe local operations?

phlptp:
The setInfo/getInfo command works exactly the same as before.  You can't get the information from another federate.

nightlark:
Okay. I'm not seeing how a collision between someone calling setTag("info") and setInfo() wouldn't happen, unless the setInfo() functions are only available when using something with a local-only BasicHandleInfo object, therefore making the setTag("info") call local-only as well.

phlptp:
should I make make the tag for setInfo something like "local_info_"    it would be extremely unlikely for that to collide?

nightlark:
Yea, I think using a name that's unlikely to be chosen would be good -- probably don't even need to mention it in the docstrings then, other than if we want to say not to set a tag with that specific name.

src/helics/core/BasicHandleInfo.hpp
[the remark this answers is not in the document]
phlptp: Ok I like that for formatting purposes somewhat better.  I think I still want to support the name value keys since then you specify things in a json schema if desired.  But supporting the general format as described above is probably better.  This would require putting some limitations on tag names that would need to be checked in the core and would generate an error if a tag string was not json compatible.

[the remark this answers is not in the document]
nightlark: Actually, here are some interesting things in jsoncpp:
- https://github.com/open-source-parsers/jsoncpp/blob/master/src/lib_json/json_writer.cpp#L280
- https://github.com/open-source-parsers/jsoncpp/blob/master/src/lib_json/json_writer.cpp#L493

It looks like the `valueToQuotedStringN` function is called by `valueToQuotedString`, which is used when writing string values and member names, so we might not need to do our own checking.

src/helics/core/queryHelpers.cpp
```diff
@@ -14,14 +14,16 @@ namespace helics {
...
         for (size_t ii = 0; ii < bhi.tagCount(); ++ii) {
             const auto& tg = bhi.getTagByIndex(ii);
-            Json::Value tagBlock;
-            tagBlock["name"] = tg.first;
-            tagBlock["value"] = tg.second;
-            v["tags"].append(tagBlock);
+            if (tg.second != "true") {
+                tagBlock[tg.first] = tg.second;
+            } else {
+                tagBlock[tg.first] = true;
```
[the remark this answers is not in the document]
phlptp: So you think I should just leave everything as strings in the output?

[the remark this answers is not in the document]
nightlark: I think so -- at least in JS the check to see if the tag is defined is the same whether its a string or boolean, and internally HELICS converts it to a string on load so it doesn't matter.

I'm not sure about consuming the json from other languages, but just telling them it will always be a string seems simpler rather than making a potential corner case that needs to be handled in a different way.

src/helics/core/CommonCore.hpp
```diff
@@ -258,9 +258,19 @@ class CommonCore: public Core, public BrokerBase {
      * may need a helper class of some sort*/
     virtual void processDisconnect(bool skipUnregister = false) override final;
 
+    /** set the `info` tag for the interface; `setInterfaceInfo(handle, "infostr")` is equivalent to
+     * `setTag(handle, "info", "infostr")`*/
```
nightlark, line 262:
```suggestion
    /** set the local information field of the interface*/
```

src/helics/core/CommonCore.hpp
```diff
@@ -258,9 +258,19 @@ class CommonCore: public Core, public BrokerBase {
...
     virtual void setInterfaceInfo(InterfaceHandle handle, std::string info) override final;
+    /** gets the `info` tag for the interface; `getInterfaceInfo(handle)` is equivalent to
+     * `getTag(handle, "info")`*/
```
nightlark, line 265:
```suggestion
    /** get the local information field of the interface*/
```

src/helics/cpp98/Input.hpp
```diff
@@ -193,13 +193,24 @@ class Input {
     const char* getType() const { return helicsInputGetType(inp); }
     /** get an associated target*/
     const char* getTarget() const { return helicsSubscriptionGetTarget(inp); }
-    /** get the interface information field of the filter*/
+    /** get the interface information field of the input; equivalent to `getTag("info")`*/
```
nightlark, line 196:
```suggestion
    /** get the interface information field of the input`*/
```

src/helics/cpp98/Input.hpp
```diff
@@ -193,13 +193,24 @@ class Input {
...
     const char* getInfo() const { return helicsInputGetInfo(inp); }
-    /** set the interface information field of the publication*/
+    /** set the interface information field of the input; `setInfo("infostr")` is equivalent to
+     * `setTag("info", "infostr")`*/
```
nightlark, line 199:
```suggestion
    /** set the interface information field of the input*/
```

src/helics/shared_api_library/MessageFederate.h
```diff
@@ -402,9 +402,9 @@ HELICS_EXPORT const char* helicsEndpointGetName(HelicsEndpoint endpoint);
 HELICS_EXPORT int helicsFederateGetEndpointCount(HelicsFederate fed);
 
 /**
- * Get the data in the info field of a filter.
+ * Get the data in the "info" tag of an endpoint.
```
nightlark, line 405:
```suggestion
 * Get the local information field of an endpoint.
```

src/helics/shared_api_library/MessageFederate.h
```diff
@@ -421,6 +421,27 @@ HELICS_EXPORT const char* helicsEndpointGetInfo(HelicsEndpoint end);
  */
 HELICS_EXPORT void helicsEndpointSetInfo(HelicsEndpoint endpoint, const char* info, HelicsError* err);
```
nightlark, line 422: The docstring for this on line 414 should be updated to say `endpoint` instead of `filter`

src/helics/shared_api_library/backup/helics.h
```diff
@@ -3477,9 +3524,9 @@ HELICS_EXPORT const char* helicsEndpointGetName(HelicsEndpoint endpoint);
 HELICS_EXPORT int helicsFederateGetEndpointCount(HelicsFederate fed);
 
 /**
- * Get the data in the info field of a filter.
+ * Get the data in the "info" tag of an endpoint.
```
nightlark, line 3527:
```suggestion
 * Get the local information field of an endpoint.
```

nightlark (commented):
Just a few changes to the docstrings due to the `info` to `local_info_` tag change.

nightlark (approved):

PR #1931, opened 2021-05-27
Query timeouts

phlptp:
### Summary

If merged this pull request will add timeouts for queries, In the rare case that a disconnect and query cross paths.  The query will timeout and result in a timeout result.  
The query timeout can be controlled via command line 

### Proposed changes

- add the command line option for query timeouts
- add timeout management code for local queries
- add tickforwarding controls in the broker base with reasons (and flexibility for adding more reasons in the future).  
- update the concurrency library with some new methods to the the delayedObject class.
- clean up the queries on disconnect from a core

### Notes
This is in response to some reported bugs about queries hanging

### DIFF
--- a/src/helics/common/JsonBuilder.cpp
+++ b/src/helics/common/JsonBuilder.cpp
@@ -73,6 +73,12 @@ bool JsonMapBuilder::clearComponents(int32_t code)
     return false;
 }
 
+bool JsonMapBuilder::clearComponents()
+{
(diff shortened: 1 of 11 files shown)


nightlark (approved):

nightlark:
Oh, strange. All the checks have passed.

phlptp:
I seriously doubt this fixed the root issue.

PR #1933, opened 2021-05-28
add a test case and try to fix the wait for current time bug

phlptp:
### Summary

If merged this pull request will tweak the time granting for wait for current time,  add some specific tests related to #1926

### Proposed changes

- tweak time grant under wait_for_current_time_flag
- add some testing support in TimeCoordinator.cpp

### DIFF
--- a/tests/helics/core/TimeCoordinatorTests.cpp
+++ b/tests/helics/core/TimeCoordinatorTests.cpp
@@ -139,9 +139,11 @@ TEST(timeCoord_tests, dependent_test_message)
     EXPECT_TRUE(deps[0] == fed3);
 }
 
-class TimeCoordinatorTester1: public ::testing::Test,helics::TimeCoordinator {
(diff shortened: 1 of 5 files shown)


tests/helics/core/TimeCoordinatorTests.cpp
```diff
@@ -138,3 +138,27 @@ TEST(timeCoord_tests, dependent_test_message)
     EXPECT_EQ(deps.size(), 1U);
     EXPECT_TRUE(deps[0] == fed3);
 }
+
```
phlptp, line 141: I was trying something here.  I might keep playing with it, but I figured out the issue before needing the testing here,  but left it in as I am fairly confident there will be more complex issues needing to test the specifics of the time coordinator so will need to build this out more thoroughly at some point

nightlark (approved):

PR #1942, opened 2021-06-04
H2 fixes merge3

phlptp:
### Summary

If merged this pull request will merge the bug fixes from HELICS 2 into HELICS 3


phlptp:
This does not include the doc changes,  only the code changes.  The docs are a little more complicated given the large change to the user guide, so I need to deal with those separately after this gets merged

nightlark (approved):

phlptp:
Full compatibility with HELCIS 3 requires some bigger changes to the queries added in this PR,  to not confuse things those will be made in a separate PR.

PR #1957, opened 2021-06-10
Use calibreapp/image-actions@main for compressing images

nightlark:
### Summary

If merged this pull request will switch to using `calibreapp/image-actions@main` instead of `calibreapp/image-actions@master`

### Proposed changes

- Use calibreapp/image-actions@main for compressing images workflow


nightlark:
I think this will get rid of the `Update required: Update image-actions configuration to the latest version before 1/1/21. See README for instructions.` message.

phlptp (approved):

PR #1961, opened 2021-06-12
Fix #1960 by adding missing #include <thread>.

### LINKED ISSUE #1960: missing `#include <thread>` in AsioContextManager.cpp and helicsExport.cpp
Compiling under Fedora 34 using gcc 11.1.1, the compiler complains that `std::this_thread::sleep_for` is not part of `std::thread`.  The cause is that the code is missing `#include <thread>`.  This probably compiles with other versions of gcc or clang because another include, such as `<mutex>` happens to already include `<thread>`, but of course that's not portable.

beroset:
Signed-off-by: Ed Beroset <beroset@ieee.org>

### Summary

If merged this pull request will fix #1960 

### Proposed changes

Minimal change to add missing #include <thread>

### DIFF
--- a/src/helics/common/AsioContextManager.cpp
+++ b/src/helics/common/AsioContextManager.cpp
@@ -23,6 +23,7 @@ SPDX-License-Identifier: BSD-3-Clause
 #include <map>
 #include <mutex>
 #include <stdexcept>
+#include <thread>
 #include <utility>
 
 /** a storage system for the available core objects allowing references by name to the core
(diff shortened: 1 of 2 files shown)


phlptp (approved):

phlptp:
Can we include you in the contributors document?  If so just github username or any additional information?

beroset:
> Can we include you in the contributors document? If so just github username or any additional information?

Yes, and just github username is good for me.  Thanks!

phlptp:
Thanks again

Issue #1963, opened 2021-06-17
Add clang 12 CI build


nightlark:
Add a clang 12 build - similar to #1962

Issue #1962, opened 2021-06-12
Add GCC 11 CI build


phlptp:
Gcc 11.1 was released in Apr 2021.  We need to add a CI build with GCC 11 to make sure we don't get any errors/warnings on that compiler

nightlark:
* Upgrade AZP gcc10 build to gcc11
* Enable C++20 on the gcc11 build

PR #1968, opened 2021-06-18
Use latest gcc and clang releases for CI builds

nightlark:
### Summary

If merged this pull request will bump the CI builds for the newest gcc and clang builds from gcc 10 to gcc 11, and clang 10 to clang 12. Addresses #1962 and #1963 

### Proposed changes

- Update the gcc-10 builder to gcc-11
- Update the clang-10 builder to clang-12

### DIFF
--- a/.ci/azure-pipelines.yml
+++ b/.ci/azure-pipelines.yml
@@ -15,10 +15,10 @@ jobs:
         ubuntuDefault:
           containerImage: 'helics/buildenv:ubuntu18.04-default-builder'
           test_config: 'ci'
-        gcc10:
+        gcc11:
           containerImage: 'helics/buildenv:gcc11-builder'
           test_config: 'ci'
-        clang10:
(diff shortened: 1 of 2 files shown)


nightlark:
@phlptp I'm having trouble finding an mpi package on Fedora that includes working `mpicc` commands; is gcc 11 not available on Ubuntu yet?

phlptp:
No they came out around the same time, so gcc 11 won't be on ubuntu to 21.10 release

nightlark:
I figured out how to get `mpicc` working on Fedora; not sure if there's an easy way to make sure `module load mpi` runs in the Dockerfile or if the CI yaml file will need to be changed to run the command.

nightlark:
Both clang12 and gcc11 builds are working as well as the previous compilers - just intermittent failing tests now.

phlptp (approved):

PR #1973, opened 2021-06-23
cmake change

nitin-barthwal:
cmake instead of cCMake

### Summary

### DIFF
--- a/docs/installation/linux.md
+++ b/docs/installation/linux.md
@@ -55,7 +55,7 @@ cd HELICS
 mkdir build
 cd build
 CMake ../
-cCMake . # optional, to change install path or other configuration settings
+cmake . # optional, to change install path or other configuration settings
 make
 make install
 ```


phlptp (approved):

phlptp:
This change conflicts with the previous PR, should be easy to resolve.

docs/installation/linux.md
```diff
@@ -55,7 +55,7 @@ cd HELICS
 mkdir build
 cd build
 CMake ../
-cCMake . # optional, to change install path or other configuration settings
+cmake . # optional, to change install path or other configuration settings
```
nightlark, line 58:
```suggestion
cmake . # optional, to change install path or other configuration settings
```

Should the comment be expanded to say how the configs can be changed (editing CMakeCache.txt file, providing -D arguments to this command invocation)? I think the idea behind using `ccmake` was that it brings up a command line interface that lists the available options and their current settings from the CMakeCache.txt file.

phlptp:
This change has been merged in #1988.  
Thank you for the suggestion

PR #1979, opened 2021-06-26
Wait flag endpoint update

phlptp:
### Summary

If merged this pull request will fix some issues with `wait_for_current_time_flag` and its use with endpoints.  

### Proposed changes

-  add some tests
- add error checks for multiple federates using the flag
- change the timing so wait_for_current_time works on time 0 as well
- make it so wait_for_current_time works with endpoints

### DIFF
--- a/tests/helics/apps/TracerTests.cpp
+++ b/tests/helics/apps/TracerTests.cpp
@@ -647,8 +647,9 @@ INSTANTIATE_TEST_SUITE_P(tracer_tests,
 /*
 #ifdef ENABLE_ZMQ_CORE
 #    ifndef DISABLE_SYSTEM_CALL_TESTS
-//TODO: I think the EXE tests need to be moved to a different structure.  The EXE runner doesn't always work right for some reason
-// 
+//TODO: I think the EXE tests need to be moved to a different structure.  The EXE runner doesn't
+always work right for some reason
+//
 TEST_P(tracer_message_file_tests, test_message_files_exe)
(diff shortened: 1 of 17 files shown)


nightlark (approved):

nightlark:
It looks like there is an issue with `ftc_tests.timing_test1` consistently [failing](https://dev.azure.com/HELICS-test/HELICS/_build/results?buildId=9371&view=logs&j=93c97b79-db44-5fec-340e-53842163cada&t=70e148cc-01f4-56bc-cde3-8f079ab6897c&l=138) that should be resolved before merging.

phlptp:
I fixed that test,  there are a few other tests I am trying to resolve as well before I would merge this, as well as the clang-tidy issues.

tests/helics/apps/TracerTests.cpp
```diff
@@ -644,19 +644,20 @@ INSTANTIATE_TEST_SUITE_P(tracer_tests,
                          tracer_clone_file_tests,
                          ::testing::ValuesIn(simple_clone_test_files));
 
+/*
 #ifdef ENABLE_ZMQ_CORE
```
nightlark, line 648: The #ifdef looks like it should be updated if the test is no longer using ZMQ.

phlptp: I am playing with this test yet to see if it passes.  I will clean up once I figure what I want to do with it

PR #1985, opened 2021-06-29
Update Java and Matlab swig interface files

nightlark:
### Summary

If merged this pull request will update the swig generated files for the Java and Matlab interfaces.

### Proposed changes

- Update swig generated Java and Matlab interface files


nightlark:
Looks like the default swig for my WSL2 ubuntu is a bit older... not sure if that has affected the output significantly beyond just a different version number in the header for the Java interface.

phlptp:
Is there any changes in this besides the interface files?

nightlark:
There shouldn’t be any other changes. Updates to the CMakeLists and GitHub
actions workflow will be a separate PR

phlptp (approved):

PR #2008, opened 2021-07-12
Rtd link cleanup

trevorhardy:
### Summary

Further link clean-up and relocation of the "Apps" documentation to the "References" section (and including it in the index).


docs/user-guide/advanced_topics/queries.md
```diff
@@ -60,45 +54,45 @@ The following queries are defined for federates. Federates may specify a callbac
 +--------------------+------------------------------------------------------------+
 | queryString        | Description                                                |
 +====================+============================================================+
-| ``name``           | the identifier of the federate [string]                    |
```
phlptp, line 63: I was leaving these types in the description since there is a difference between returning  full JSON structure and a json compatible string or just the word `true` or something.  I am willing to take it out but it could still be useful to leave them in?

trevorhardy: Ahh, I see what you mean. Hmmm.....

I'm fine leaving them in for now. We'll see if we get any comments about it being confusing to say they are JSON but have those supplemental descriptions of the JSON on each line.

phlptp: How about changing all the ones with `[JSON]` to `[structure]` or `[object]` and then adding and additional line describing the notation further above the table?

trevorhardy: That could work; let's try it and see how it goes. Is that something you can do?

phlptp: I updated the queries and merged in the latest helics 3 commit,  can you take a look @trevorhardy

trevorhardy: Looks good; let's go with it.

phlptp: ok I will merge once the CI passes
thanks

PR #2030, opened 2021-07-20
Remove README in-line html details tag to fix linting error

nightlark:
### Summary

If merged this pull request will remove the html details tags surrounding the optional 3rd party components in the README -- the markdownlint ruby gem doesn't handle the markdown links correctly when they are in that section.

There is a nodejs markdownlint tool that could be considered as an alternative, but the config file format might not be compatible with the current tool used.

### Proposed changes

- Remove html `<details>` tags surrounding the optional 3rd party components in the README

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,9 +135,6 @@ Contributors are welcome see the [Contributing](CONTRIBUTING.md) guidelines for 
 
 ### Optional components
 
-<details>
-  <summary>Click to expand!</summary>
-
 A list of optional component that are not included in HELICS but are optionally used by the library
 
 #### [zmq](http://www.zeromq.org)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -135,9 +135,6 @@ Contributors are welcome see the [Contributing](CONTRIBUTING.md) guidelines for
 
 ### Optional components
 
-<details>
-  <summary>Click to expand!</summary>
-
 A list of optional component that are not included in HELICS but are optionally used by the library
```
phlptp, line 138: may be at some point we should just move this to a separate md file and link it in the readme.

phlptp (approved):

README.md
nightlark: They seem like they serve a similar purpose as the list of libraries on https://github.com/GMLC-TDC/HELICS/blob/main/CONTRIBUTORS.md#used-libraries-or-code so maybe the two lists could get merged later?

phlptp: the distinction is that one list is required and one is optional,  which has some legal distinction particularly with zmq.  But we may be able to put an optional component section in the contributors list.

nightlark: I think both lists are just explaining how the libraries get used so separate lists don't make much difference. If we want to make an optional component section for libraries, we'd want to move Boost to that list as well.

It does feel like the list is less for users/developers and more for the convenience of whoever is reviewing software release requests.

nightlark: I'll leave this PR as-is for now, and that can be a later change.

phlptp: Agreed lets clean this up by the next release, probably move this list to a section of the contributors md file.

PR #2043, opened 2021-08-09
Json serialization

phlptp:
### Summary

If merged this pull request will add a json serialization method for the action Messages

This is step 1 for a compatibility layer.  
Step 2 is a json translation of the value layer.  

### Proposed changes

- add json serialization for the ActionMessages.   
- add some test cases to test the serialization as unit tests and as part of a federation.  
- add a benchmark test for the serialization ( basically 100x slower)

### DIFF
--- a/benchmarks/helics/ActionMessageBenchmarks.cpp
+++ b/benchmarks/helics/ActionMessageBenchmarks.cpp
@@ -10,15 +10,22 @@ SPDX-License-Identifier: BSD-3-Clause
 
 using namespace helics;  // NOLINT
 
-static void BMtoString(benchmark::State& state)
-{
+static ActionMessage generateTestMessage1() {
     ActionMessage obj(CMD_REG_FED);
     obj.name("the name of the federate is really long");
     obj.setStringData("this is a new string to add to the string data");
+    return obj;
+}
+
+static const auto testMessage1 = generateTestMessage1();
+
+static void BMtoString(benchmark::State& state)
+{
+    
     std::string load;
     load.reserve(500);
     for (auto _ : state) {
-        obj.to_string(load);
+        testMessage1.to_string(load);
     }
 }
 // Register the function as a benchmark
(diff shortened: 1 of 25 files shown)


phlptp:
This is the first of a couple PR's.
The next one will try to deal with the value serialization to get a json translation of that part.    This one allows a connection to work between 2 and 3 and probably messages,  but values won't work so well.    The values might be trickier but we will see.

benchmarks/helics/ActionMessageBenchmarks.cpp
```diff
@@ -134,4 +134,136 @@ static void BMdepacketizeStrings(benchmark::State& state)
...
+    }
+}
+// Register the function as a benchmark
+BENCHMARK(BMpacketizeJson);
+
+static void BMdepacketizeJson(benchmark::State& state)
+{
+    ActionMessage obj(CMD_REG_FED);
+    obj.name("the name of the federate is really long");
+    obj.setStringData("this is a new string to add to the string data");
```
nightlark, line 225: Is this object something that could just get set up in one place and used for all the benchmarks?

phlptp: I am sure it could.  I just copied all the benchmarks nearly identically,

src/helics/core/ActionMessage.cpp
```diff
@@ -281,12 +283,105 @@ int ActionMessage::serializedByteCount() const
...
+        packet["Te"] = Te.getBaseTimeCode();
+        packet["Tdemin"] = Tdemin.getBaseTimeCode();
+        packet["Tso"] = Tso.getBaseTimeCode();
+    }
+    packet["payload"] = b64Encode(payload.to_string());
+    packet["stringCount"] = stringData.size();
+    if (!stringData.empty()) {
+        packet["strings"] = Json::arrayValue;
+        for (const auto& str : stringData){
+            packet["strings"].append(b64Encode(str));
```
nightlark, line 379: What do you think about skipping the base64 encoding and let jsoncpp do its escaping as needed for string values? I think either way jsoncpp will be checking strings to see if it needs to escape them and creating a new std::string object, though we'd reduce the number of string objects and checks on our side.

I guess the downside would be encoded size, with the worst case for not encoding as base64 being a large blob of data that only has byte values below 32; and the worst case for encoding as base64 would be if there is a large blob of binary data with only a few characters below 32 triggering the base64 encoding resulting in the (~33%?) overhead of base64 encoding.

phlptp: I tried this seems to work.  Probably will be slower in a few cases but might not matter that much

phlptp: The binary strings tests takes quite a bit longer but that might be expected being that is 50000 bytes which would all need to be escaped

PR #2053, opened 2021-08-16
Json data serialization

phlptp:
### Summary

If merged this pull request will add json serialization methods and HELICS_JSON data type.  Federates with the `--json` flag will force all the outgoing data to this format.


phlptp:
I tried various methods of getting around this but I am going to have to add the interpreters for HELICS 3 data serialization into HELICS 2.     There was no way to force a publisher to the json type without breaking the layer separate of HELICS either by the data interpretation or forcing the type system from the core to a federate.    So HELICS 2 is going to need the translaters for HELICS 3 data.  But I think this is everything needed on the HELICS 3 side.

nightlark (approved):

PR #2085, opened 2021-08-30
Json serialization2

phlptp:
### Summary

If merged this pull request will merge the actionMessage serialization into HELICS 2

### DIFF
--- a/src/helics/core/ActionMessage.cpp
+++ b/src/helics/core/ActionMessage.cpp
@@ -282,7 +282,7 @@ std::string ActionMessage::to_string() const
     } else {
         auto sz = serializedByteCount();
         data.resize(sz);
-        toByteArray(reinterpret_cast<std::byte*>(&(data[0])), sz);
+        toByteArray(&(data[0]), sz);
     }
     return data;
 }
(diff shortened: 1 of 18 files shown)


phlptp:
This is the first of 2 PR's 
the next will add the data serialization

src/helics/core/Core.hpp
```diff
@@ -579,11 +579,12 @@ class Core {
     @param federateName the name of the dependent federate
     */
     virtual void addDependency(local_federate_id federateID, const std::string& federateName) = 0;
+
     /**
-     * Register known frequently communicating source/destination end points.
+     * Register communicating source/destination end points targets.
```
nightlark, line 584:
```suggestion
     * Register communicating source/destination endpoint targets.
```

src/helics/core/ActionMessage.cpp
```diff
@@ -275,12 +277,48 @@ int ActionMessage::serializedByteCount() const
...
+        toByteArray(&(data[0]), sz);
+    }
     return data;
 }
 
+std::string ActionMessage::to_json_string() const
+{
+    Json::Value packet;
+    packet["version"] =
+        HELICS_VERSION_MAJOR * 10000 + HELICS_VERSION_MINOR * 100 + HELICS_VERSION_PATCH;
```
nightlark, line 294: What will the version be used for? Is it something that could just be sent once when the connection is first established?

phlptp: We need some sort of version indicator to match up future changes to the json output. I don't think we could only send it once, otherwise we would have to record it and send it to the interpreter every time.    Right now it isn't used since there is only one version,  but it would start to get used if we ever upgrade the message format.

nightlark (approved):

PR #2086, opened 2021-08-31
Add CodeQL Analysis workflow

nightlark:
### Summary

If merged this pull request will add a CodeQL Analysis workflow.

### Proposed changes

- Add a CodeQL Analysis workflow


phlptp:
What does this do that the other static analyzers don't?

nightlark:
It checks for likely bugs, I think with a focus on mostly memory management/security related -- maybe more similar to Klocwork than clang-tidy/cpplint in the things it checks for and the way it ties into a build.

Though it seems like many of the source files are producing "compilation errors".

PR #2106, opened 2021-09-15
Remove fix for ASIO_ERROR_CATEGORY_NOEXCEPT definition bug affecting Boost 1.58

nightlark:
### Summary

If merged this pull request will remove a fix for `ASIO_ERROR_CATEGORY_NOEXCEPT` not getting set correctly due to Boost 1.58 not setting `BOOST_NOEXCEPT`. This change will be needed to use anything included from the asio ssl headers.

### Proposed changes

- Removed redefinition of `ASIO_ERROR_CATEGORY_NOEXCEPT` that was needed for Boost 1.58 compatibility

### DIFF
--- a/src/helics/common/AsioContextManager.h
+++ b/src/helics/common/AsioContextManager.h
@@ -27,16 +27,7 @@ SPDX-License-Identifier: BSD-3-Clause
 #include <string>
 #include <utility>
 
-// The choice for noexcept isn't set correctly in asio::io_context (including asio.hpp instead
-// didn't help) With Boost 1.58 this resulted in a compile error, apparently from the BOOST_NOEXCEPT
-// define being empty
-#ifdef ASIO_ERROR_CATEGORY_NOEXCEPT
-#    undef ASIO_ERROR_CATEGORY_NOEXCEPT
-#endif
-
-#define ASIO_ERROR_CATEGORY_NOEXCEPT noexcept(true)
 #include <asio/io_context.hpp>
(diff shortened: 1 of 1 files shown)


src/helics/common/AsioContextManager.h
```diff
@@ -27,17 +28,6 @@ SPDX-License-Identifier: BSD-3-Clause
 #include <string>
 #include <utility>
 
-// The choice for noexcept isn't set correctly in asio::io_context (including asio.hpp instead
```
phlptp, line 30: The reason we can remove this is because we raised the minimum boost requirement?

nightlark: I think so -- as far as I can tell from the comments in https://github.com/GMLC-TDC/HELICS/pull/670 it was to get the Boost 1.58 build working.

phlptp: where were we testing that?

phlptp (approved):

src/helics/common/AsioContextManager.h
nightlark: One of the old Travis CI builds -- the old logs aren't available now, but from the commit success/fail statuses on the PR and the comment I think it's likely the builds with newer versions of Boost compiled.

I'm not seeing compile errors with our current CI jobs either, so our new minimum Boost version should be setting `BOOST_NOEXCEPT`.

PR #2114, opened 2021-09-22
Endpoint link

phlptp:
### Summary

If merged this pull request will add the endpoint link function to CoreApp, brokerApp, and Broker, and actually add some tests to make sure it is working properly.

### DIFF
--- a/src/helics/core/fileConnections.hpp
+++ b/src/helics/core/fileConnections.hpp
@@ -48,9 +48,62 @@ namespace fileops {
                         });
                     } else {
                         std::string ipt = getOrDefault(conn, "input", std::string());
-                        addTargets(conn, "targets", [brk, &ipt](const std::string& target) {
-                            brk->dataLink(target, ipt);
+                        if (!ipt.empty()) {
+                            addTargets(conn, "targets", [brk, &ipt](const std::string& target) {
+                                brk->dataLink(target, ipt);
+                            });
+                            addTargets(conn, "sources", [brk, &ipt](const std::string& target) {
+                                brk->dataLink(target, ipt);
+                            });
+                        } else {
+                            std::string ept = getOrDefault(conn, "endpoint", std::string());
+                            if (!ept.empty()) {
+                                addTargets(conn, "targets", [brk, &ept](const std::string& target) {
+                                    brk->linkEndpoints(ept,target);
+                                });
+                                addTargets(conn, "sources", [brk, &ept](const std::string& target) {
+                                    brk->linkEndpoints(target, ept);
+                                });
+                            }
+                        }
+                       
+                    }
+                }
+            }
+        }
+        auto lnks = toml::find_or(doc, "links", uVal);
+        if (!lnks.is_uninitialized()) {
+            auto& connArray = lnks.as_array();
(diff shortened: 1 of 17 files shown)


src/helics/core/fileConnections.hpp
```diff
@@ -48,9 +48,61 @@ namespace fileops {
...
                     } else {
                         std::string ipt = getOrDefault(conn, "input", std::string());
-                        addTargets(conn, "targets", [brk, &ipt](const std::string& target) {
-                            brk->dataLink(target, ipt);
+                        if (!ipt.empty()) {
+                            addTargets(conn, "targets", [brk, &ipt](const std::string& target) {
+                                brk->dataLink(target, ipt);
+                            });
+                            addTargets(conn, "sources", [brk, &ipt](const std::string& target) {
+                                brk->dataLink(target, ipt);
```
nightlark, line 56: It seems like the various dataLink functions arguments might be better named `publication` and `input` in a later PR.

Is a data link created from the same publication/input getting added to both "targets" and "sources" because the created link is bidirectional?

```suggestion
                            addTargets(conn, "targets", [brk, &ipt](const std::string& pub) {
                                brk->dataLink(pub, ipt);
                            });
                            addTargets(conn, "sources", [brk, &ipt](const std::string& pub) {
                                brk->dataLink(pub, ipt);
```

tests/helics/test_files/example_connections6.json
```diff
@@ -0,0 +1,14 @@
...
+{
+  //example json connection file  connections can be an array of two element string arrays  ["publication", "input"] or a map with "publication" and "target" for a single string or an array strings as "targets"
+  "connections": [
+    {
+      "publication": "pub1",
+      "targets": "inp1"
+    },
+    {
+      "endpoint": "ept1",
+      "target": "ept2"
```
nightlark, line 11: This example is using a `targets` field instead of `target` -- is that a typo, or is the entry intended to allow an array such as `["inp1"]`?

I'm wondering if maybe the keys should be `input` instead of `target` for links between a publication and input, and something more descriptive for linking endpoints -- `source_endpoint`/`destination_endpoint`?

tests/helics/test_files/example_connections5.toml
```diff
@@ -0,0 +1,7 @@
+#example toml connection file connections can be an array of two element string arrays  ["publication", "input"]
+connections=[
+["pub1", "inp1"]
+]
+links=[
+["ept1", "ept2"]
+]
```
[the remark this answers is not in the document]
phlptp: I am open to suggestions here, but "connections" was being used for the value linkage if the elements formed an array,  so there needed to be a different keyword for array values for making links between two endpoints.  If it is a structure that issue isn't present.

tests/helics/test_files/example_connections6.json
phlptp: it is actually testing some flexibility in the reader to handle both "target" and "targets"

phlptp: I like that idea though for "source_endpoint" and "destination_endpoint"  I might add that in a subsequent PR.

PR #2140, opened 2021-10-22
Switch docker clang-format hook to use pypi clang-format hook

nightlark:
### Summary

If merged this pull request will replace the docker based clang-format hook with one that uses PyPI. This should make installing pre-commit hooks locally a bit easier (no docker dependency), and makes so that this step can also be run by the `pre-commit.ci` service if we decide to switch.

### Proposed changes

- Replace the local docker clang-format hook with the `pre-commit/mirrors-clang-format` hook that uses the clang-format PyPI package
- Probably using a newer version of clang-format than was in the docker container


nightlark:
#2141 has two formatting changes; I think the docker image may have had clang-format 8, whereas this is using clang-format 13.

Issue #2163, opened 2021-11-04
UnicodeDecodeError from HelicsMessage.data for messages containing long strings


kdheepak:
**Describe the bug**

Using long strings when sending data via endpoints causes the null terminator to be ignored when reading the data on the receiving endpoint.

This is the function that causes the issue:

https://github.com/GMLC-TDC/HELICS/blob/91f4cbcfb2a003cae955521a22e83461cbf2b629/src/helics/core/SmallBuffer.hpp#L227

This was tested using HELICS 3.0.1 and pyhelics `main` branch. Related to https://github.com/GMLC-TDC/pyhelics/issues/37

PR #2164, opened 2021-11-04
fix null terminator issue in the C library for message objects

phlptp:
### Summary

If merged this pull request will add a null terminator in a few select spots in the C API for message objects.

### DIFF
--- a/src/helics/core/SmallBuffer.hpp
+++ b/src/helics/core/SmallBuffer.hpp
@@ -229,9 +229,7 @@ class SmallBuffer {
     {
         if (bufferCapacity > bufferSize) {
             heap[bufferSize] = std::byte(0);
-        }
-        else
-        {
+        } else {
             push_back('\0');
             pop_back();
         }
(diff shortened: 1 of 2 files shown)


phlptp:
Closes #2163

src/helics/core/SmallBuffer.hpp
```diff
@@ -279,15 +292,15 @@ class SmallBuffer {
             if (size > 0x010'0000'0000U) {
                 throw(std::bad_alloc());
             }
-            auto* ndata = new std::byte[size];
+            auto* ndata = new std::byte[size+8];
```
nightlark, line 295: Why the extra 8 bytes? Just a bit of extra capacity in case more data gets added?

phlptp: Now there is a fairly common use case with a need to add a single character, no sense in just adding a single character though. I suppose I could add some sort of exponential growth factor as well like std::string.

nightlark (approved):

kdheepak (approved):

PR #2152, opened 2021-10-27
type_conversion docs

phlptp:
add some additional operations in the C library for converting strings to a dataType, and documentation about the type conversions.

### Summary

If merged this pull request will add operations for getting a dataType from a string in the C library and docs for type conversion operations.

PR #2170, opened 2021-11-09
Add blacken-docs pre-commit hook

nightlark:
### Summary

If merged this pull request will cherry-pick the addition of the blacken-docs pre-commit hook from #2152

### Proposed changes

- Add blacken-docs pre-commit hook

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -11,7 +11,7 @@ repos:
       - id: black
         args: ["--line-length=100"]
   - repo: https://github.com/asottile/blacken-docs
-    rev: v1.8.0
+    rev: v1.11.0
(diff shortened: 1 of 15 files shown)


phlptp:
good call pulling this out.  Saw the pre-commit check in another project and thought that it looked useful

nightlark:
Nice how it fixes up the extra indentation we had.

phlptp (approved):

phlptp:
looks like there are a few parse issues in the docs that need to be addressed yet

nightlark:
It seems to not like the start of a code block (while, if, etc) without any subsequent lines. I've added `pass` commands in those cases with a "placeholder" comment.

PR #2203, opened 2021-11-20
Fix warnings

phlptp:
### Summary

If merged this pull request will clean up a number of warnings from various compilers and situations
and update the toml11 library.

### DIFF
--- a/src/helics/application_api/HelicsPrimaryTypes.hpp
+++ b/src/helics/application_api/HelicsPrimaryTypes.hpp
@@ -34,7 +34,7 @@ using defV = std::variant<double,
                           NamedPoint>;
 
 /**enumeration of the order inside the variant so the Which function returns match the enumeration*/
-enum type_location:std::uint32_t {
+enum type_location : std::uint32_t {
     double_loc = 0U,
     int_loc = 1U,
     string_loc = 2U,
--- a/src/helics/application_api/HelicsPrimaryTypes.hpp
+++ b/src/helics/application_api/HelicsPrimaryTypes.hpp
@@ -34,14 +34,14 @@ using defV = std::variant<double,
                           NamedPoint>;
 
 /**enumeration of the order inside the variant so the Which function returns match the enumeration*/
-enum type_location {
-    double_loc = 0,
-    int_loc = 1,
-    string_loc = 2,
-    complex_loc = 3,
-    vector_loc = 4,
-    complex_vector_loc = 5,
-    named_point_loc = 6
+enum type_location:std::uint32_t {
+    double_loc = 0U,
+    int_loc = 1U,
+    string_loc = 2U,
+    complex_loc = 3U,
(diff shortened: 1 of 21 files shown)


src/helics/application_api/HelicsPrimaryTypes.hpp
```diff
@@ -34,14 +34,14 @@ using defV = std::variant<double,
                           NamedPoint>;
 
 /**enumeration of the order inside the variant so the Which function returns match the enumeration*/
```
nightlark, line 36: Unrelated to the warnings, but I'm a bit confused what this comment is saying.

Is it something like "enumeration of the order inside the variant so the order of the function returns match the enumeration"?

src/helics/apps/AsioBrokerServer.cpp
```diff
@@ -94,7 +94,7 @@ namespace udp {
 
         asio::ip::udp::socket mSocket;
         asio::ip::udp::endpoint mRemoteEndpoint;
-        std::array<char, 1024> mRecvBuffer{0};
+        std::array<char, 1024> mRecvBuffer{{0}};
```
nightlark, line 97: Is this something like nested value initialization, with the inner `{0}` specifying the value for the individual char entries, and `{{0}}` setting the initial value for the entire array to 0's?

src/helics/apps/AsioBrokerServer.cpp
```diff
@@ -270,10 +270,12 @@ namespace apps {
     void AsioBrokerServer::mainLoop()
     {
         auto ioctx = AsioContextManager::getContextPointer();
+
 #ifdef HELICS_ENABLE_TCP_CORE
         if (tcp_enabled_) {
             tcpserver = loadTCPserver(ioctx->getBaseContext());
             tcpserver->setDataCall(
+                // NOLINTNEXTLINE
```
nightlark, line 278: What warning was caused by the next line?

src/helics/application_api/HelicsPrimaryTypes.hpp
phlptp: yes

src/helics/apps/AsioBrokerServer.cpp
phlptp: clang-tidy warning about the parameter being able to be passed by reference but it is a shared pointer that needs its lifetime extended so it shouldn't be passed by reference.

src/helics/apps/AsioBrokerServer.cpp
phlptp: one of the clang compilers gave a warning on this, something about partial initializer lists and recommended this fix.

phlptp: and no it is just the first value that is set to 0

nightlark (approved):

PR #2214, opened 2021-11-26
Develop

phlptp:
### Summary

If merged this pull request will merge develop into main for the HELICS 3.1 release


phlptp:
@nightlark  Haven't seen the MSYS2 CI/build before?  Do we need to do something about the failure?

nightlark:
The MSYS2 CI build is set up to run for branches with names starting with various msys/mingw prefixes, and run nightly on develop (though I don't think that will start happening until the workflow file gets merged to main).

I'm not sure what changed in the past month that would have made this start failing in spdlog or HELICS, though it looks like expanding these two conditions to exclude the syslog sink with `__MINGW32__` would be the easiest way to get it building for the release:
https://github.com/GMLC-TDC/HELICS/blame/develop/src/helics/core/BrokerBase.cpp#L23
https://github.com/GMLC-TDC/HELICS/blame/develop/src/helics/core/BrokerBase.cpp#L296

nightlark:
@phlptp I'm doing a build of the release installers -- I think the NSIS windows installer packaging might be having issues.

nightlark (approved):
I think this should be okay... and no surprises when we go to update packages or the installers build.

PR #2215, opened 2021-11-26
add conditions for not including syslog

phlptp:
### Summary

If merged this pull request will update the conditions for including syslog for mingw

### DIFF
--- a/src/helics/core/BrokerBase.cpp
+++ b/src/helics/core/BrokerBase.cpp
@@ -28,7 +28,7 @@ SPDX-License-Identifier: BSD-3-Clause
 
 #    include <asio/steady_timer.hpp>
 #else
-#    if defined(_WIN32) || defined(WIN32) || defined(__MINGW32__) || defined(CYGWIN)
+#    if defined(_WIN32) || defined(WIN32)
 #        include <windows.h>
 #    else
 #        include <unistd.h>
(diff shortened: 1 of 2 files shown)


src/helics/core/BrokerBase.cpp
```diff
@@ -28,7 +28,7 @@ SPDX-License-Identifier: BSD-3-Clause
 
 #    include <asio/steady_timer.hpp>
 #else
-#    ifdef _WIN32
+#    if defined(_WIN32) || defined(WIN32) || defined(__MINGW32__) || defined(CYGWIN)
```
phlptp, line 31: not sure about this one @nightlark does this look right to you as far as what conditions windows.h should be included?  Of course this is only if ASIO is disabled

phlptp:
@nightlark  you can go ahead and merge if you are happy with and see if the MSYS builds pass on develop

src/helics/core/BrokerBase.cpp
nightlark: I think the conditions for this should be _WIN32 or WIN32 -- it seems like MSYS failure is because for some reason WIN32 is no longer being defined while _WIN32 is,  maybe something changed in the latest version.

```suggestion
#    if defined(_WIN32) || defined(WIN32)
```

nightlark:
It seems to the MSYS CI jobs run the `mingw` in the branch name needs to be lower case. In https://github.com/GMLC-TDC/HELICS/commit/e28740630c3e49a6a60176318643a72ebd8dc621 I found that changing it from `WIN32` to `_WIN32` makes the builds succeed again, so I think there could have been a change to MSYS/MinGW that made so `WIN32` is no longer being defined.

I think being conservative like this with the extra conditions for MINGW32 and CYGWIN to include syslog_sink makes sense for this release. For including `windows.h` I'll change it to just `WIN32` or `_WIN32` since MinGW and Cygwin should either define those if they have windows.h, or include unistd.h.

nightlark (approved):

PR #2220, opened 2021-12-03
Submodule move

phlptp:
### Summary

If merged this pull request will move some of the networking code into a separate repository and move toml to the upstream version instead of the GMLC-TDC maintained one. 

### Proposed changes

- update toml submodule
- update units submodule
- move networking code to separate repo
- test with CMake 3.22
- change version numbers to mark build changes


CMakeLists.txt
```diff
@@ -704,6 +709,25 @@ if(NOT HELICS_DISABLE_BOOST)
...
+add_subdirectory(ThirdParty/networking)
+
+hide_variable(GMLC_NETWORKING_GENERATE_DOXYGEN_DOC)
+hide_variable(GMLC_NETWORKING_INCLUDE_BOOST)
+hide_variable(GMLC_NETWORKING_WITH_CMAKE_PACKAGE)
+hide_variable(GMLC_NETWORKING_OBJECT_LIB)
+hide_variable(GMLC_NETWORKING_STATIC_LIB)
+hide_variable(GMLC_NETWORKING_CLANG_TIDY_OPTIONS)
+
+add_library(HELICS::networking ALIAS gmlc_networking)
```
nightlark, line 728: This sort of seems like gmlc_networking should export an alias target called GMLC::networking

src/helics/apps/MultiBroker.hpp
```diff
@@ -30,7 +30,8 @@ class MultiBroker: public CoreBroker {
     std::unique_ptr<CommsInterface>
         masterComm;  //!< the primary comms object or the one that links with the master
     NetworkBrokerData netInfo{
-        InterfaceTypes::TCP};  //!< structure containing the networking information
+        NetworkBrokerData::InterfaceTypes::TCP};  //!< structure containing the
```
nightlark, line 33: How were you deciding which parts use InterfaceTypes from NetworkBrokerData vs gmlc::networking?

CMakeLists.txt
phlptp: We might want to consider a refresh on all the submodule libraries in GMLC-TDC that would standardize some of this.  But this is probably a good idea, I want to make sure it is consistent with all of them though.

src/helics/apps/MultiBroker.hpp
phlptp: they are aliased to eachother so are the same, but what I sort of tried to do is if it was directly in relation to NetworkBrokerData I would use the networkBrokerData alias.    Might need to be a bit more consistent or just get rid of the NetworkBrokerData alias to make sure everything is consistent.

CMakeLists.txt
```diff
@@ -696,14 +696,33 @@ hide_variable(GMLC_UTILITIES_STATIC_LIB)
...
+add_subdirectory(ThirdParty/networking)
+
+hide_variable(GMLC_NETWORKING_GENERATE_DOXYGEN_DOC)
+hide_variable(GMLC_NETWORKING_INCLUDE_BOOST)
+hide_variable(GMLC_NETWORKING_WITH_CMAKE_PACKAGE)
+hide_variable(GMLC_NETWORKING_OBJECT_LIB)
+hide_variable(GMLC_NETWORKING_STATIC_LIB)
+hide_variable(GMLC_NETWORKING_CLANG_TIDY_OPTIONS)
+
+set_target_properties(gmlc_networking PROPERTIES FOLDER Extern)
```
nightlark, line 724: Can the namespaced alias target name be used here?

```suggestion
set_target_properties(gmlc::networking PROPERTIES FOLDER Extern)
```

nightlark (approved):

PR #386, opened 2018-07-17
Clang tidy cleanup

phlptp:
some code cleanup and some updates to the cmake code.

PR #2222, opened 2021-12-06
Remove setting of FindThreads output result variable

nightlark:
### Summary

If merged this pull request will remove setting a FindThreads output result variable that signals WIN32 threads are being used, and combine all the non-Windows thread cases into a single check.

### Proposed changes

- Remove (redundant) setting the FindThreads.cmake output variable `CMAKE_USE_WIN32_THREADS_INIT`
- Combine the conditions for setting `THREADS_PREFER_PTHREAD_FLAG`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -481,12 +481,8 @@ endif(HELICS_ENABLE_MPI_CORE)
 # -------------------------------------------------------------
 # add threading support
 # -------------------------------------------------------------
-if(NOT WIN32)
+if(MSYS OR CYGWIN OR NOT WIN32)
     set(THREADS_PREFER_PTHREAD_FLAG ON)
-elseif(MSYS OR CYGWIN)
-    set(THREADS_PREFER_PTHREAD_FLAG ON)
-else()
-    set(CMAKE_USE_WIN32_THREADS_INIT ON)
 endif()
 find_package(Threads REQUIRED)
 


CMakeLists.txt
```diff
@@ -481,12 +481,8 @@ endif(HELICS_ENABLE_MPI_CORE)
 # -------------------------------------------------------------
 # add threading support
 # -------------------------------------------------------------
-if(NOT WIN32)
-    set(THREADS_PREFER_PTHREAD_FLAG ON)
-elseif(MSYS OR CYGWIN)
+if(MSYS OR CYGWIN OR NOT WIN32)
     set(THREADS_PREFER_PTHREAD_FLAG ON)
-else()
-    set(CMAKE_USE_WIN32_THREADS_INIT ON)
```
phlptp, line 489: Do we know why we set this in the first place.  It is possible it was dealing with something in older cmake versions?

nightlark: There's no useful info about why it was added; the first time it appeared was clang-tidy PR https://github.com/GMLC-TDC/HELICS/pull/386 as part of a much larger commit called "some updates to cmake and requested fixes".

nightlark: The closest conditions surrounding when it gets set in https://github.com/Kitware/CMake/blame/master/Modules/FindThreads.cmake haven't changed significantly in about the past decade.

Setting the variable ourselves doesn't have any affect on what FindThreads.cmake does, and we never reference it in our code either.

phlptp (approved):

PR #2226, opened 2021-12-07
More grant timeout

phlptp:
### Summary

If merged this pull request will add grant_timeout support in core/broker and add a maxcosimduration to kill a cosim if it hangs and produce some debugging information.


src/helics/core/FederateState.cpp
```diff
@@ -1254,8 +1273,21 @@ MessageProcessingResult FederateState::processActionMessage(ActionMessage& cmd)
...
+                    routeMessage(brokerTimeoutCheck);
+                    LOG_WARNING(
+                        fmt::format("sending grant time out check to {}", parentID.baseValue()));
+                }
             } else if (cmd.counter == 10) {
-                LOG_WARNING("grant timeout stage 4 error actions(none available)");
+                if (cmd.actionTime == Time::maxVal()) {
+                    LOG_WARNING("finalize blocking");
+                } else {
+                    LOG_WARNING("grant timeout stage 4 error actions(none available)");
```
nightlark, line 1289:
```suggestion
                    LOG_WARNING("grant timeout stage 4 error actions (none available)");
```

src/helics/core/BrokerBase.cpp
```diff
@@ -238,26 +238,31 @@ std::shared_ptr<helicsCLI11App> BrokerBase::generateBaseCLI()
...
+            timeout,
+            "time to wait to establish a network or for a connection to communicate, default "
+            "unit is in ms (can also be entered as "
+            "a time like '10s' or '45ms') ")
+        ->capture_default_str();
+    timeout_group
+        ->add_option(
+            "--networktimeout",
+            networkTimeout,
+            "time to wait for a broker connection default unit is in ms(can also be entered as a time "
```
nightlark, line 263:
```suggestion
            "time to wait for a broker connection, default unit is in ms (can also be entered as a time "
```

src/helics/core/BrokerBase.cpp
```diff
@@ -238,26 +238,31 @@ std::shared_ptr<helicsCLI11App> BrokerBase::generateBaseCLI()
...
-        "--networktimeout",
-        networkTimeout,
-        "time to wait for a broker connection default unit is in ms(can also be entered as a time "
-        "like '10s' or '45ms') ");
+    timeout_group
+        ->add_option(
+            "--tick",
+            tickTimer,
+            "heartbeat time in ms, if there is no broker communication for 2 ticks then "
+            "secondary actions are taken  (can also be entered as a time like '10s' or '45ms')")
```
nightlark, line 246:
```suggestion
            "secondary actions are taken (can also be entered as a time like '10s' or '45ms')")
```

src/helics/core/BrokerBase.hpp
```diff
@@ -59,17 +59,18 @@ class BrokerBase {
                               //!< period respond with timeout error
     Time errorDelay{10.0};  //!< time to delay before terminating after error state
     Time grantTimeout{-1.0};  //!< timeout for triggering diagnostic action waiting for a time grant
+    Time maxCoSimDuration{-1.0};  //!< the maximum lifetime(wall clock time) of the co-simulation
```
nightlark, line 62:
```suggestion
    Time maxCoSimDuration{-1.0};  //!< the maximum lifetime (wall clock time) of the co-simulation
```

src/helics/core/BrokerBase.hpp
```diff
@@ -85,12 +86,12 @@ class BrokerBase {
...
-        false};  //!< flag indicating that the main processing loop is running
+    /// flag indicating that the main processing loop is running
+    std::atomic<bool> mainLoopIsRunning{false};
     bool dumplog{false};  //!< flag indicating the broker should capture a dump log
     std::atomic<bool> forceLoggingFlush{false};  //!< force the log to flush after every message
-    bool queueDisabled{
-        false};  //!< flag indicating that the message queue should not be used and all functions
+    /// flag indicating that the message queue should not be used and all functions
+    bool queueDisabled{false};
     //!< called directly instead of distinct thread
```
nightlark, line 95: `called directly instead of distinct thread` looks like it should be moved up to join the other comment line above the variable declaration.

src/helics/core/BrokerBase.hpp
```diff
@@ -135,8 +137,8 @@ class BrokerBase {
     /** specify that outgoing connection should use json serialization */
     bool useJsonSerialization{false};
     bool enable_profiling{false};  //!< indicator that profiling is enabled
-    decltype(std::chrono::steady_clock::now())
-        errorTimeStart;  //!< time when the error condition started related to the errorDelay
+    /// time when the error condition started related to the errorDelay
```
nightlark, line 140:
```suggestion
    /// time when the error condition started relative to the errorDelay
```
 Relative to, or just related?

```suggestion
    /// time when the error condition started; related to the errorDelay
```

docs/user-guide/advanced_topics/timeouts.md
```diff
@@ -0,0 +1,42 @@
...
...ead processing messages in the background from the user code. They also make use of system timers and can generate messages for processing independent of the main loop. The foundation of this is a heartbeat timer on the core and brokers. All options in HELICS that take a time are defaulted to milliseconds if only a number is supplied in the argument field. They also can take a time unit with the number such as `5 s` or `2 min` or `265234 ms`. The `--tick` sets up the heartbeat which is fundamental to timeouts at the broker and core level. Enabling this requires HELICS be built with ASIO, so in situations where the CMake option `-DHELICS_DISABLE_ASIO=ON` was given all timeouts are non functional. The tick timer can also be disabled using `--disable_timer`,`--no_tick` or `--debugging`. (NOTE: debugging also changes a few other things). The tick effectively establishes a timing resolution for the timeouts, if the tick is set to 1 second timeouts will have to be exceeded by at least 1 tick for actions to trigger. The default tick is 5 seconds so generally timeouts should be specified in multiples of the tick timer or slightly less than an integer multiple if desired to trigger earlier.
```
nightlark, line 11:
```suggestion
Typical brokers and cores in HELICS operate on a separate thread processing messages in the background from the user code. They also make use of system timers and can generate messages for processing independent of the main loop. The foundation of this is a heartbeat timer on the core and brokers. All options in HELICS that take a time are defaulted to milliseconds if only a number is supplied in the argument field. They also can take a time unit with the number such as `5 s` or `2 min` or `265234 ms`. Using `--tick` sets up the heartbeat which is fundamental to timeouts at the broker and core level. Enabling this requires HELICS be built with ASIO, so in situations where the CMake option `-DHELICS_DISABLE_ASIO=ON` was given all timeouts are non-functional. The tick timer can also be disabled using `--disable_timer`,`--no_tick` or `--debugging`. (NOTE: debugging also changes a few other things). The tick effectively establishes a timing resolution for the timeouts, if the tick is set to 1 second timeouts will have to be exceeded by at least 1 tick for actions to trigger. The default tick is 5 seconds so generally timeouts should be specified in multiples of the tick timer or slightly less than an integer multiple if desired to trigger earlier.
```

docs/user-guide/advanced_topics/timeouts.md
```diff
@@ -0,0 +1,42 @@
...
+
+It is expected this process will change as more experience with it is gained.
+
+At the broker level a `--maxcosimduration` option is available. By default it is disabled. If specified on a broker or core the co-simulation will terminate if running for more than the specified total time. This can be useful if running a large number of co-simulations in an automated fashion and the expected time is known. This can prevent a situation where a single co-simulation hangs indefinitely blocking others and doing no useful work.
+
+## Other timeouts
+
+There are two other timeouts that can be specified. The first is a query timeout. (`--querytimeout=X`) the default is 15 seconds. This applies to queries which are blocking calls, and if a federate is shutting down or fails as a query is in flight it is possible (not likely) that it could result in a non response and the query never completing. In that rare situation the query timeout will trigger and result in an error on the query. In very large co-simulations it is possible a complex query could exceed the timeout without there being a problem so it may be necessary to lengthen it.
```
nightlark, line 36:
```suggestion
There are two other timeouts that can be specified. The first is a query timeout (`--querytimeout=X`). The default is 15 seconds. This applies to queries which are blocking calls, and if a federate is shutting down or fails as a query is in flight it is possible (not likely) that it could result in a non-response and the query never completing. In that rare situation the query timeout will trigger and result in an error on the query. In very large co-simulations it is possible a complex query could exceed the timeout without there being a problem so it may be necessary to lengthen it.
```

docs/user-guide/advanced_topics/timeouts.md
```diff
@@ -0,0 +1,42 @@
...
+
+## Co-simulation startup
+
+Starting up a co-simulation often depends on a number of factors. There are multiple pieces in a co-simulation; at least 1 broker, and often several federates. These communicate via network resources. And things can go wrong, so you don't want to have the cosimulation just hang forever waiting on certain pieces, so the most general option is `--timeout` This is a the timeout for establishing a broker connection, and it is also the default for some other more specific timeouts. The default is 30 seconds. If you are setting up a large co-simulation which may have a lot of contention during the setup phase, it may be necessary to lengthen this, or if you are not starting up the broker before the federates it can be necessary to change to a longer time. Currently 30 seconds is used as a compromise that works in most settings where HELICS is used.
```
nightlark, line 15:
```suggestion
Starting up a co-simulation often depends on a number of factors. There are multiple pieces in a co-simulation; at least 1 broker, and often several federates. These communicate via network resources. And things can go wrong, so you don't want to have the cosimulation just hang forever waiting on certain pieces, so the most general option is `--timeout`. This is a the timeout for establishing a broker connection, and it is also the default for some other more specific timeouts. The default is 30 seconds. If you are setting up a large co-simulation which may have a lot of contention during the setup phase, it may be necessary to lengthen this, or if you are not starting up the broker before the federates it can be necessary to change to a longer time. Currently 30 seconds is used as a compromise that works in most settings where HELICS is used.
```

docs/user-guide/advanced_topics/timeouts.md
```diff
@@ -0,0 +1,42 @@
...
+
+At the federate level a --granttimeout option is available. This depends on ASIO being compiled in and works similar to the real time mode of federates. At present it is primarily for diagnostics as it cannot generate an error, though this may change at some point in the future. It is disabled by default since some federates could legitimately take a long time to execute so having a timeout on others doesn't make sense. So any grant timeout needs to be handled based on the user knowledge of how the different federates operate. This operation is subject to change but at present the operation is handled in 4 stages.
```
nightlark, line 23:
```suggestion
At the federate level a `--granttimeout` option is available. This depends on ASIO being compiled in and works similar to the real time mode of federates. At present it is primarily for diagnostics as it cannot generate an error, though this may change at some point in the future. It is disabled by default since some federates could legitimately take a long time to execute so having a timeout on others doesn't make sense. So any grant timeout needs to be handled based on the user knowledge of how the different federates operate. This operation is subject to change but at present the operation is handled in 4 stages.
```

docs/user-guide/advanced_topics/timeouts.md
```diff
@@ -0,0 +1,42 @@
...
+
+## Co-simulation operation
+
+While a cosimulation is executing, bad things can happen networks can go down, federates can crash. If left on its own this may result in a co-simulation hanging and doing nothing waiting for the failed federate. Detecting and handling this is the primary role of the heartbeat timer. If a broker has not received any communication in a tick, it sends out a ping to its parent or children. IF the parent responds then no action is taken. If no response is received in a certain period of time an error is generated and the process terminated. Sometime cores are not able to respond to pings in which case the `--slowresponding` flag should be used on that federate/core/broker to prevent co-sim termination from lack of ping responses. The `--debugging` option specified early is really a macro flag which turns on `--slowresponding` and `--disable_timer`. The primary use case is for doing step wise debugging of a federate in a code debugger, otherwise the ticks and other federates could cause issues from the timeouts and pings to the federate being debugged.
```
nightlark, line 21:
```suggestion
While a cosimulation is executing, bad things can happen; networks can go down, federates can crash. If left on its own this may result in a co-simulation hanging and doing nothing waiting for the failed federate. Detecting and handling this is the primary role of the heartbeat timer. If a broker has not received any communication in a tick, it sends out a ping to its parent or children. IF the parent responds then no action is taken. If no response is received in a certain period of time an error is generated and the process is terminated. Sometime cores are not able to respond to pings in which case the `--slowresponding` flag should be used on that federate/core/broker to prevent co-sim termination from lack of ping responses. The `--debugging` option specified early is really a macro flag which turns on `--slowresponding` and `--disable_timer`. The primary use case is for doing step wise debugging of a federate in a code debugger, otherwise the ticks and other federates could cause issues from the timeouts and pings to the federate being debugged.
```

src/helics/core/BrokerBase.hpp
phlptp: indication that this variable is connected to errorDelay,  vs relative

src/helics/core/BrokerBase.hpp
phlptp: I think I cleared this up in the latest commit

nightlark (approved):

PR #2230, opened 2021-12-09
Fix cmake HELICS_ENABLE_LOGGING bug

kdheepak:
Fixes #2229

PR #2231, opened 2021-12-09
Fix logging tsan

phlptp:
### Summary

If merged this pull request will fix the tsan warning generated from fixing #2230

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -226,6 +226,16 @@ if(NOT HELICS_DISABLE_C_SHARED_LIB OR HELICS_BUILD_CXX_SHARED_LIB
     set(CMAKE_POSITION_INDEPENDENT_CODE ON)
 endif()
 
+option(HELICS_ENABLE_LOGGING "enable normal, debug, and trace logging in HELICS" ON)
+
+cmake_dependent_advanced_option(
+    HELICS_ENABLE_TRACE_LOGGING "enable trace logging" ON "HELICS_ENABLE_LOGGING" ON
+)
+cmake_dependent_advanced_option(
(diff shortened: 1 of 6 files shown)


kdheepak (approved):

kdheepak:
I’m worried there may be other cmake related issues like this. Any ideas on how we might find them?

phlptp:
I don't know,  we have a number of CI builds with different CMake options.   I have looked around on occasion for a CMake "fuzzer" but haven't found anything like that yet.  

I was looking at it a bit closer trying to figure out why the tests passed.  What was being checked is the logging system, via api generated messages, which wasn't affected.    What the CMake flag controlled was the internally generated log messages which were being shut off.  Could probably add a test to check for them but not sure it would add much value right now.

nightlark:
What would you want a CMake "fuzzer" to do? Just run the configure step with a bunch of flag combinations (valid + invalid) and report which ones fail configuration (or produce a CMakeCache.txt file that doesn't match some expected output), or would a build + test run also need to be done?

phlptp:
For a fuzzer  the first would be to start trying a variety of CMake options in combinations.  And just trying to build.  There are several issues we had in the past that would have been caught that way.  I suppose a more time consuming step would be to change the options and run the tests make sure the tests pass under the different conditions.

nightlark:
I guess the easiest way to start out would be having the user provide a list of options that in isolation are valid, and then have it pick random combinations -- discovering options could get complicated for dependent options.

CMake provides a --trace option that shows every line executed, so a coverage tool that checks if any paths in the CMake files are skipped might be relatively easy to create.

PR #2243, opened 2021-12-16
clean up the usage of the network default ports 

phlptp:
to make it easier if we decide to use an environment variable later.

### DIFF
--- a/src/helics/network/networkDefaults.cpp
+++ b/src/helics/network/networkDefaults.cpp
@@ -14,15 +14,15 @@ int getDefaultPort(int coreType)
 {
     switch (coreType) {
         case HELICS_CORE_TYPE_TCP:
-            return network::DEFAULT_TCP_BROKER_PORT_NUMBER;
+            return network::DEFAULT_TCP_PORT;
         case HELICS_CORE_TYPE_TCP_SS:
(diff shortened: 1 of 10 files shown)


src/helics/network/networkDefaults.cpp
```diff
@@ -0,0 +1,33 @@
...
+#include "../helics_enums.h"
+
+namespace helics {
+int getDefaultPort(int coreType)
+{
+    switch (coreType) {
+        case HELICS_CORE_TYPE_TCP:
+            return network::DEFAULT_TCP_BROKER_PORT_NUMBER;
+        case HELICS_CORE_TYPE_TCP_SS:
+            return network::DEFAULT_TCPSS_PORT;
```
nightlark, line 19: Seems a bit strange using different naming conventions here -- `DEFAULT_<coretype>_PORT` instead of `DEFAULT_<coretype>_BROKER_PORT_NUMBER`. I kind of prefer the naming used for the TCPSS core, maybe aliases with the same naming could be added for the other core types in the networking namespce.

nightlark (approved):

src/helics/network/networkDefaults.cpp
phlptp: renamed them

PR #2250, opened 2021-12-30
Targeted endpoint target specification

phlptp:
### Summary

If merged this pull request will tweak the operation of `sendTo` and `sendToAt` for targeted endpoints.  Basically setting and empty destination will result in equivalent behavior to `send` and `sendAt`.  And if a destination is specified it must be in the target list,  if not an exception is generated.  

This is a change from previous behavior where those function would just generate a blanket exception for  targetedEndpoints.  This allows specific messages to be sent to specific destinations within known targets.

### DIFF
--- a/src/helics/core/CommonCore.cpp
+++ b/src/helics/core/CommonCore.cpp
@@ -1842,7 +1842,6 @@ void CommonCore::sendToAt(InterfaceHandle sourceHandle,
                           std::string_view destination,
                           Time sendTime)
 {
-    
     if (destination.empty()) {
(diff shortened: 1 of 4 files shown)


trevorhardy (approved):

trevorhardy:
I like this change as it allows the same API to be used on targeted and non-targeted endpoints when trying to send a message to a valid target. Simplifies life just a little bit.

phlptp:
There were some situations with translators that required this as well.    And this change didn't effect the timing restrictions on targeted endpoints so seemed like a good idea to relax it a little and some capabilities.

nightlark (approved):

PR #2255, opened 2022-01-06
Time monitor

phlptp:
### Summary

If merged this pull request will add an ability for brokers to monitor time on a specific federate.  Generate log messages for it and use it as the basis for the broker simulation time, this does not affect the simulation itself mainly for logs, monitoring and diagnostic purposes.  The log output is also changed such that special times will no longer display in the logs instead of time and cryptic number the current broker state will display eg.  `[connected]`,  `[terminated]`


phlptp:
need to update docs yet, but I wanted to run the code checks first

trevorhardy (approved):

PR #2256, opened 2022-01-09
Log buffer

phlptp:
### Summary

If merged this pull request will add a log buffer capability, including queries, commands, CLI to federates, brokers, and cores.


docs/user-guide/fundamental_topics/logging.md
```diff
@@ -171,6 +171,11 @@ The callback take 3 parameters about a message and in the case of `C` callbacks
 - identifier a string with the name of the object generating the message (may be empty)
 - message the actual message to log
 
+## Log Buffer
+
+As of Version 3.2
+HELICS cores brokers and federates have the capability to buffer logs messages. This can be activated via `--logbuffer` flag, or `--logbuffersize=X` option. The default size is 10 messages for the `--logbuffer` flag. For cores and federates there is a `HELICS_PROPERTY_INT_LOG_BUFFER_SIZE` property that can be set. And it can also be activated via the [`logbuffer <X>` command](../advanced_topics/commandInterface.md) remotely. `<X>` is the desired size of the buffer. `logbuffer stop` will deactivate the buffer in a remote command. The logs can be retrieved via the ["logs"](../advanced_topics/queries.md) query. The buffer is available even after disconnect for cores and brokers from the local object.
```
nightlark, line 177:
```suggestion
HELICS cores, brokers, and federates have the capability to buffer log messages. This can be activated via the `--logbuffer` flag, or `--logbuffersize=X` option. The default size is 10 messages for the `--logbuffer` flag. For cores and federates there is a `HELICS_PROPERTY_INT_LOG_BUFFER_SIZE` property that can be set. And it can also be activated via the [`logbuffer <X>` command](../advanced_topics/commandInterface.md) remotely. `<X>` is the desired size of the buffer. `logbuffer stop` will deactivate the buffer in a remote command. The logs can be retrieved via the ["logs"](../advanced_topics/queries.md) query. The buffer is available even after disconnect for cores and brokers from the local object.
```

src/helics/application_api/FederateInfo.cpp
```diff
@@ -110,6 +110,9 @@ static const std::unordered_map<std::string, int> propStringsTranslations{
     {"maxIterations", HELICS_PROPERTY_INT_MAX_ITERATIONS},
     {"intmaxiterations", HELICS_PROPERTY_INT_MAX_ITERATIONS},
     {"intMaxIterations", HELICS_PROPERTY_INT_MAX_ITERATIONS},
+    {"logbuffersize", HELICS_PROPERTY_INT_LOG_BUFFER_SIZE},
+    {"logBufferSize", HELICS_PROPERTY_INT_LOG_BUFFER_SIZE},
+    {"log_buffer_size", HELICS_PROPERTY_INT_LOG_BUFFER_SIZE},
     {"int_max_iterations", HELICS_PROPERTY_INT_MAX_ITERATIONS}};
```
nightlark, line 116:
```suggestion
    {"int_max_iterations", HELICS_PROPERTY_INT_MAX_ITERATIONS},
    {"logbuffersize", HELICS_PROPERTY_INT_LOG_BUFFER_SIZE},
    {"logBufferSize", HELICS_PROPERTY_INT_LOG_BUFFER_SIZE},
    {"log_buffer_size", HELICS_PROPERTY_INT_LOG_BUFFER_SIZE}};
```

src/helics/helics_enums.h
```diff
@@ -273,7 +273,10 @@ typedef enum {
     HELICS_PROPERTY_INT_FILE_LOG_LEVEL = 272,
     /** integer property controlling the log level for file logging in a federate see \ref
```
nightlark, line 274:
```suggestion
    /** integer property controlling the log level for console logging in a federate see \ref
```

src/helics/helics_enums.h
```diff
@@ -273,7 +273,10 @@ typedef enum {
...
        HelicsLogLevels*/
-    HELICS_PROPERTY_INT_CONSOLE_LOG_LEVEL = 274
+    HELICS_PROPERTY_INT_CONSOLE_LOG_LEVEL = 274,
+    /** integer property controlling the log level for file logging in a federate see \ref
+       HelicsLogLevels*/
```
nightlark, line 278:
```suggestion
    /** integer property controlling the size of the log buffer*/
```

src/helics/core/BrokerBase.cpp
```diff
@@ -235,7 +236,11 @@ std::shared_ptr<helicsCLI11App> BrokerBase::generateBaseCLI()
         "--dumplog",
         dumplog,
         "capture a record of all messages and dump a complete log to file or console on termination");
-
+    auto* lbs = logging_group->add_option(
+        "--logbuffersize",
+        mlogBufferSize,
+        "specify the size of the circular buffer for storing log messages for later retrieval");
+    logging_group->add_flag("--logbuffer{10}", mlogBufferSize, "specify that the ")->excludes(lbs);
```
nightlark, line 243: The help string for this flag looks like it got cut off.

docs/references/configuration_options_reference.md
```diff
@@ -308,6 +309,8 @@ String used to define the configuration of the broker if one is autogenerated. S
 - `--profiler=log` - Send the profiling messages to the default logging file. `log` can be replaced with a path to an alternative file where only the profiling messages will be sent. See the [User Guide page on profiling](../user-guide/advanced_topics/profiling.md) for further details.
 - `--timemonitor=` - Specify the name of the federate to monitor the time from and generate periodic log messages in the broker as the federate updates its time.
 - `--timemonitorperiod=` - can only be used with `--timemonitor`, set the minimum time period which must elapse in simulation before another log message from the time monitor is generated
+- `--logbuffer` - Specify that the last 10 log messages should be stored in a circular buffer to be retrieved via a query
+- `--logbuffersize=` - Specify that the last X log messages should be stored in a circular buffer to be retrieved via a query (only 1 can be used of logbuffersize, and logbuffer)
```
nightlark, line 313: Could these be combined into a single `--logbuffer[=size]` option that enables storing log messages in a circular buffer of the size provided (or 10 as a default if just the option with no size is given)?

src/helics/common/LogBuffer.hpp
```diff
@@ -0,0 +1,46 @@
...
+#include <string_view>
+#include <tuple>
+
+namespace Json {
+class Value;
+}
+
+namespace helics {
+/** defining a threadsafe small buffer for log messages the buffer will store a maximum of mMaxSize
+ * messages*/
```
nightlark, line 24:
```suggestion
/** a threadsafe small buffer for log messages that will store a maximum of mMaxSize messages*/
```

docs/references/configuration_options_reference.md
phlptp: log buffer already does that,  open to changing it if you think that would be better with just a single option.

src/helics/core/BrokerBase.cpp
phlptp: Do you think it would be clearer as just a single option?

docs/references/configuration_options_reference.md
phlptp: changed it to a single option and updated the documentation and test cases

phlptp:
they would come out in the query result in the order in which they would have been printed.    Whether that is actually chronological depends on the timing of the threads different log messages might come from.

trevorhardy (approved):

docs/references/configuration_options_reference.md
```diff
@@ -308,6 +309,7 @@ String used to define the configuration of the broker if one is autogenerated. S
 - `--profiler=log` - Send the profiling messages to the default logging file. `log` can be replaced with a path to an alternative file where only the profiling messages will be sent. See the [User Guide page on profiling](../user-guide/advanced_topics/profiling.md) for further details.
 - `--timemonitor=` - Specify the name of the federate to monitor the time from and generate periodic log messages in the broker as the federate updates its time.
 - `--timemonitorperiod=` - can only be used with `--timemonitor`, set the minimum time period which must elapse in simulation before another log message from the time monitor is generated
+- `--logbuffer` - Optionally specify the size of the circular logbuffer, defaults to 10 messages if no size is supplied.
```
nightlark, line 312:
```suggestion
- `--logbuffer` - Enable buffering recent log messages for retrieval with the "logs" query. Optionally specify the size of the circular log buffer; defaults to 10 messages if no size is supplied.
```

nightlark (approved):

trevorhardy:
If an error in the co-simulation is causing the co-simulation to crash out, do the buffered logs get printed to files? I would recommend this as a default behavior if its possible.

phlptp:
@trevorhardy  what scenarios would you be thinking of?  It something really crashes there is no code execution after the crash.

trevorhardy:
One federate crashing out, you're right; nothing to be done about it. But it will eventually take down the whole co-simulation and the cause of the crash could be rooted in something one of the other federates sent over. One of one the projects I'm working on we were just dealing with this and had to trouble-shoot it by adding print statements to source code. If we would have had the last few logging messages showing what was being sent we would have seen immediately what the cause of the problem was.

phlptp:
@trevorhardy  lets revisit this once I get the remote logging capability in as well, then maybe we can work through some scenarios and see how they play out with the new capabilities.

trevorhardy:
Works for me.

nightlark:
If the crash is on Linux with a segfault of some form (and the OS is configured to create core dumps) then it might be possible to load the core dump file into gdb and extract whatever the last few log messages in this buffer were.

Issue #2263, opened 2022-01-17
comboFed example in HELICS repo not running


trevorhardy:
**Describe the bug**
Running `./run.sh` in `/src/HELICS/examples/comboFederates1` does not successfully run the example co-simulation.

**What is the expected behavior?**
`./run.sh` would run the example comboFederate co-simulation.

**To Reproduce**
Execute `./run.sh`

**Environment (please complete the following information):**

- Operating System: Linux Ubuntu 20.04.2
- Language Extension: C++
- what compiler or setup process did you use: build from source
- HELICS version: 3.1.1. (2021-12-14)

**Additional context and information**
It is not clear to me how the shell script should be working for developers so I don't want to change things myself but the pieces are all there. `comboFed` got built (but maybe not installed) when I built with examples on and the helics_broker built and installed successfully. Run individually, none of the command lines in the shell script successfully launch a federate.

```
$ /comboFed --name fed1 --target fed2 >fed1.out
The following arguments were not expected: fed1 --name
```
```
$ helics_broker 2 --loglevel=warning >broker.out
The following argument was not expected: 2
```

trevorhardy:
Needed a small patch in the comboFederate1 source before parsing the command line options to allow the `--name`` option to work.

```
app.allow_extras();
```

trevorhardy:
I'm re-writing the shell scripts so that they run under the assumption that helics_broker is in PATH. This will allow a test of the installation rather than just the build. (Up to this point the shell scripts have been unused so I'm assuming this won't be a problem for anybody.)

phlptp:
Can this be closed now?

trevorhardy:
Fixed in PR #2266 and 2267.

PR #2264, opened 2022-01-17
Update documentation on Ubuntu build

trevorhardy:
### Summary
Small documentation changes to fix the instructions for installing under Ubuntu. Small changes to get the comboFederate example working to validate the installation.

Fixes issue #2263.

A similar PR will be made provide these fixes to develop. This PR is being applied directly to main so that the public documentation will be fixed more quickly.


phlptp:
I think you need to start with main and make the new changes.   Looks like you are merging everything from develop into this which is not what we want to do right now

trevorhardy:
Good point. I'll kill this PR and do another that way.

trevorhardy:
Closing PR, will re-open with the same fixes applied to develop (from which this branch was created.)

PR #2286, opened 2022-02-15
Update gmlc::networking to version with encryption support

nightlark:
### Summary

If merged this pull request will update the gmlc::networking library to a version that includes support for encrypted sockets.

### Proposed changes

- Update gmlc::networking library

### DIFF
--- a/.ci/azure-pipelines.yml
+++ b/.ci/azure-pipelines.yml
@@ -52,6 +52,10 @@ jobs:
           ZMQ_FORCE_SUBPROJECT: $[variables['zmq_subproject']]
         displayName: 'Build HELICS'
 
+      - bash: ldd bin/helics_broker
+        workingDirectory: 'build'
+        displayName: '[DEBUG] Check libraries linked to HELICS'
+
       - bash: ../scripts/run-ci-tests.sh "$TEST_CONFIG"
         env:
           TEST_CONFIG: $[variables['test_config']]
--- a/.gitmodules
+++ b/.gitmodules
@@ -30,3 +30,4 @@
 [submodule "ThirdParty/networking"]
 	path = ThirdParty/networking
 	url = ../../GMLC-TDC/networking.git
+	branch = main


nightlark:
Updating the gmlc::networking submodule and making sure CI builds work (aside from flaky tests). Subsequent PR will make the changes needed to turn on encrypted comms.

CMakeLists.txt
```diff
@@ -713,10 +713,8 @@ set(GMLC_NETWORKING_CONCURRENCY_INCLUDE ${PROJECT_SOURCE_DIR}/ThirdParty/concurr
 set(GMLC_NETWORKING_DISABLE_ASIO ${HELICS_DISABLE_ASIO})
 
 if(HELICS_ENABLE_ENCRYPTION)
-    set(HELICS_ENABLE_ENCRYPTED_COMMS 1)
```
nightlark, line 716: This step was happening after the helics-config.h file is already generated without the define -- also seemed a bit redundant having an extra variable with the same meaning.

src/helics/network/tcp/TcpCommsSS.cpp
```diff
@@ -151,6 +151,8 @@ void TcpCommsSS::queue_tx_function()
     }
     gmlc::networking::TcpServer::pointer server;
     auto ioctx = gmlc::networking::AsioContextManager::getContextPointer();
+    auto sf = encrypted ? gmlc::networking::SocketFactory(encryption_config) :
```
nightlark, line 154: The downside of reading the encryption config file here is that a user could theoretically change the contents of the config file before it gets used to create a socket.

CMakeLists.txt
phlptp: The config.h does use this, so we might need to move this up in the CMAKE so it gets used by the config.h generation.  Since I think the code does need to know if the encrypted option is even available.

src/helics/network/tcp/TcpCommsSS.cpp
phlptp: Even if they did, not sure why that would be a problem.

CMakeLists.txt
phlptp: Have you verified that option turns on and off properly?

phlptp:
looks like the change to the app caused some failures.  I am working on another that might resolve it in a better way

CMakeLists.txt
nightlark: I changed it because before it was always showing the error message about HELICS not being compiled without encryption support, even when the option was turned on.

In the CI builds there are a mix of Linux jobs that have the option ON/OFF, and earlier builds for this PR printed the libraries linked to one of the HELICS binaries.

gcc11 build linked with OpenSSL: https://dev.azure.com/HELICS-test/HELICS/_build/results?buildId=11666&view=logs&jobId=0cbb91d6-5794-5751-6296-8904318eba52&j=0cbb91d6-5794-5751-6296-8904318eba52&t=f7b3e03b-d64b-5c39-d720-8c777666cf2a

clang13 build without OpenSSL: https://dev.azure.com/HELICS-test/HELICS/_build/results?buildId=11666&view=logs&jobId=0cbb91d6-5794-5751-6296-8904318eba52&j=ec8c292a-0c4e-5923-6a96-214ddcb87113&t=0bd14658-bf30-5e59-496c-3881c52cd135

phlptp:
#2287 should resolve the argument parsing and still leave the positional as is.

nightlark:
This merges cleanly with the argument parsing change in #2287.

phlptp:
@nightlark  do you want to rebase this then I think it would be ready to merge

nightlark:
Rebase didn't want to happen -- the  commit with the positional arg parsing changes was causing conflicts (despite being later reverted). Squash and merge should get rid of the merge commits and rebase in the develop branch history.

phlptp (approved):

phlptp:
I merged this for now.  We should probably make either a test case or example that actually builds and attempts a full encrypted communication path through HELICS

PR #2289, opened 2022-02-22
add the start of the swagger documentation for the HELICS REST API

phlptp:
### Summary

If merged this pull request will add swgger files for the REST API.  There is more to come so this is main just the start and to start working with the files to begin the process of how to display and host them.


phlptp:
@nightlark @trevorhardy @kdheepak 
Can you take a look at this,  I think this is the extent of what I can get done for the 3.2 release.  I expect it will be updated as revised in later updates

docs/swagger/models/BrokerCreation.yaml
```diff
@@ -0,0 +1,35 @@
+title: BrokerCreation
+type: object
+properties:
+  type:
+    type: string
+    description: the HELICS core type to use when creating a broker
+  CoreType:
+    type: string
+    description: same as type
```
trevorhardy, line 9: I don't know what "same as type" means.

docs/swagger/models/FederateInfo.yaml
```diff
@@ -0,0 +1,60 @@
+title: FederateInfo
```
trevorhardy, line 1: This is going to be a beast, right? So many options to support or do we have a better way?

trevorhardy (approved):
It would be great to come up with a better way to view the swagger API stuff.

docs/swagger/models/FederateInfo.yaml
phlptp: All the specific options that are not flags are listed already.   I will probably expand this later to include the flags.
This is eventually meant to document the json configuration files but not fully there yet.

nightlark:
Here's a sphinx extension for redoc:
https://sphinxcontrib-redoc.readthedocs.io/en/stable/

And their demo of a generated documentation page (on readthedocs):
https://sphinxcontrib-redoc.readthedocs.io/en/stable/api/github/#

RTD conf.py script for their example build:
https://github.com/sphinx-contrib/redoc/blob/master/docs/conf.py

phlptp:
Should we try to make those changes for swagger working with sphinx in this PR or a subsequent one?

trevorhardy:
@nightlark, do you want to take a crack at this or should I? I can probably put a little bit of time into it in the next day or two if you can't/don't.

phlptp:
@trevorhardy  if you want to take a crack at it go ahead.  It is going to be at least a few days yet before a release so we can reassess once a release is immanent if we are not making much progress.

nightlark:
I think we can merge this PR and do the RTD changes as a separate PR.

docs/conf.py
```diff
@@ -170,6 +171,17 @@ def ext_candidates(fpath):
 # If true, `todo` and `todoList` produce output, else they produce nothing.
 todo_include_todos = False
 
+# redoc config
+redoc = [
+    {
+        "name": "HELICS REST API",
+        "page": "references/api-reference/rest_queries_api",
```
nightlark: I think it would be better to generate this in the `swagger` folder.
```suggestion
        "page": "swagger",
```

And then also add a directive to RTD saying to include the contents of the `swagger` folder so that all the yaml/json specs and model files are copied to the website.

trevorhardy: @nightlark, If you think you know how to get this working, be my guest. I'm kind of guessing at how to get all the pathing right.

nightlark: I need to look into the redoc settings, but I think the original pathing and generated swagger files were fine. The two issues I think for why it didn't work are:

- RTD conf.py needs to specify that the `swagger` folder with model yaml files be included in the deployed site; embed is set to True for redoc, so it stores a copy of queries.yaml in the generated api page but the stored copy has paths referring to the other yaml files which aren't included on the RTD site
- embed should probably be left as the default value False, making so that the relative paths in queries.yaml point to the right place; when queries.yaml is embedded it looks like the generated in an arbitrary folder that breaks the relative paths

I'll see if I can try to get it working later today or tonight, if it isn't already working.

trevorhardy: I'll rollback to the an earlier version and try just turning `embed` off.

trevorhardy: I do think the `swagger` folder needs to end up in the `api-reference` folder but we can tweak that once we get things working.

nightlark: The swagger folder doesn't need to end up in `api-reference` (we'll do the same thing with it as we do the doxygen folder).

https://github.com/GMLC-TDC/HELICS/blob/swagger/docs/conf.py#L104
I think changing this line from:
```
html_extra_path = [os.path.abspath(os.path.join(doxygen_build_dir, "docs", "html"))]
```
to
```
html_extra_path = ["swagger", os.path.abspath(os.path.join(doxygen_build_dir, "docs", "html"))]
```

will include the swagger folder in the right place.

nightlark: From my understanding of how redoc works, it looks like reverting the changes back to commit [ded62ea](https://github.com/GMLC-TDC/HELICS/pull/2289/commits/ded62eae63cc76d02432ce27ba63226cc4ab0ebc), making the `html_extra_path` change, and turning `embed` off could work.

trevorhardy: Well, [this link](https://docs.helics.org/en/swagger/references/api-reference/rest_queries_api.html) still returns a 404. I'll see if I come up with any other great ideas.

nightlark: Yeah -- it looks like https://docs.helics.org/en/swagger/swagger/models/Filter.yaml doesn't exist, so I think it didn't copy the folder. Might be that the absolute path to the swagger folder needs to be given instead to make RTD include it... or relative paths are to something like the repository root instead rather than the docs subfolder.

phlptp:
@nightlark @trevorhardy  where are we at with this,  I know you were working on figuring out how to get it built into the readthedocs?  any Luck?

trevorhardy:
@phlptp I haven't been able to get it working and haven't touched it since last week. I was hoping @nightlark might work his magic...

nightlark:
I got a local build working (virtualenv folder was in the docs folder, so it died when trying to parse some random markdown/rst files included in pip packages).

So far, the smaller issue is that including the `swagger` folder is just copying the contents of `swagger` into the root html output folder; I think I can work around that by creating an `doc_extras` folder and move `swagger` into that.

The bigger issue is the sphinx redoc extension -- it has no option to tell it not to move the included OpenAPI/swagger spec file when that is provided as a relative path; and any movement of the API spec file relative to the other swagger files will break relative references. The three possible fixes I'm seeing are:
- making a patched version of the sphinx redoc extension with options to give more control over placement of the final spec file; modified version of the extension would need to be uploaded to PyPI as it seems the original maintainer isn't able to maintain it anymore: https://github.com/sphinx-contrib/redoc/issues/31#issuecomment-605467405)
- turning on the embed option and trying to coerce it into placing the final html file in the exact folder as the rest of the swagger model files
- getting a JSON file similar to `swagger/references/web-server.json` to be used instead of `queries.yaml` with a bunch of referenced yaml files, since the JSON file has all the needed info with no external references (@phlptp is there an option in the tool you were using that can do this?); with the redoc embed option turned on, this should just work, so I think it would be the preferred option

phlptp:
we can probably convert queries.yaml to a json file, but it would still make reference to all the model files.  Since those model files are used in multiple locations, it would be bad software practice to try to mash everything into a single file, so I don't want to do that.  And in that case I am not sure converting to a json would make much difference.

nightlark (approved):
Swagger docs build using redoc on RTD -- link to the generated redoc page for the REST Queries API is on the page under References->API References. I think there are some issues related to the yaml models, which result in warnings being shown.

PR #2290, opened 2022-02-25
Fix missing timing dependency when subscribing from an endpoint

phlptp:
### Summary

If merged this pull request will fix a timing issue when subscribing to publications from an endpoint


trevorhardy:
I just did the testing I suggested on my federation (subscribing endpoint requesting infinite time to make sure it is granted time at the new publication) and it works for me. Probably still a good idea to add testing for it.

PR #2294, opened 2022-02-28
Fix benchmark builds

nightlark:
### Summary

If merged this pull request will fix the failing benchmark builds.

Successful benchmark build workflow run with these changes: https://github.com/GMLC-TDC/HELICS/actions/runs/1912091746
Successful release build workflow run with these changes: https://github.com/GMLC-TDC/HELICS/actions/runs/1912092086

### Proposed changes

- Pins the Windows benchmark build to windows-2022 VM instead of windows-latest (to ensure cmake generator to use also gets changed at the same time)
- Changes the macOS SDK install script to ensure that a macOS SDK version that matches the deployment target gets installed (avoids a newer SDK being installed while trying to build a program that targets an older SDK version)
- Passes the CPack generator used in matrix strategy entries to the Linux builder local GitHub action as an environment variable, so the benchmark builds for Linux only include a .tar.gz installer (slims down the  artifact storage requirements for Linux from 100+ MB to 30MB)


.github/actions/linux-release-builder/entrypoint.sh
```diff
@@ -1,3 +1,8 @@
 #!/bin/bash
 
+if [[ ! -z "${INPUT_CPACK_GEN}" ]]; then
```
phlptp, line 3: looks like a shellcheck issue on this line

.github/actions/setup-macos-sdk.sh
```diff
@@ -3,13 +3,15 @@
 XCODE_MACOS_PLATFORM_PATH="/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform"
 XCODE_MACOS_SDK_PATH="${XCODE_MACOS_PLATFORM_PATH}/Developer/SDKs"
 
-SDK_VER=10.14
+if [[ -z ${MACOS_SDK_VER+x} ]]; then
+    MACOS_SDK_VER=10.14
+fi
```
phlptp, line 8: Am I reading this correctly that this is stating that if the string MACOS_SDK_VER is empty then define it.  What is the +x for?

nightlark: The +x is some bash parameter substitution; doesn't look like it is needed here and works fine without it.

The default macOS SDK version we're targeting is 10.14; all our uses of this script should be defining MACOS_SDK_VER already (to 10.14)

phlptp (approved):

PR #2296, opened 2022-02-28
Remove VS2017 CI builds and add a VS2022 build

nightlark:
### Summary

If merged this pull request will remove the VS2017 CI builds from Azure Pipelines and the release builds workflow. A VS2022 is added to the Azure Pipelines CI jobs. Installation docs are updated to recommend VS2019. CI documentation pages have some changes made to remove mentions of VS2017 and add VS2022, though those pages seem like they could use more extensive updates.

### Proposed changes

- Remove VS2017 CI build from Azure Pipelines
- Remove VS2017 release build
- Add VS2022 CI build to Azure Pipelines (release builds already include VS2022)


docs/user-guide/installation/windows.md
```diff
@@ -89,8 +89,8 @@ Getting and building from source:
     If you need CMake to use a generator other than the default (ex:
     selecting between a 32-bit or 64-bit project), the -G option can be
     used to specify one of the generators listed by CMake --help. For
-    Visual Studio 2017, the generator name would be
-    `Visual Studio 15 2017 [arch]`, where \[arch\] is optional and can be
+    Visual Studio 2019, the generator name would be
+    `Visual Studio 16 2019 [arch]`, where \[arch\] is optional and can be
```
phlptp, line 93: I think the default changed to 64 bit on visual studio 2019 generator for cmake, so you need to specify 32 bit if you want that.

nightlark: It looks like this form of specifying the Win64 architecture is no longer supported in CMake as of VS2019, and specifying -A is the way to choose it now (with the default being the one matching the host): https://cmake.org/cmake/help/latest/generator/Visual%20Studio%2016%202019.html

phlptp (approved):

nightlark:
Update after #2298 is merged and re-run tests.

PR #2306, opened 2022-03-10
Combo file test for CPP98

phlptp:
### Summary

If merged this pull request will fix a missing call in the C++98 interface, and resolve some discrepancies' between subscription blocks and input blocks for input files.  

### Proposed changes

- add input test file
- resolve differences in block interpretation between input and subscription blocks
- add combo file tests
- add getFilterCount() to Cpp98 API

### DIFF
--- a/src/helics/application_api/ValueFederate.cpp
+++ b/src/helics/application_api/ValueFederate.cpp
@@ -371,10 +371,9 @@ void ValueFederate::registerValueInterfacesToml(const std::string& tomlString)
 
                 id = &registerInput(emptyStr, type, units);
                 if (!skipNameTarget) {
-                    //this check is to prevent some warnings since targets get added later
+                    // this check is to prevent some warnings since targets get added later
                     id->addTarget(name);
                 }
-                
             }
 
             loadOptions(this, sub, *id);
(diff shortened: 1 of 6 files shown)


src/helics/application_api/ValueFederate.cpp
```diff
@@ -264,14 +264,22 @@ void ValueFederate::registerValueInterfacesJson(const std::string& jsonString)
     if (doc.isMember("subscriptions")) {
         auto subs = doc["subscriptions"];
         for (const auto& sub : subs) {
+            bool skipNameTarget{false};
             auto name = fileops::getName(sub);
+            if (name.empty()) {
```
afisher1, line 269: What happens if name is not empty? I tried testing this also by setting a name in the subscriptions. Things still seemed to work pub/sub wise, but helicsSubscriptionGetTarget(ipt) returned what I put in the name key and not what I put in the target key.

afisher1 (changes requested):

src/helics/application_api/ValueFederate.cpp
phlptp: For subscriptions the key is the target,  so if you specify a name or a key that is the first target.  anything in target(s) field gets added to it.  

So unless you intentionally want to subscribe to multiple targets you really should only specify a key or a target.

afisher1 (approved):

PR #2310, opened 2022-03-12
Update Docker images and migrate builds to GitHub Actions

phlptp:
### Summary

If merged this pull request will update the docker images and add some github actions to build them and add to dockerhub


nightlark (commented):
We should combine the workflows, since they are nearly identical.

.github/actions/docker/main_docker.sh
```diff
@@ -0,0 +1,14 @@
+#! /bin/sh
+# Require a valid Git ref as an argument
+if [ "$#" -ne 1 ] || ! git rev-parse --verify "$1" >/dev/null; then
+    echo "Usage: $0 GIT_REF TAG_NAME" >&2
```
nightlark, line 4: These arguments aren't actually getting used, are they?

phlptp:
I was trying to figure out if we could combine them but wasn't sure how to discriminate the tags and triggering since they all have different triggers and go to different docker tags and the scripts use the different DockerFiles

.github/actions/docker/main_docker.sh
phlptp: I am not 100% sure what this does.  I got the impression it verified that this was actually part of the git repo to ensure that the right structure was in place.    I copied it from the buildenv repo

nightlark (approved):
I combined the workflows into a single workflow file, and made some tweaks to the workflow_dispatch to give some control over what test builds will do.

The docker/metadata-action tries to generate a set of labels/images that can be consumed by the docker/build-push-action -- I don't want to start using that yet, but I want to see if the set of labels/tags generated by the metadata action looks okay for the various types of builds (release, push main, push develop, push test). If it does, I think switching from the bash scripts to the docker push and build action will provide an easy route to pushing images to both Docker Hub and the GitHub Container Registry. It could also make it easy to build multi-arch images that also work on ARM.

PR #2313, opened 2022-03-18
Windows build instructions

eranschweitzer:
### Summary

If merged this pull request will improve the documentation for how to build custom branches of HELICS on windows using Visual Studio and linking the custom `helics.dll` to pyhelics.

### Proposed changes

Documentation changes only.

### DIFF
--- a/docs/user-guide/installation/windows.md
+++ b/docs/user-guide/installation/windows.md
@@ -18,18 +18,18 @@ _Note_: Keep in mind that your CMake version should be newer than the boost vers
 
 ### Set up your Environment:
 
-**1.**  Install Microsoft Visual C++ 2019 or newer (2017 may work, but is no longer tested by CI builds)[MSVC](https://visualstudio.microsoft.com/)
+**1.** Install Microsoft Visual C++ 2019 or newer (2017 may work, but is no longer tested by CI builds)[MSVC](https://visualstudio.microsoft.com/)
 
-**2.**  Install
-    [Boost](https://www.boost.org/doc/libs/1_74_0/more/getting_started/windows.html)
-    [Windows downloads](https://dl.bintray.com/boostorg/release/1.74.0/binaries/)
-    1.67 or later recommended. For CMake to detect it automatically either
-    extract Boost to the root of your drive, or set the `BOOST_INSTALL_PATH`
-    environment variable to the install location. The CMake will only automatically find
-    Boost 1.67 or newer.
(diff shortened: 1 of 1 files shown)


nightlark:
The PR target branch should be changed to `develop` instead of `main` (I can't seem to do this from the GH mobile iOS app).

There are a few places with system paths listed that include a user name (powershell "prompt"; could probably just remove the current directory path and make it `>` or try to make the path something generic).

trevorhardy (approved):
I did a little clean-up (mostly formatting).

eranschweitzer:
Thanks @trevorhardy! Based on your corrections and comments from @nightlark it seems like there is nothing more for me to do 😃 (though any other change requests are welcomed).

nightlark (commented):
I think removing the leading space before all of the lines that fall within a numbered bullet point has messed up some of the formatting? The RTD/GitHub markdown parsers no longer consider any of the lines after the first one to be part of the contents of the numbered list entry they fall under.

When that formatting is correct, I don't think bolding the numbers is necessary since the start and end of the content for each step stands out based on the indentation. The bolded numbered list style also differs from every other numbered list in the docs -- if that's something we want to do, I think bolding numbers would be better handled by changing the CSS used by the RTD theme.

eranschweitzer:
@nightlark, the number formatting was not working correctly already, because the numbering is reset every time an item is longer and or includes something like code block.

If you look at the *current* documentation: https://docs.helics.org/en/latest/user-guide/installation/windows.html
you'll see that there are numbers 1-5, then it resets and we have 1-2, then 1-2 again, then two 1s in a row and another 2.

I've run into this problem many times and at some point just gave up on trying to fix it.
My solution is to *go around* the automatic numbering and create the numbered list manually.
I'm not married to the formatting, we can remove the bold etc. However, what I don't know is how to get the indent the way the automatic numbered lists gets it.

I'm open to any and all suggestions but I don't know how to get the number list to behave in markdown the way that you would like it to.

phlptp (approved):

nightlark:
Oh this is one of the pages that already had broken formatting; let's skip fixing it and it can be fixed in some future pass that includes the other doc pages with formatting issues.

The numbers resetting are almost always from missing indentation after a line split or the code block (\`\`\`) directives not being indented.

* Not indented
```bash
    git clone ...
```
* Indented
  ```bash
  git clone ...
  ```
---
    * Not indented
    ```bash
      git clone ...
    ```
    * Indented
      ```bash
      git clone ...
      ```

nightlark (approved):

PR #2319, opened 2022-03-24
Use external libraries

### LINKED ISSUE #2297: Ensure external spdlog library can be used
A user was having issues with the HELICS copy of spdlog interfering with other copies of spdlog. Make sure there is a way in HELICS to use an external/system copy of spdlog. Specifically, this needs to work when using spack.

nightlark:
### Summary

If merged this pull request will allow users to use external copies of the {fmt}, spdlog, jsoncpp, and units libraries with HELICS instead of the copies included in the HELICS repo as submodules. The `HELICS::` namespaced aliases for some of these libraries weren't being exported with installs, and creating the ALIAS doesn't work with imported libraries so they were removed in favor of using the namespaced variant that already exists from find_package or add_subdirectory.

### Proposed changes

- Add `HELICS_USE_EXTERNAL_FMT`, `HELICS_USE_EXTERNAL_SPDLOG`, `HELICS_USE_EXTERNAL_JSONCPP`, and `HELICS_USE_EXTERNAL_UNITS` options for using external copies of these libraries
- Document use of the added options for using external libraries
- Update paths for including the units header files to be a more focused match (using the include path from the include path associated with the units target, instead of the top level ThirdParty/ folder)
- Remove `HELICS::spdlog` and `HELICS::units` aliases in favor of just using the already existing namespaced aliases `spdlog::spdlog` and `units::units`
- Create and use the `HELICS::jsoncpp` ALIAS when the `JsonCpp::JsonCpp` target doesn't exist (which indicates that the vendored copy of jsoncpp is being used; the scope of the option to use the external jsoncpp didn't allow using elsewhere for these checks)

Closes #2297.
A step towards meeting criteria M21 for #821 xSDK compatibility.


docs/user-guide/installation/helics_cmake_options.md
```diff
@@ -73,6 +73,19 @@ Options effect the connection of libraries used in HELICS and how they are linke
...
 - `HELICS_ENABLE_CLANG_TOOLS`: \[Default=OFF\] Enables some helper targets for using clang-tidy and clang-format.
 
+#### Options related to using external/vendored libraries
+
+These options are for making HELICS look for an external installation or system copy of libraries instead of using the vendored copy included with HELICS. The search process for locating the external copy of libraries follows the typical [CMake Config Mode Search Procedure](https://cmake.org/cmake/help/latest/command/find_package.html#config-mode-search-procedure). Most users should leave these options `OFF`.
+
+When enabling these options, if the static library variant is found it _must_ have been compiled using the option `CMAKE_POSITION_INDEPENDENT_CODE=ON` or _you will encounter linking errors_.
+
+Furthermore, there are no guarantees that HELICS will compile using arbitrary versions of these libraries other than the exact version that is included as a vendored library in the HELICS repository.
+
```
phlptp, line 83: Should we add any comments about which versions of these libraries will likely work?

phlptp: Are we testing any of this in any CI builds (likely not)  and should we be?

nightlark: I don't want to list any exact versions, since as I found while testing the feature, the version in development of `fmtlib` that comes after 8.1.1 (the tagged version corresponding to the current submodule git commit hash) results in compiler errors about `const` being lost when using `this`.

I added more information on getting what the current commit hashes for submodules are; hopefully the use cases for these options are fairly advanced scenarios, so users should already have some idea how to troubleshoot issues that arise from conflicting versions.

nightlark: These aren't tested in any CI builds. Some testing should eventually be figured out, but it needs to use an install of the exact same version that we vendor or we might encounter situations where a minor change in a new version breaks the external library build but not the vendored build (or the other way around).

PR #2321, opened 2022-03-24
Fix missing submodule source code in the HELICS release source archive

nightlark:
### Summary

If merged this pull request will fix the script that generates the HELICS source archive that includes submodule code for the HELICS releases.

### Proposed changes

- recursive git submodule update to ensure submodules used by dependencies also get checkedout (e.g. NetIF in the `networking` submodule)
- use `$displaypath` instead of `$path` (or `$sm_path`, https://git-scm.com/docs/git-submodule/2.24.0 has more info on these variables) for the prefix, so the full relative path (from the HELICS repo root folder) to submodule folder is preserved (so that e.g. `ThirdParty/networking/ThirdParty/concurrency` is unpacked to `ThirdParty/networking/ThirdParty/concurrency` instead of `ThirdParty/concurrency`; or so `ThirdParty/networking/ThirdParty/NetIF` doesn't get unpacked to `ThirdParty/NetIF`)
- include `$displaypath` derived string in archived folder name for submodules, to avoid duplicates if both a direct HELICS submodule and a nested submodule are identical
- update name of `toml` submodule directory (toml11->toml) for slimming/cleanup step
- remove duplicate `asio` and `concurrency` submodules from the `networking` submodule; they aren't needed since the `networking` submodule is told to use the `asio` and `concurrency` modules that HELICS includes as immediate submodules

### DIFF
--- a/scripts/_git-all-archive.sh
+++ b/scripts/_git-all-archive.sh
@@ -57,7 +57,7 @@ if [[ -n "$tag" ]]; then
 fi
 
 # Get git submodule source code
-git submodule update --init
+git submodule update --init --recursive
 
 # Set the output name based on if a release name was provided
 OUTPUT_BASENAME="Helics-source"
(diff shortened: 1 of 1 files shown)


nightlark:
Artifacts for a manually triggered test run of the release workflow are at: https://github.com/GMLC-TDC/HELICS/actions/runs/2040522934

scripts/_git-all-archive.sh
```diff
@@ -116,6 +118,8 @@ rmdir_list=(
     'concurrency/docs'
     'concurrency/.ci'
     'concurrency/.circleci'
+    'networking/ThirdParty/concurrency'
+    'networking/ThirdParty/asio'
```
phlptp, line 122: did we check if any of the other libraries have submodules?

phlptp: also could probably remove the concurrency and networking tests directories as well right?

nightlark: I don't think other libraries had submodules -- if they do then they are probably required, since I think in HELICS the only special case was telling gmlc::networking to use the immediate concurrency and asio submodules from HELICS instead of the ones that are submodules of networking.

nightlark: `concurrency/tests` is already removed, line 116.

phlptp (approved):

PR #2322, opened 2022-03-27
Add ability to (re)build tags using updated workflow and actions/shell scripts

nightlark:
### Summary

If merged this pull request will replace the functionality of the repository_dispatch trigger with the more easily used workflow_dispatch, and enable rebuilding release packages/tags with an updated release build workflow and shell scripts from another branch. This enables more easily fixing issues where something has been left out of the source archive or enabled CMake options for a binary package weren't right.

The new steps added make the initial checkout do a full clone (fetch-depth = 0) if a workflow dispatch event input specifies a commit to rebuild. The initial code checkout had to be the workflow file branch/tag -- this convoluted order was because checking out the "commit to rebuild" first resulted in git being weird and not recognizing the branch name it should be checking out the build related shell scripts from.

### Proposed changes

- Move repository_dispatch functionality to workflow_dispatch
- Add capability to build a package using the code from a tag, but the updated release build workflow and actions/shell scripts

### DIFF
--- a/.github/workflows/release-build.yml
+++ b/.github/workflows/release-build.yml
@@ -1,6 +1,7 @@
 name: Build release artifacts
 
 on:
+  # Picking a workflow branch that is a tag, in combination with using the commit field is likely to fail for magic git clone/checkout reasons
   workflow_dispatch:
     inputs:
       packages:
@@ -8,7 +9,7 @@ on:
         required: true
         default: 'everything'
       commit:
-        description: 'Commit-ish (tag, branchname, sha) to do the release build with (uses scripts from the same branch as the workflow)'
+        description: 'Commit-ish (tag, branchname, sha) for code to build (uses scripts from the same branch as the workflow)'
         default: ''
   schedule:
     - cron: '15 09 * * *' # Run in the early hours of the morning (UTC)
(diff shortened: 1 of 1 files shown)


.github/workflows/release-build.yml
```diff
@@ -1,12 +1,16 @@
...
+    inputs:
+      packages:
+        description: 'List of packages to build (everything, archive, msvc, installer, sharedlib)'
+        required: true
+        default: 'everything'
+      commit:
+        description: 'Commit-ish (tag, branchname, sha) for code to build (uses scripts from the same branch as the workflow)'
+        default: ''
   schedule:
     - cron: '15 09 * * *' # Run in the early hours of the morning (UTC)
```
phlptp, line 15: do we do anything with these builds?

.github/workflows/release-build.yml
```diff
@@ -19,25 +23,42 @@ jobs:
...
+    - name: Determine fetch depth
+      id: determine_fd
+      shell: bash
+      run: |
+        unset DEPTH
+        if [ -n '${{ github.event.inputs.commit }}' ]; then DEPTH='0'; else DEPTH='1'; fi
+        echo "::set-output name=DEPTH::${DEPTH}"
+
+    - name: Checkout code
+      uses: actions/checkout@v3
```
phlptp, line 37: are there other places we need to move to v3 of the checkout action?

.github/workflows/release-build.yml
```diff
@@ -1,12 +1,16 @@
...
 on:
-  # Handles a 'helics-release-build' dispatch event type
-  # The event payload should contain these fields:
-  # packages - a list of which packages to build (everything, archive, msvc, installer, sharedlib)
-  # commit (optional) - the commit-ish to do a release build with
-  repository_dispatch:
+  # Picking a workflow branch that is a tag, in combination with using the commit field is likely to fail for magic git clone/checkout reasons
   workflow_dispatch:
+    inputs:
+      packages:
```
phlptp, line 7: where does this input get specified?  Is this a manual option?  if so are the listed options here all the recognized ones?

nightlark: It shows up when you go to manually trigger a GitHub workflow (only shows up when the workflow file in the default/main branch has a list of inputs). See https://github.com/GMLC-TDC/helics-action/actions/workflows/test-helics-install.yml for an example.

The options listed are the words that will be recognized.

.github/workflows/release-build.yml
nightlark: Catching changes that break the automated release builds, hopefully while they are still in develop and before the pending release has been merged into main.

phlptp (approved):

.github/workflows/release-build.yml
nightlark: It just uses a newer version of node.js, no major changes and v2 still works fine. I'm looking at setting up Dependabot to keep the versions of actions used updated so that should take care of upgrading other places eventually.

PR #2329, opened 2022-04-09
Global time coordinator

phlptp:
### Summary

If merged this pull request will add a global time coordinator and merge in the time request verify code as well

PR #2332, opened 2022-04-27
Sequencing fixes

phlptp:
### Summary

If merged this pull request will add a retry in case of try_lock spurious failure

### DIFF
--- a/src/helics/core/FederateState.cpp
+++ b/src/helics/core/FederateState.cpp
@@ -679,9 +679,6 @@ iteration_time FederateState::requestTime(Time nextTime, IterationRequest iterat
             allowed_send_time = timeCoord->allowedSendTime();
             iterating = (ret == MessageProcessingResult::ITERATING);
         }
-        if (lastTime == time_granted) {
-            std::cerr << "equivalent Times new and old " << lastTime << time_granted << std::endl;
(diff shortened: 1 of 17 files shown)


phlptp:
This will get rebased once #2329 is merged, but pr for testing

src/helics/core/FederateState.cpp
```diff
@@ -709,6 +720,16 @@ iteration_time FederateState::requestTime(Time nextTime, IterationRequest iterat
         }
         return retTime;
     }
+
+    // if this is not true then try again the core may have been handing something short so try
```
nightlark, line 724:
```suggestion
    // if this is not true then try again the core may have been handling something short so try
```

src/helics/core/FederateState.cpp
```diff
@@ -786,7 +807,14 @@ MessageProcessingResult FederateState::genericUnspecifiedQueueProcess(bool busyR
         unlock();
         return ret;
     }
-
+    // if this is not true then try again the core may have been handing something short so try
```
nightlark, line 810:
```suggestion
    // if this is not true then try again the core may have been handling something short so try
```

nightlark (approved):

PR #2339, opened 2022-05-04
Ipc update

phlptp:
### Summary

If merged this pull request will update the IPC core to correct some deficiencies in the message size and capacity


phlptp:
This will probably be a long running PR, as it will take a while to get everything working again.

PR #2356, opened 2022-05-18
add protection functions in the C API 

phlptp:
so federates can be found later if freed but not destroyed

### Summary

If merged this pull request will add helicsFederateProtect and helicsFederateUnProtect

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -150,6 +150,15 @@ if(CMAKE_PROJECT_NAME STREQUAL PROJECT_NAME)
     endif()
 endif()
 
+if(NOT CMAKE_VERSION VERSION_LESS 3.12)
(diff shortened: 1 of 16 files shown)


phlptp:
@kdheepak   Take a look at this.  A couple questions should the protection be counting or flag like?  meaning you need equal number of unprotect calls as protect calls.

phlptp:
What this also means is the destructor in wrapper objects should not call finalize/disconnect.    They should just call helicsFederateFree.  

Also there is no protect on the core or brokers since they were meant from the beginning to be in the background most of the time and the reteival mechanism is different from federates so would work as expected.

phlptp:
@kdheepak  can you look through this.  I will fix the clang tidy issues yet but want you to look over the API

kdheepak (approved):

PR #2361, opened 2022-05-24
Use linker option to exclude external library symbols from the C shared library

nightlark:
Changes symbols from external libraries (spdlog, fmt, etc) to be excluded from the HELICS C shared library, so only symbols for functions that we intentionally export from the library are included. Reduced the C shared library size by a bit over 1MB (15-20% reduction in overall size).


phlptp (approved):

phlptp:
I think I will try to do this to the mac build as well,  try making an export list with all the symbols.  It looks like they can use wildcards so it might not be too hard.

PR #2379, opened 2022-06-19
long_file_name_removal

phlptp:
### Summary

If merged this pull request will remove the very long file name from the source directory and have it be automatically generated by cmake


nightlark (approved):

nightlark:
Note -- this PR fixes a bug with creating the HELICS source tar archive with all submodules for releases; the `tar` command only supports file path names up to 256 characters ([but only if the filepath can be split at a directory separator into two parts, with the first part being up to 155 characters long](https://www.gnu.org/software/tar/manual/html_node/Formats.html#:~:text=restrictions%20as%20well%3A-,The%20maximum%20length%20of%20a%20file%20name%20is%20limited%20to,be%20shorter%20than%20256%20characters.)).

PR #2399, opened 2022-07-14
Update utilities

phlptp:
### Summary

If merged this pull request will update to the C++17 version of the utilities library

### DIFF
--- a/src/helics/application_api/FilterOperations.cpp
+++ b/src/helics/application_api/FilterOperations.cpp
@@ -46,7 +46,7 @@ void DelayFilterOperation::setString(std::string_view property, std::string_view
 {
     if (property == "delay") {
         try {
-            delay = gmlc::utilities::loadTimeFromString<helics::Time>(std::string(val));
+            delay = gmlc::utilities::loadTimeFromString<helics::Time>(val);
         }
         catch (const std::invalid_argument&) {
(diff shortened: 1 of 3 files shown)


nightlark (approved):

src/helics/application_api/FilterOperations.cpp
```diff
@@ -224,8 +224,8 @@ std::shared_ptr<FilterOperator> RandomDelayFilterOperation::getOperator()
 }
 
 RandomDropFilterOperation::RandomDropFilterOperation():
-    tcond(std::make_shared<MessageConditionalOperator>([this](const Message*) {
-        return (randDouble(random_dists_t::bernoulli, (1.0 - dropProb), 1.0) > 0.1);
+    tcond(std::make_shared<MessageConditionalOperator>([this](const Message* /*unused*/) {
+        return (randDouble(RandomDistributions::BERNOULLI, (1.0 - dropProb), 1.0) > 0.1);
```
nightlark, line 228: Is the random distribution used for this filter something that should be possible for a user to change?

phlptp: Yes but they would do it via a string, which hasn't changed.  This enumeration definition is not user visible.

PR #2403, opened 2022-07-18
EndpointData refactor

phlptp:
try to make the code structures for endpoint and inputs more similar for callback in the application API.
This may also resolve some of the failing test cases related to callbacks 

### Summary

If merged this pull request will refactor the EndpointData object in the application API to make it a little faster, match with the InputData object, update some of the object Names to better match conventions and hopefully resolve some of the failing test cases related to endpoint callbacks.


phlptp:
This doesn't look like it fully resolved the issue, but I still like the change so I want to keep it

I will keep trying

nightlark (approved):

PR #2407, opened 2022-07-26
add an HELICS_STATE_UNKNOWN result to the helicsFederateGetState

phlptp:
### Summary

If merged this pull request will add a HELICS_STATE_UNKNOWN to the helicsFederateGetState output


kdheepak:
That looks good to me. It doesn’t break existing API but guaranteed `noexcept`. I’ll update the enum in Python before the next release.

kdheepak (approved):

PR #2417, opened 2022-08-14
Add encryption feature to vcpkg support

nightlark:
### Summary

If merged this pull request will add the encryption feature to the vcpkg.json file that adds openssl as a dependency to install, along with a few other vcpkg related tweaks.

### Proposed changes

- Add encryption feature to vcpkg.json for installing OpenSSL
- Add `vcpkg_installed/` directory to `.gitignore` (local cache of compiled copies of all dependencies built by vcpkg, both debug and release variants by default)
- Set `VCPKG_MANIFEST_FEATURES` variable based on CMake options used; if vcpkg isn't getting used, this variable will just be ignored

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -481,6 +481,17 @@ if(HELICS_BUILD_CONFIGURATION STREQUAL "PI")
     target_link_libraries(helics_base INTERFACE atomic)
 endif()
 
+# Add to list of enabled vcpkg features based on core type
+if(HELICS_ENABLE_ZMQ_CORE)
+    list(APPEND VCPKG_MANIFEST_FEATURES "zeromq")
+endif()
+if(HELICS_ENABLE_IPC_CORE)
+    list(APPEND VCPKG_MANIFEST_FEATURES "ipc")
+endif()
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -481,6 +481,17 @@ if(HELICS_BUILD_CONFIGURATION STREQUAL "PI")
     target_link_libraries(helics_base INTERFACE atomic)
 endif()
 
+# Add to list of enabled vcpkg features based on core type
+if(HELICS_ENABLE_ZMQ_CORE)
+    list(APPEND VCPKG_MANIFEST_FEATURES "zeromq")
```
phlptp, line 486: where does this list get used

nightlark: In theory when building using vcpkg it can specify which features are enabled, so it will automatically built the right subset based on which have been enabled/disabled. I think to actually make that work the related CMake options and appending to VCPKG_MANIFEST_FEATURES needs to be reordered so they appear at the top of CMakeLists.txt, prior to setting the CMAKE_TOOLCHAIN_FILE to vcpkg.cmake and the project() command.

phlptp (approved):

PR #2429, opened 2022-09-07
Index Groups

phlptp:
Add support for an index group parameter to allow users direct influence of the federateID, which can in a few select cases eliminate a source of randomness in ordering (such as tiebreaking in timing, and message ordering).  All these are pretty rare but we did need a way to clear them up if it matters in a particular situation.


src/helics/core/CoreBroker.cpp
```diff
@@ -517,18 +507,25 @@ void CoreBroker::fedRegistration(ActionMessage&& command)
...
-                                             gGlobalFederateIdShift);
+        if (command.counter > 0 && command.counter <= 16) {
+            mFederates.back().global_id = GlobalFederateId(
+                static_cast<GlobalFederateId::BaseType>(lookupIndex) + gGlobalFederateIdShift +
+                command.counter * gGlobalPriorityBlockSize);
+        } else {
+            mFederates.back().global_id = GlobalFederateId(
+                static_cast<GlobalFederateId::BaseType>(lookupIndex) + gGlobalFederateIdShift);
+        }
+        mFederates.addSearchTermForIndex(mFederates.back().global_id, lookupIndex);
```
nightlark, line 526: Is the ID assigned to each new federate that connects just based on the current overall size of mFederates (excluding offsets due to shifts/index groups)? e.g. first federate gets ID 1, second one gets ID 2, etc?

nightlark (approved):

PR #2433, opened 2022-09-15
Develop

phlptp:
### Summary

If merged this pull request will merge to main for the 3.3 release


phlptp:
@trevorhardy take a loot at the changes to the Readme,  otherwise everything else is code changes that have been reviewed previously.

nightlark (approved):

phlptp:
it looks like there is a race condition in the callback federate operations.  I took a look and I don't think it is a critical one.  It will be fixed but it amounts to a race condition on a data field that doesn't get used later in the case of callback federates, so I am comfortable with leaving it in the release since callback federates are considered in Beta for this release.

trevorhardy (approved):

PR #2434, opened 2022-09-15
Main merger

phlptp:
### Summary

If merged this pull request will update develop so it can merge with main


phlptp:
There is a conflict in the merge to main, so this PR attempts to resolve that.  It needs to be a regular merge and not a squash and merge

nightlark (approved):

PR #2435, opened 2022-09-15
add the bcrypt library to all win32 builds

phlptp:
### Summary

If merged this pull request will fix an issue with bcrypt library and tweak a doc string


nightlark (approved):
Here's a test run of the release build workflow: https://github.com/GMLC-TDC/HELICS/actions/runs/3062216649

If a Windows installer gets uploaded and no linking errors appear in the build log, the fix works.

PR #2440, opened 2022-09-20
Add a disable operation for terminate commands

phlptp:
so federates/core/brokers cannot be remote controlled.

### Summary

If merged this pull request will add a disable flag for enabling/disabling a remote termination command.  The default was left as is so it shouldn't be a breaking change.

added some tests to verify operation.


nightlark (approved):

src/helics/application_api/FederateInfo.cpp
```diff
@@ -131,7 +131,7 @@ static constexpr frozen::unordered_map<std::string_view, int, 66> propStringsTra
     {"logBuffer", HELICS_PROPERTY_INT_LOG_BUFFER},
     {"log_buffer", HELICS_PROPERTY_INT_LOG_BUFFER}};
 
-static constexpr frozen::unordered_map<std::string_view, int, 83> flagStringsTranslations{
+static constexpr frozen::unordered_map<std::string_view, int, 86> flagStringsTranslations{
```
phlptp, line 134:
```suggestion
static constexpr frozen::unordered_map<std::string_view, int, 89> flagStringsTranslations{
```

Issue #2411, opened 2022-08-12
`--profiler=profile.txt` should write to `profile.txt` in `write` mode instead of `append` mode


kdheepak:
It would be nice if when I ran the following:

```
helics_broker -f3 --profiler=profile.txt
```

and a `profile.txt` file already exists, it writes to the file as though it were a new file, instead of writing to the file in append mode.

phlptp:
Doing this suggestion could be rather complex and problematic in some cases.  The way it was designed multiple different federates could write to the output file so it must append.  And the buffer is written periodically meaning it must append in that case as well otherwise you will only get partial results.   So I think appending must be the default.  

What could be done is if the file has a special character as the first character of the file name the file will be reset when the profiler is started  for example
```sh
helics_broker -f3 --profiler=-profile.txt
```

It also might be worth adding a divider comment that makes a clear separation when it first starts.

PR #2441, opened 2022-09-20
Profiler append

phlptp:
### Summary

If merged this pull request will clean up the profiling to not always append,  Append is now optional with "+" in front of the file name.


phlptp:
Fix #2411

docs/user-guide/advanced_topics/profiling.md
```diff
@@ -38,16 +38,18 @@ Profiling can be enabled at any level of the hierarchy in HELICS and when enable
 Profiling is enabled via the command prompt by passing the `--profiler` option when calling `helics_broker`.
 
 - `--profiler=save_profile2.txt` will save profiling data to a text file `save_profile2.txt`
+- `--profiler=+save_profile2.txt` will append profiling data to the text file `save_profile2.txt`
 - `--profiler=log` will capture the profile text output to the normal log file or callback
 - `--profiler` is the same as `--profiler=log`
 
-Enabling this flag will pass in the appropriate flags to all children brokers and cores.
+Enabling this flag will pass in the appropriate flags to all children brokers and cores. A leading `+` on the file name triggers appending new data, otherwise the file is cleared or a new one created when the broker starts
```
nightlark, line 45:
```suggestion
Enabling this flag will pass in the appropriate flags to all children brokers and cores. A leading `+` on the file name triggers appending new data, otherwise the file is cleared or a new file is created when the broker starts.
```

src/helics/core/ProfilerBuffer.cpp
```diff
@@ -54,4 +54,24 @@ void ProfilerBuffer::writeFile()
...
 
+void ProfilerBuffer::setOutputFile(std::string fileName)
+{
+    if (fileName.empty()) {
+        mFileName.clear();
+        return;
+    }
+    if (fileName.front() == '+') {
+        mFileName = fileName.substr(1);
+        return;
```
nightlark, line 65: I think this would be better as a separate option that changes log files to append mode -- file and folder names with a `+` are valid on Windows and Linux.

phlptp: I modified it so there is a separate --profiler_append=filename option instead of the '+'    I didn't like the idea of a separate option that would modify the --profiler,  I though that would be confusing

src/helics/core/ProfilerBuffer.hpp
```diff
@@ -17,7 +17,13 @@ class ProfilerBuffer {
     void addMessage(const std::string& data);
     void addMessage(std::string&& data);
     void writeFile();
-    void setOutputFile(std::string fileName) { mFileName = std::move(fileName); }
+    /** specify the output file for writing the profile too
```
nightlark, line 20:
```suggestion
    /** specify the output file for writing the profiler information to
```

src/helics/core/ProfilerBuffer.hpp
```diff
@@ -17,7 +17,13 @@ class ProfilerBuffer {
...
+    @details the file will be cleared when the output file is set unless the append parameter is
+    specified as true
+    @param fileName the name of the file to write the profile too
```
nightlark, line 23:
```suggestion
    @param fileName the name of the file to write the profile too
```
```suggestion
    @param fileName the name of the file to write the profile to
```

src/helics/application_api/FederateInfo.cpp
```diff
@@ -549,7 +549,7 @@ std::unique_ptr<helicsCLI11App> FederateInfo::makeCLIApp()
     app->add_option(
            "--profiler",
            profilerFileName,
-           "Enable profiling and specify a file name, \"log\" for sending profiling messages to the logger, otherwise specify the output file name")
+           "Enable profiling and specify a file name (NOTE: use --profiler_append=<filename> in the core init string to append an existing file)")
```
nightlark, line 552:
```suggestion
           "Enable profiling and specify a file name (NOTE: use --profiler_append=<filename> in the core init string to append to an existing file)")
```

docs/references/configuration_options_reference.md
```diff
@@ -268,7 +269,7 @@ In addition to these options, all options shown in the `broker_init_string` are
...
+Automatically generate a broker if one cannot be connected to. For federations with only one broker (often the case) and/or with federations containing custom federates that were developed for this particular application, it can be convenient to create the broker in the process of creating a specific federate; this option allows that to take place. The downside to this is it creates a federation with a small amount of mystery as the broker is not clearly shown to be launched as its own federate alongside the other federates and those unfamiliar with the federation composition may have to spend some extra time to understand where the broker is coming from. This option does not work for all core type, specifically designed for inproc and testing cores. Others may be added later.
```
nightlark, line 272:
```suggestion
Automatically generate a broker if one cannot be connected to. For federations with only one broker (often the case) and/or with federations containing custom federates that were developed for this particular application, it can be convenient to create the broker in the process of creating a specific federate; this option allows that to take place. The downside to this is it creates a federation with a small amount of mystery as the broker is not clearly shown to be launched as its own federate alongside the other federates and those unfamiliar with the federation composition may have to spend some extra time to understand where the broker is coming from. This option does not work for all core types, it is specifically designed for inproc and testing cores. Others may be added later.
```

docs/references/configuration_options_reference.md
```diff
@@ -256,9 +256,10 @@ A command-line-like string that specifies options for the core as it connects to
 - `--broker_rank=`: For MPI cores only; identifies the MPI rank of the broker
 - `--broker_tag=`: For MPI cores only; identifies the MPI tag of the broker
 - `--localport=`: Port number to use when communicating with this core
-- `--autobroker`: When included the core will automatically generate a broker
+- `--autobroker`: When included the core will automatically generate a broker(does not work for all core types)
```
nightlark, line 259:
```suggestion
- `--autobroker`: When included the core will automatically generate a broker (does not work for all core types)
```

nightlark (approved):

PR #2443, opened 2022-09-28
Single thread fed

phlptp:
### Summary

If merged this pull request will add the single thread federate capability.  

NOTE:  this will be first of a series of PR's that fully execute the new capability.  This one does it for federates only.  The next will start work on doing it for core which will be a much bigger deal.  later one there might be some doing something similar for the comms, which would enable a true single thread federate.  

So this PR adds the flags,  uses the opt objects from the shared library and handle some restrictions.


src/helics/application_api/FederateInfo.cpp
```diff
@@ -195,6 +197,12 @@ static constexpr frozen::unordered_map<std::string_view, int, 89> flagStringsTra
     {"single_thread_federate", HELICS_FLAG_SINGLE_THREAD_FEDERATE},
     {"singlethreadfederate", HELICS_FLAG_SINGLE_THREAD_FEDERATE},
     {"singleThreadFederate", HELICS_FLAG_SINGLE_THREAD_FEDERATE},
+    {"single_thread_core", HELICS_FLAG_SINGLE_THREAD_CORE},
+    {"singlethreadcore", HELICS_FLAG_SINGLE_THREAD_CORE},
+    {"singleThreadcore", HELICS_FLAG_SINGLE_THREAD_CORE},
+    {"multi_thread_core", HELICS_FLAG_MULTI_THREAD_CORE},
+    {"multithreadcore", HELICS_FLAG_MULTI_THREAD_CORE},
+    {"multiThreadcore", HELICS_FLAG_MULTI_THREAD_CORE},
```
nightlark, line 205:
```suggestion
    {"single_thread_core", HELICS_FLAG_SINGLE_THREAD_CORE},
    {"singlethreadcore", HELICS_FLAG_SINGLE_THREAD_CORE},
    {"singleThreadCore", HELICS_FLAG_SINGLE_THREAD_CORE},
    {"multi_thread_core", HELICS_FLAG_MULTI_THREAD_CORE},
    {"multithreadcore", HELICS_FLAG_MULTI_THREAD_CORE},
    {"multiThreadCore", HELICS_FLAG_MULTI_THREAD_CORE},
```

I think the camelCase version should match the snake_case variant for splitting up words.

phlptp: I agree good catch!

nightlark (approved):

PR #2444, opened 2022-10-08
External subprojects

phlptp:
### Summary

If merged this pull request will update the cmake files to better support use of HELICS as a subproject


docs/user-guide/installation/helics_cmake_options.md
```diff
@@ -98,3 +98,4 @@ There are a few options in the CMake system that are not visible in the GUI they
 - `HELICS_DISABLE_GIT_OPERATIONS` : will turn off any of the helper tools that require git, this is useful in a couple cases for building packages and other situations where updates shouldn't be checked and no modifications should be made.
 - `HELICS_SKIP_ZMQ_INSTALL`: This is only relevant if ZMQ is built as part of the compilation process, but it skips the installation of zmq as part of HELICS install in that case.
 - `HELICS_BENCHMARK_SHIFT_FACTOR`: For running the benchmarks this shift factor can be used to scale the number of federates used for the benchmark tests. If used it is required to be a number and is power of 2 shift from nominal values. For example for a small system a shift factor of -1 or -2 might be appropriate for the benchmarks not to take too long. The default for systems with 4 or fewer cores is `-1` and 0 for larger compute systems. For small 2 core systems a value of -2 might be appropriate. For some very large systems a bigger value might be able to be used.
+- `HELICS_HIDE_CMAKE_VARIABLES`: When using HELICS as a subproject in a CMake build this option can be enabled to hide all HELICS related variables in CMake so they won't show up CMake GUI.
```
nightlark, line 101:
```suggestion
- `HELICS_HIDE_CMAKE_VARIABLES`: When using HELICS as a subproject in a CMake build this option can be enabled to hide all HELICS related variables in CMake so they won't show up in the CMake GUI.
```

config/cmake/addfmt.cmake
```diff
@@ -13,15 +13,15 @@
 
 option(${PROJECT_NAME}_USE_EXTERNAL_FMT "Use external copy of {fmt}" OFF)
 mark_as_advanced(${PROJECT_NAME}_USE_EXTERNAL_FMT)
-
+if (NOT TARGET fmt::fmt)
 if(${PROJECT_NAME}_USE_EXTERNAL_FMT)
```
nightlark, line 17: It looks like this block of code needs to be reformatted to go a level deeper.

phlptp: why is the precommit not catching that?

nightlark (approved):

.pre-commit-config.yaml
```diff
@@ -60,7 +60,7 @@ repos:
     rev: v0.6.13
     hooks:
       - id: cmake-format
-        exclude: "cmake+"
+        exclude: "cmake_+"
```
nightlark, line 63: This also avoid formatting things in the `ThirdParty/` folder?

nightlark: Ah, that's covered by the exclude at the top of this config file.

phlptp:
NOTE:  HELICS_FMI still uses this branch.  That will get updated tomorrow and then I will delete this branch

PR #2447, opened 2022-10-12
fix some issues with the code coverage 

phlptp:
and few other warnings and potential bug fixes

One of that static analyzers detected a use after move which might have caused an issue with some of the only update on change flag usage.

### DIFF
--- a/.github/workflows/code-coverage.yml
+++ b/.github/workflows/code-coverage.yml
@@ -39,6 +39,7 @@ jobs:
     - name: Build HELICS
       run: |
         unset VCPKG_ROOT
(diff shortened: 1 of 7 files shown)


.github/workflows/code-coverage.yml
```diff
@@ -39,6 +39,7 @@ jobs:
     - name: Build HELICS
       run: |
         unset VCPKG_ROOT
+        git config --global --add safe.directory /__w/HELICS/HELICS
```
nightlark, line 42: I think this should either use the `$GITHUB_WORKSPACE` environment variable or `${{ github.workspace }}` workflow variable.

phlptp: done

nightlark (approved):

PR #2456, opened 2022-10-22
Split the application_api_test

phlptp:
into separate tests to allow reduction in max test time and organize a bit better

### Summary

If merged this pull request will split the application-api-tests into 4 separate tests to reduce complexity and individual max testing time and eventually a streamlining of the tests to reduce total time for the CI tests.


tests/helics/application_api/CMakeLists.txt
```diff
@@ -7,57 +7,91 @@
...
     ValueFederateAdditionalTests.cpp
-    CombinationFederateTests.cpp
     ValueFederateSingleTransfer.cpp
     ValueFederateDualTransfer.cpp
     helicsTypeTests.cpp
+    MultiInputTests.cpp
+)
+
+set(messageFederate_test_sources MessageFederateKeyTests.cpp MessageFederateAdditionalTests.cpp
+                                 ValueFederateAdditionalTests.cpp CombinationFederateTests.cpp
```
nightlark, line 30: `ValueFederateAdditionalTests` doesn't seem like it should be part of the message federate tests.

tests/helics/application_api/CMakeLists.txt
```diff
@@ -7,57 +7,91 @@
...
+    ValueFederateTestTemplates.hpp
     PrimaryTypeConversionTests.cpp
-    CallbackFederateTests.cpp
-    FilterAdditionalTests.cpp
     ValueConverterTests.cpp
-    application-api-tests.cpp
     data_viewTests.cpp
-    MessageFederateAdditionalTests.cpp
-    FederateTests.cpp
     subPubObjectTests.cpp
```
nightlark, line 21: These seem like they are a better fit for the application api tests, rather than value federate (depending on what the subPubObjectTests.cpp are doing).

tests/helics/application_api/CMakeLists.txt
```diff
@@ -7,57 +7,91 @@
...
-    GrantTimeoutTests.cpp
     ValueFederateAdditionalTests.cpp
-    CombinationFederateTests.cpp
     ValueFederateSingleTransfer.cpp
     ValueFederateDualTransfer.cpp
     helicsTypeTests.cpp
+    MultiInputTests.cpp
```
nightlark, line 26: These also sound like they are less value federate related and more application api tests.

tests/helics/application_api/CMakeLists.txt
phlptp: The PrimaryTypeConversion, ValueConverterTests, and subPubObjectTests are inherent parts of the value Federate data translation layers, so I think they belong in there,  dataView could be be too but I might move that to the applicationApi since that isn't completely specific to the valueFederate

tests/helics/application_api/CMakeLists.txt
phlptp: the multiInputTests relate to the specific case of an input having multiple sources so are quite connected to the valueFederate

nightlark (approved):

PR #2459, opened 2022-10-27
Submod update

phlptp:
### Summary

If merged this pull request will update submodules including containers, units, and CLI11


nightlark (approved):

nightlark:
It looks like the updated modules are `units` and `containers`, and not `CLI11` or `utilities`.

phlptp:
I forgot to push one of the commits, and I updated the PR description

PR #2470, opened 2022-11-14
add a remote terminal ability

phlptp:
HELICS to link to a rest API through the command line

### Summary

### Proposed changes

- add remote terminal capability
- refactor the api request capability in the test use in the remote terminal
- refactor the broker error return ack
- add a test flag to test connection (partial)
- add default stop on the probe app


src/helics/apps/helics-broker.cpp
```diff
@@ -288,3 +304,158 @@ void terminalFunction(std::vector<std::string> args)
...
+    auto* getsub = termProg.add_subcommand("get", "run a get query on the server")
+                       ->callback([&get, &target]() { get(target); });
+    getsub->add_option("target,--target", target, "specify the target uri to GET")->required();
+    // POST call
+    auto* postsub = termProg.add_subcommand("post", "post data to the server")
+                        ->callback([&post, &target, &body]() { post(target, body); });
+    postsub->add_option("target,--target", target, "specify the target uri to POST")->required();
+    postsub->add_option("body,--body", body, "specify the body of the POST message");
+
+    // POST call
```
nightlark, line 420:
```suggestion
    // PUT call
```

src/helics/apps/RestApiConnection.hpp
```diff
@@ -0,0 +1,54 @@
...
+#include <string_view>
+#include <vector>
+
+namespace helics::apps {
+
+class RestApiConnection {
+  public:
+    // Per-test-suite set-up.
+    // Called before the first test in this test suite.
+    // Can be omitted if not needed.
```
nightlark, line 29:
```suggestion
```

Stray comment from the class this was based on?

PR #2471, opened 2022-11-15
Arm in circle ci

phlptp:
### Summary

If merged this pull request will add an arm circle-ci build

### Proposed changes

- add arm based circle ci build
- add --system flag to the executables
- refactor some the CLI11 interactions and update the style on the enumeration to match standard


phlptp:
if we want to get rid of drone this is an option though the musl stuff is attractive as well to test.
I will probably move this to a daily test before we merge it

nightlark (approved):

PR #2521, opened 2023-02-25
Make some tweaks to support clang-cl for visual studio cleanly

phlptp:
### Summary

If merged this pull request will make a few tweaks generated from warnings when building HELICS with clang-cl through Visual studio

### DIFF
--- a/src/helics/application_api/Subscriptions.hpp
+++ b/src/helics/application_api/Subscriptions.hpp
@@ -20,32 +20,6 @@ SPDX-License-Identifier: BSD-3-Clause
 
 namespace helics {
 
-/** get a value directly from the subscription key name
-@details this is a convenience function to get a value directly from the subscription key name
-this function should not be used as the primary means of retrieving value as it does involve an
-additional map find operation vs the member getValue calls
-@param fed a reference to a valueFederate
-@param key  the name of the publication
-*/
-template<class X>
-X getValue(ValueFederate& fed, std::string_view key)
-{
-    return fed.getInput(key).getValue<X>();
-}
-
-/** get a value directly from the subscription key name
-@details this is a convenience function to get a value directly from the subscription key name
-this function should not be used as the primary means of retrieving value as it does involve an
-additional map find operation vs the member getValue calls
-@param fed a reference to a valueFederate
-@param key  the name of the publication
-@param obj the obj to store the retrieved value
-*/
-template<class X>
-void getValue(ValueFederate& fed, std::string_view key, X& obj)
-{
-    fed.getSubscription(key).getValue<X>(obj);
-}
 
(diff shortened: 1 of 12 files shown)


src/helics/core/helicsTime.hpp
```diff
@@ -39,12 +39,6 @@ constexpr Time cBigTime = Time{static_cast<int64_t>(HELICS_BIG_NUMBER * 1000000)
 /** common definition of currentTime in initialization mode*/
 constexpr Time initializationTime = negEpsilon;
 
-/** user defined literal for a time variable*/
-constexpr Time operator"" _t(long double val)
-{
-    return {static_cast<double>(val)};
-}  // NOLINT
-
```
nightlark, line 47: Is this function part of the public API?

src/helics/core/CoreBroker.cpp
```diff
@@ -2573,16 +2573,16 @@ static InterfaceType getMatchType(InterfaceType type)
...
     const BasicHandleInfo& origin,
+    uint32_t originFlags,
     const BasicHandleInfo& target,
-    uint32_t flagsSource,
-    uint32_t flagsDest,
+    uint32_t targetFlags,
     std::pair<action_message_def::action_t, action_message_def::action_t> actions)
 {
-    // notify the target about a source
+    // notify the target about a origin
```
nightlark, line 2581:
```suggestion
    // notify the target about an origin
```

src/helics/application_api/Subscriptions.hpp
```diff
@@ -20,33 +20,6 @@ SPDX-License-Identifier: BSD-3-Clause
...
-@param fed a reference to a valueFederate
-@param key  the name of the publication
-@param obj the obj to store the retrieved value
-*/
-template<class X>
-void getValue(ValueFederate& fed, std::string_view key, X& obj)
-{
-    fed.getSubscription(key).getValue<X>(obj);
-}
-
```
nightlark, line 49: If these are part of the public API, we should update the changelog to note that they are gone now and what alternative function to use.

phlptp: The C++ API only,  so yes it was public but also not used as far as I can tell, and generated some deprecation warnings and I think they have nearly completely been superceeded by other operations.  And I agree the changelog will need to be updated.

src/helics/core/helicsTime.hpp
phlptp: yes,  though like the above generated several warnings, is not very useful, and was generally confusing and not used other than a test case which I changed.

phlptp: but a changelog note is warranted.

PR #2536, opened 2023-04-11
Lower memory for find

phlptp:
### Summary

If merged this pull request will modify the get<Interface>By<TYPE> methods to not create a new object unless necessary and reuse existing ones where possible.  This will reduce memory consumption in some cases and prevent ever increasing memory consumption in a few usage patterns.


src/helics/shared_api_library/ValueFederateExport.cpp
```diff
@@ -455,10 +508,14 @@ HelicsInput helicsFederateGetInputByTarget(HelicsFederate fed, const char* targe
...
-        inp->fedptr = std::move(fedObj);
-        return addInput(fed, std::move(inp));
+        auto hInp = findInput(fed, id.getHandle());
+        if (hInp == nullptr) {
+            auto inp = std::make_unique<helics::InputObject>();
+            inp->inputPtr = &id;
+            inp->fedptr = std::move(fedObj);
+            hInp = addInput(fed, std::move(inp));
+        }
+        return hInp;
```
nightlark, line 518: This pattern feels like it could be its own function, or added as part of the `add*` or `find*` functions if it is a check that always needs to happen (`findOrCreate*?`).

phlptp: I submitted an update with this suggestion.

nightlark (commented):
Are functions for core filters/translators also going to be added?

phlptp:
I noticed that too,  Not in this PR but I was going to look a bit more into why they weren't added earlier then probably add them later.    If I had to guess it was since filters and translators added to a core wouldn't be through a file, so the user would have the object from the creation call,  whereas for the federates those can be added from a file with a need to get the object for later use.

nightlark (approved):

PR #2542, opened 2023-05-09
Test and handle config files given as argument with a space 

phlptp:
in the fie/path name.

### Summary

If merged this pull request will fix an issue with spaces in a config file name or path given through config constructor.  And add a series of test cases to verify this


trevorhardy (approved):
Tested with Fundamental Default example in User Guide in my Mac and it worked. This is compared to  the 3.4.0 build I was testing on yesterday where we got an error. As far as I can tell, problem fixed.

PR #2560, opened 2023-08-30
Translator config

phlptp:
### Summary

If merged this pull request will add support for translators via config files


trevorhardy:
Do we think this PR fixes the issues I was having? Also, do you want me to file a bug on it?

phlptp:
@trevorhardy  file a bug report.  I will probably need to add a specific test to verify.

src/helics/common/addTargets.hpp
```diff
@@ -51,13 +58,32 @@ void addTargets(const Json::Value& section, std::string targetName, Callable cal
...
         }
     }
+    return found;
+}
+
+template<typename Block, typename Callable>
+void addTargetVariations(const Block& section,
+                         std::string name1,
+                         std::string name2,
+                         Callable callback)
```
nightlark, line 77: Should make sure the variants searched for are documented where they might affect a user, including dropping the "s" at the end of a target name. Using similar names could get confusing if a user notices that things still work even though one place has an extra "s" at the end of the name, and if someone else wrote the config file could raise questions on if the behavior was intended.

phlptp: the basic rule was that [targets] could be an array,  [target] had to be a singular element

src/helics/application_api/ValueFederate.cpp
```diff
@@ -104,7 +104,7 @@ void ValueFederate::disconnect()
 ValueFederate& ValueFederate::operator=(ValueFederate&& fed) noexcept
 {
     vfManager = std::move(fed.vfManager);
-    if (getID() != fed.getID()) {  // the id won't be moved, as it is copied so use it as a test if
+    if (getID() != fed.getID()) {  // the inp won't be moved, as it is copied so use it as a test if
```
nightlark, line 107: inp == input? `id` seems to make more sense in this comment

```suggestion
    if (getID() != fed.getID()) {  // the id won't be moved, as it is copied so use it as a test if
```

tests/helics/application_api/TranslatorTests.cpp
```diff
@@ -449,25 +449,25 @@ TEST_F(translator, translator_connections9_nosan)
...
+    input1.setOption(HELICS_HANDLE_OPTION_CONNECTION_REQUIRED);
 
-    e1.addSourceTarget("t1");
-    p1.addDestinationTarget("t1");
-    i1.addSourceTarget("t1");
-    e1.addDestinationTarget("t1");
+    end1.addSourceTarget("t1");
+    pub1.addDestinationTarget("t1");
+    input1.addSourceTarget("t1");
+    end1.addDestinationTarget("t1");
```
nightlark, line 470: If we want to be consistent with the other tests, it looks like `end1` should be renamed to `endpoint1`.

src/helics/application_api/Federate.hpp
```diff
@@ -852,19 +853,24 @@ received
     @param tomlString  the location of the file or config String to load to generate the interfaces
     */
     void registerConnectorInterfacesToml(const std::string& tomlString);
+    /** check if a filter type and operation is valid
+    if strict is true throw an error
```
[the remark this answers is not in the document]
phlptp: relic of a previous iteration.   My first cut had this method as a separate function but then I needed too many parameters to make it work right, so it went back to a method in which case I didn't need strict any more(strict config checking), as it was already in the object.

src/helics/application_api/ValueFederate.cpp
phlptp: change all error I didn't catch

nightlark (approved):

PR #2580, opened 2023-12-29
Version updates

phlptp:
### Summary

If merged this pull request will update a number of the third party libraries including
fmt,
units,
asio
spdlog

### DIFF
--- a/config/cmake/AddGoogletest.cmake
+++ b/config/cmake/AddGoogletest.cmake
@@ -12,7 +12,7 @@
 # output on failed tests without having to set an environment variable.
 #
 
-set(gtest_version release-1.12.1)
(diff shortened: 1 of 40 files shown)


config/cmake/AddGoogletest.cmake
nightlark, line 1: We should also update the Google benchmarks version.

phlptp: I will look into that and google test after dealing with the clang-tidy stuff which seems to have gotten much more sensitive.  

We will also need to update the macos build since I think our oldest is no longer supported

phlptp:
@nightlark  think this is ready

nightlark (approved):

PR #2581, opened 2023-12-29
Create codeql.yml

phlptp:
only check C++ code

### Summary

If merged this pull request will add code-ql

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,81 @@
+# For most projects, this workflow file will not need changing; you simply need
+# to commit it to your repository.
+#
+# You may wish to alter this file to override the set of languages analyzed,
+# or to provide custom queries or build logic.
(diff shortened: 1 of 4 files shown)


phlptp:
guess this will take some work to get running

.github/workflows/codeql.yml
```diff
@@ -0,0 +1,81 @@
...
+# or to provide custom queries or build logic.
+#
+# ******** NOTE ********
+# We have attempted to detect the languages in your repository. Please check
+# the `language` matrix defined below to confirm you have the correct set of
+# supported CodeQL languages.
+#
+name: "CodeQL"
+
+on:
```
nightlark, line 14: I think this could also use a workflow dispatch trigger so we can manually trigger it (once the workflow makes its way into the main/default branch).

nightlark (approved):

PR #2595, opened 2024-01-20
update dates to 2024

phlptp:
### Summary

If merged this pull request will update dates to 2024

### DIFF
--- a/benchmarks/helics/BenchmarkMain.cpp
+++ b/benchmarks/helics/BenchmarkMain.cpp
@@ -1,5 +1,5 @@
 /*
-Copyright (c) 4,
+Copyright (c) 2017-2024,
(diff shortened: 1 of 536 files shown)


benchmarks/helics/BenchmarkMain.cpp
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright (c) 2017-2023,
+Copyright (c) 4,
```
nightlark, line 2:
```suggestion
Copyright (c) 2017-2024,
```

phlptp: nice catch!

nightlark (approved):

Issue #2600, opened 2024-01-25
helics_recorder configuration in plain text


temcdrm:
**Describe the bug**
We used to specify a plain-text configuration file behind **--config-file** and it worked properly, e.g., at https://github.com/pnnl/tesp/blob/develop/examples/capabilities/pypower/runhpp.sh

But in HELICS 3.4 we can only use the plain-text configuration file as a positional argument, e.g., at https://github.com/pnnl/pecblocks/blob/master/examples/helics/test/test_player.sh  Otherwise, the recorder won't subscribe to anything  and it's silent about the reason.

**What is the expected behavior?**
Any or all of the following:

- Clarify the documentation, including **--help**, to indicate that a plain-text config file must be a positional argument. However, this may break other examples that used to work.
- Issue a warning with diagnostic if the recorder has a problem with the **--config-file**
- Restore the operation of **--config-file** with plain-text configuration

Issue #2601, opened 2024-01-25
helics_player not keeping separate tags


temcdrm:
**Describe the bug**
A player file with 6 different publications is actually publishing them all under the same tag.  The values and timings are correct, but it's publishing under 1 tag, which is neither the first nor last encountered in the player file. This example used to work.  We could organize the player file in separate sections for each publication, and the player seemed to sort them correctly by time, and publish values under the correct tag. I don't have a workaround yet.

**To Reproduce**
This is a minimal test case: https://github.com/pnnl/pecblocks/tree/master/examples/helics/test. It's under a Python project but you don't need Python or any modules, just the HELICS apps.

**Environment (please complete the following information):**
HELICS 3.4 binaries on Windows and Linux.

phlptp:
Thank you for identifying this issue.    It appears to be an issue with the use of local and the use of text file input.   This was an oversight in testing for that combination.   A test has been added and a fix included.

temcdrm:
Thanks.  Do you know when this will show up in a binary release?  And in the Python Helics module?

PR #2602, opened 2024-01-25
clean up the config file processing for federates and apps

phlptp:
### Summary

If merged this pull request will clean up the config file processing and unify the file input a little better.


phlptp:
Will attempt to resolve #2600 and #2601

src/helics/common/JsonProcessingFunctions.cpp
```diff
@@ -16,6 +16,17 @@ SPDX-License-Identifier: BSD-3-Clause
...
 
+bool looksLikeCommandLine(std::string_view testString)
+{
+    if (testString.empty()) {
+        return false;
+    }
+    if (testString.front() == '-') {
+        return true;
+    }
+    return testString.find("--") != std::string::npos;
```
nightlark, line 27:
```suggestion
    return testString.front() == "--");
```

This seems like it should also be looking at just the front of the string. Or both this and the previous check need to check for whitespace preceeding these characters.

I'm guessing this functio is only used when the command line args are given in a config file, whereas running the helics apps and specifying these arguments directly should have the argument parsing handled by CLI11?

phlptp: It will only get used when the config is passed as a single string.    which could be a file of various types or command line arguments in a string that needs to be split and parsed.

nightlark (approved):

nightlark (approved):

PR #2607, opened 2024-02-03
Check if helicsCpp98_ide target exists before modifying properties

nightlark:
CMake gives an error on non-Windows platforms when including HELICS as a subproject due to the helicsCpp98_ide target not existing.


src/helics/CMakeLists.txt
```diff
@@ -72,6 +72,8 @@ if(NOT HELICS_DISABLE_C_SHARED_LIB)
     add_subdirectory(cpp98)
 
     if(NOT CMAKE_PROJECT_NAME STREQUAL PROJECT_NAME)
-        set_target_properties(helics helicsCpp98_ide PROPERTIES FOLDER helics)
+        if(TARGET helicsCpp98_ide)
+            set_target_properties(helics helicsCpp98_ide PROPERTIES FOLDER helics)
```
phlptp, line 76: This condition probably works fine since the folder only matters if if the ide is used.   But logically the condition should only apply to the helicsCpp98_ide, but it is setting it for the helics target as well which is pretty much guaranteed to exist.

nightlark: Ah right, those should be split up -- the helicsCpp98_ide target is only created if MSVC is being used; probably the most common IDE, but in theory XCode could see some use. It looks like CMake 3.27 has deprecated generators for all other IDEs (Eclipse, CodeBlocks, etc) in favor of the IDEs using the cmake-file-api.

phlptp (approved):

PR #2623, opened 2024-02-26
Add support a "helics" section of json files and a potential_interfac…

phlptp:
…es section to support connector operations

### Summary

If merged this pull request will add support for a "helics" section in json files, and a "potential_interfaces" section to support two stage initialization operations related to the connector.


src/helics/application_api/PotentialInterfacesManager.cpp
```diff
@@ -0,0 +1,124 @@
...
+namespace helics {
+PotentialInterfacesManager::PotentialInterfacesManager(Core* core, Federate* fed):
+    corePtr(core), fedPtr(fed)
+{
+}
+
+void PotentialInterfacesManager::loadPotentialInterfaces(Json::Value& json)
+{
+    static const std::set<std::string> interfaceTypes{
+        "publications", "inputs", "endpoints", "filters", "translators", "datasink"};
```
nightlark, line 25: Not `datasinks`?

src/helics/core/Core.hpp
```diff
@@ -848,9 +848,12 @@ class Core {
     @param queryFunction  a function object that returns a string as a result of a query in the form
     of const string ref. This callback will be called when a federate received a query that cannot
     be answered that directed at a particular federate
+    @param order multiple query callbacks can be set (0 or 1) is the first, then 2 and so on up to
```
nightlark, line 851: Docstring is a bit confusing -- is it some form of priority order that query callbacks should be run in if multiple query callbacks are set? Can multiple query callbacks be set with the same order value and what happens if you do? I think it should also mention that a value of 0 (or less) is equivalent to 1, and a value greater than 10 is equivalent to 10.

nightlark (commented):
What is the intended use case for the `helics` and `helics.helics` JSON sections?

src/helics/core/Core.hpp
phlptp: I needed a way to handle multiple layers of callback queries.   Setting the same order just replaces any existing one, and I will clarify the docstring

phlptp:
> What is the intended use case for the `helics` and `helics.helics` JSON sections?

I was encountering a series of configuration scenarios where the HELICS related configuration was a small portion of the overall config, and the other applications limited what could be in the root of the JSON, so all the HELICS related configuration had to go into a helics section for things to work nicely, otherwise the configuration management got ugly fast.   So it made sense just to directly support a helics section, and the entire config file could be fed in with no issues.

src/helics/core/Core.hpp
```diff
@@ -848,9 +848,13 @@ class Core {
     @param queryFunction  a function object that returns a string as a result of a query in the form
     of const string ref. This callback will be called when a federate received a query that cannot
     be answered that directed at a particular federate
+    @param order indicator of the execution order slot for query callbacks value is bound [1,10]
+    inclusive given values outside this range are clamped to the boundary values. The callback is
```
nightlark, line 852:
```suggestion
    @param order indicator of the execution order slot for query callbacks; the value is bound [1,10]
    inclusive and values given outside this range are clamped to the boundary values. The callback is
```

nightlark (approved):

PR #2629, opened 2024-03-18
Potential interface templates

phlptp:
### Summary

If merged this pull request will add the ability to handle interface template to the potential interfaces object and connector.


src/helics/apps/Connector.cpp
```diff
@@ -112,15 +113,15 @@ bool TemplateMatcher::loadTemplate(const Json::Value& iTemplate)
...
     if (tnameIndex == 0) {
-        intermediaries.push_back("");
+        intermediaries.emplace_back("");
     } else {
         intermediaries.push_back(templateName.substr(0, tnameIndex));
     }
     std::vector<std::string> valueNames;
     int index{0};
     while (tnameIndex != std::string::npos) {
         auto close = templateName.find_first_of('>', tnameIndex);
```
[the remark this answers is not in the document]
phlptp: you are probably right, I think I was initially going with the regex syntax but this isn't really a regex it is more similar to variable expansion which the `{}` may be more appropriate and used in other places.  I will change it.

docs/references/apps/Connector.md
```diff
@@ -249,4 +249,93 @@ also permissible are all arguments allowed for federates and any specific broker
...
+          ["typeC", "double", "kV"]
+        ],
+        "template": { "global": true }
+      }
+    ]
+  }
+}
+```
+
+The connector will evaluate all possibilities for the template for possible connections. The field name in `${fieldName}` is searched for in the json file. They may be duplicated, but are treated as independent for evaluation purposes as in the first example. The first template example defines over 175,000 different possible interfaces. The type of the interface can be defined as part of the template name, with the particular name as the key.
```
nightlark, line 341: Are there performance any performance related things that users need to be aware of if they create a template with a large number of possible interfaces?

phlptp: I thought about that for a while, when implementing this.   I was trying to figure out if there was a more efficient way of doing this.  The conclusion I came was that the templates are simply a more compact syntax for listing a large number of interfaces that could be present in a federate.   So the processing time and efficiency is exactly the same as if all those interfaces had been listed separately.  (slightly better from reduced communication overhead).   Ultimately it boils down to the matching process is N*logN complexity.  The multiplier on it gets pretty large but the the N is the number of interfaces, and all the other searches are map search which is logN.   Conceivably doing a large number of those searches but still logN complexity.  and the template is just a way to increase N with a small amount of input.

nightlark (approved):

PR #2634, opened 2024-03-19
fix some clang-tidy issues and update backup header files

phlptp:
### Summary

If merged this pull request will fix some clang-tidy issues in develop, update changelog, and backup header files


nightlark (approved):

docs/user-guide/advanced_topics/connector_fed.md
```diff
@@ -6,23 +6,24 @@ To support this HELICS comes with a connector application called "helics_connect
...
+The role of the "helics_connector" federate is to evaluate a mapping of the connections between federate interfaces and create the connections between these interfaces. Under the hood, it's using public APIs that allow the sources ("targets") for inputs and endpoints to be defined and destinations ("targets") to be added to publications. (Which is to say, if the stock "helics_connector" federate doesn't work the way you would like, you could actually make your own.) The documentation of the "helics_connector" app can be found on its [dedicated documentation page](https://docs.helics.org/en/latest/references/apps/Connector.html) but I'll summarize the highlights here:
```
nightlark, line 9:
```suggestion
The role of the "helics_connector" federate is to evaluate a mapping of the connections between federate interfaces and create the connections between these interfaces. Under the hood, it's using public APIs that allow the sources ("targets") for inputs and endpoints to be defined and destinations ("targets") to be added to publications. (Which is to say, if the stock "helics_connector" federate doesn't work the way you would like, you could actually make your own.) The documentation of the "helics_connector" app can be found on its [dedicated documentation page](../../references/apps/Connector.md) but I'll summarize the highlights here:
```

I think this can be a relative link, and sphinx will make sure that it links to the version of the page that matches the version of the docs that the user is browsing.

PR #2640, opened 2024-03-28
fix a discrepancy

phlptp:
between the parsing of config files and config strings in the handling of interface definitions.

### Summary

If merged this pull request will resolve a discrepancy between parsing of config files and config strings


src/helics/common/JsonProcessingFunctions.cpp
```diff
@@ -16,15 +16,34 @@ SPDX-License-Identifier: BSD-3-Clause
...
+    }
+    auto lastColonLoc = jsonString.find_last_of(':');
+    auto openBracket = jsonString.find_first_of('{');
+    if (openBracket == std::string_view::npos) {
+        return false;
+    }
+    auto closeBracket = jsonString.find_last_of('}');
+    if (closeBracket == std::string_view::npos) {
+        return false;
+    }
```
nightlark, line 42: I don't think either of these conditions is true. Valid JSON that would fail these checks include:

```json
[23, "abc", "def"]
```

```json
[]
```

```json
"abc"
```

```json
23
```

src/helics/common/JsonProcessingFunctions.cpp
```diff
@@ -16,15 +16,34 @@ SPDX-License-Identifier: BSD-3-Clause
...
 
-bool looksLikeCommandLine(std::string_view testString)
+bool looksLikeJson(std::string_view jsonString)
 {
-    if (testString.empty()) {
+    if (jsonString.empty()) {
         return false;
     }
-    if (testString.front() == '-') {
+    if ((jsonString.front() == '{') || (jsonString.front() == '//' && jsonString.back() == '}')) {
```
nightlark, line 24: I don't think this check is sufficient to determine if a file looks like JSON. Based on this, a C/C++ file such as this would look like JSON:

```c
// Copyright blah

void main() {
}
```

src/helics/common/JsonProcessingFunctions.cpp
phlptp: while true, those examples would not be valid config files

src/helics/common/JsonProcessingFunctions.cpp
phlptp: I was struggling with this a bit.  I need a very quick check that works most of the time.  If there was a very fast reliable way to check for valid json I would be interested

src/helics/common/JsonProcessingFunctions.cpp
```diff
@@ -16,15 +16,34 @@ SPDX-License-Identifier: BSD-3-Clause
...
         return true;
     }
-    return testString.find(" -") != std::string::npos;
+    if (jsonString.front() == '#' || jsonString.find("\n#") != std::string_view::npos) {
+        return false;
+    }
+    auto firstColonLoc = jsonString.find_first_of(':');
+    if (firstColonLoc == std::string_view::npos) {
+        return false;
+    }
```
nightlark, line 33: A `:` doesn't need to be present in a JSON file -- if a colon isn't found, I think firstColonLoc could be set to max value for data type returned, and lastColonLoc being set to -1 (if signed numbers are supported). The ordering check at the end looks like a pretty good one for when these elements are present.

src/helics/common/JsonProcessingFunctions.hpp
```diff
@@ -19,11 +19,10 @@ them using the jsoncpp library
 #include <string_view>
 
 namespace helics::fileops {
-/** check if a string looks like a command line string*/
-bool looksLikeCommandLine(std::string_view testString);
 /** check if the file has a valid JSON extension*/
 bool hasJsonExtension(std::string_view jsonString);
-
+/** check if the file has a valid JSON extension*/
```
nightlark, line 24:
```suggestion
/** check if the file has contents that could be valid JSON*/
```

src/helics/common/TomlProcessingFunctions.cpp
```diff
@@ -21,6 +21,18 @@ bool hasTomlExtension(std::string_view tomlString)
...
+bool looksLikeToml(std::string_view tomlString)
+{
+    if (tomlString.empty()) {
+        return false;
+    }
+    if (tomlString.find("[[") != std::string_view::npos) {
+        return true;
+    }
+
+    return (tomlString.find_first_of('=') != std::string_view::npos);
```
nightlark, line 33: Looking for an `=` is probably a pretty good indicator of a file being toml; might not even be worth looking for a nested table before hand. Technically I think an empty string is valid toml (but invalid json) -- however the results of parsing it would be pretty boring.

phlptp: The difficulty I ran into is that as part of the json or toml there could be strings with command line arguments containing '='

src/helics/common/JsonProcessingFunctions.cpp
```diff
@@ -16,15 +16,34 @@ SPDX-License-Identifier: BSD-3-Clause
 
 namespace helics::fileops {
 
-bool looksLikeCommandLine(std::string_view testString)
+bool looksLikeJson(std::string_view jsonString)
```
nightlark, line 19:
```suggestion
bool looksLikeJsonConfig(std::string_view jsonString)
```

Since we're looking for something that specifically looks like a JSON config (string), which is narrower than things that look like JSON.

src/helics/common/JsonProcessingFunctions.cpp
nightlark: If the scope is specifically a valid JSON config string, if we can guarantee that the outer most element is a JSON object ("{...}"), then I think the fastest check would strip leading whitespace and then check if the line begins with `{`. If it begins with a `//` for a comment line, do the same check for the next line.

I guess the ending of the file could also have comment lines that need to be ignored to find the last `}`.

If those two criteria don't hold, I think we can say with certainty that it is not going to be a valid JSON config string. If they are true, then we could probably skip the rest of the checks.

nightlark: In the best case, I think that check would be equivalent to the current condition given here -- worst case it turns into something closer to the subsequent checks that look for the exact locations of the brackets.

src/helics/common/JsonProcessingFunctions.cpp
```diff
@@ -16,15 +17,54 @@ SPDX-License-Identifier: BSD-3-Clause
...
+    auto firstColonLoc = jsonString.find_first_of(':');
+    if (firstColonLoc == std::string_view::npos) {
+        return false;
+    }
+    auto closeBracket = jsonString.find_last_of('}');
+    if (closeBracket == std::string_view::npos) {
+        return false;
+    }
+
+    auto afterBracket = removeSpaceAndComment(jsonString.substr(closeBracket + 1));
```
nightlark, line 66: I think this could pick up the wrong closing bracket if there are trailing comment lines and one happens to be something like `// ... }`. But this is probably good enough as a heuristic.

nightlark (approved):

PR #2659, opened 2024-07-07
update the potential interfaces spec to support simpler structures

phlptp:


src/helics/application_api/PotentialInterfacesManager.cpp
```diff
@@ -24,34 +24,55 @@ void PotentialInterfacesManager::loadPotentialInterfaces(Json::Value& json)
...
+                auto tInterface = interfaces[itype];
+                auto& pMap = potInterfaces[itype];
+                for (auto& ispec : tInterface) {
+                    auto name = fileops::getName(ispec);
+                    pMap[name] = ispec;
+                }
+            }
+            std::string tempString = itype;
+            tempString.pop_back();
+            tempString += "_templates";
```
nightlark, line 40: Should document later that the pop is to remove the trailing `s`.

nightlark (approved):

PR #2674, opened 2024-09-29
Min build requirements update

phlptp:
### Summary

If merged this pull request will update to cmake 3.30 and make google test, zeromq, and google benchmarks as submodules.


nightlark:
Is the issue with FetchContent related to the changes in https://cmake.org/cmake/help/latest/release/3.30.html#deprecated-and-removed-features about FetchContent_Populate being deprecated?

phlptp:
Yeah, the changes to fetch_content made it really awkward to do what we were doing.   I came across several others with similar problems.   It seems it was changed to make it work more like targets and single dependencies, but we were using to get a bunch of targets and configure the build, so the intention of CMake changes, and our use case then fits better with the submodule model.

.circleci/config.yml
```diff
@@ -138,19 +138,19 @@ jobs:
 
   helicsInstall1:
     docker:
-      - image: helics/buildenv:builder
+      - image: helics/buildenv:ubuntu24.04-default-builder
     environment:
-      CMAKE_FLAGS: "-DHELICS_BUILD_CXX_SHARED_LIB=ON -DHELICS_BUILD_TESTS=ON -DHELICS_BUILD_BENCHMARKS=ON -DHELICS_INSTALL_PACKAGE_TESTS=ON"
+      CMAKE_FLAGS: "-DHELICS_BUILD_CXX_SHARED_LIB=ON -DHELICS_BUILD_TESTS=ON -DHELICS_BUILD_BENCHMARKS=ON -DHELICS_INSTALL_PACKAGE_TESTS=ON "
```
nightlark, line 143: Does the trailing space make a difference?

.cirrus.yml
```diff
@@ -7,7 +7,7 @@ task:
     pkg install -y cmake libzmq4 ninja git
   build_script: |
     mkdir build && cd build
-    cmake -GNinja -DHELICS_BUILD_TESTS=ON -DHELICS_BUILD_EXAMPLES=ON ..
+    cmake -GNinja -DHELICS_BUILD_TESTS=ON -DHELICS_BUILD_EXAMPLES=ON -DCMAKE_CXX_STANDARD=17 ..
```
nightlark, line 10: Should the standard here be set to 20?

config/cmake/addBoost.cmake
```diff
@@ -61,170 +70,111 @@ if(WIN32 AND NOT UNIX_LIKE)
...
+        set(Boost_ROOT "${BOOST_INSTALL_PATH}")
+    elseif($ENV{BOOST_INSTALL_PATH})
+        set(Boost_ROOT "$ENV{BOOST_INSTALL_PATH}")
+    else()
+        set(Boost_ROOT "$ENV{BOOST_ROOT}")
     endif()
 endif()
 
+if($ENV{BOOST_ROOT})
+    set(BOOST_ROOT "$ENV{BOOST_ROOT}")
```
nightlark, line 120: Might be easier to follow if this new variable has a `_ENV` postfix added to the name, instead of just using the different case to distinguish it from `Boost_ROOT`.

vcpkg.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "helics",
-  "version-string": "3.5.3",
+  "version-string": "3.3.0",
```
nightlark, line 3:
```suggestion
  "version-string": "3.6.0",
```

nightlark (changes requested):

.cirrus.yml
phlptp: I set it to 17 since there is a compile issue that I didn't want to deal with in this PR.   When I changed the default to 20, cirrus stopped building.   Needs to be fixed but didn't want to deal with that here.

.circleci/config.yml
phlptp: probably not

config/cmake/addBoost.cmake
phlptp: BOOST_ROOT and Boost_ROOT are standard variables that are both supposed to work and do the same thing

.circleci/config.yml
phlptp, line 143:
```suggestion
      CMAKE_FLAGS: "-DHELICS_BUILD_CXX_SHARED_LIB=ON -DHELICS_BUILD_TESTS=ON -DHELICS_BUILD_BENCHMARKS=ON -DHELICS_INSTALL_PACKAGE_TESTS=ON"
```

config/cmake/addBoost.cmake
nightlark: Looking more into how find_package locates Boost, there could be some unexpected behavior finding Boost with CMake prior to 3.30 without a BoostConfig.cmake file (`Boost_NO_BOOST_CMAKE` set to `ON`). We might want to set BOOST_ROOT to mirror Boost_ROOT instead of ENV{BOOST_ROOT}.

---

For finding Boost using Config mode with find_package:
* Based on CMP0144, prior to 3.27 find_package ignores `<PACKAGENAME>_ROOT` variables, making this a no-op. For CMake >=3.27, supposedly find_package checks the `<PACKAGENAME>_ROOT` environment variable, so setting BOOST_ROOT to the environment variable is redundant, other than our find_path call on line 141 (which could use an alternate name for our temp variable or directly use `$ENV{BOOST_ROOT}`).

For finding Boost using the FindBoost module (only applies to CMake <3.30, see CMP0167):
* [FindBoost.cmake](https://github.com/Kitware/CMake/blob/master/Modules/FindBoost.cmake#L603) ignores BOOST_ROOT if Boost_ROOT is set, 
and has its own check for whatever is in $ENV{BOOST_ROOT} for trying to do a find_package call using CMake Config mode first.
* If the find_package check fails, it does a bunch of stuff to locate Boost -- the fallback search also uses the [environment variables](https://github.com/Kitware/CMake/blob/master/Modules/FindBoost.cmake#L1738) for possible locations, however it looks like it might only check the location given in [BOOST_ROOT but not Boost_ROOT](https://github.com/Kitware/CMake/blob/master/Modules/FindBoost.cmake#L1751). This could result in whatever location was given in `BOOST_INSTALL_PATH` not being searched at all.

config/cmake/addBoost.cmake
```diff
@@ -61,170 +70,111 @@ if(WIN32 AND NOT UNIX_LIKE)
...
+                    Boost-1.84.0
+                    Boost-1.83.0
+                    Boost-1.82.0
+                    Boost-1.81.0
+                    Boost-1.80.0
+                    Boost-1.79.0
+                    Boost-1.78.0
+                    Boost-1.77.0
+                    Boost-1.76.0
+                    Boost-1.75.0
```
nightlark, line 98: Should this list also include 1.74 and 1.73?

phlptp: The BoostConfig didn't work very well until 1.75.   So maybe we could add those if CMAKE <3.30.

config/cmake/addBoost.cmake
phlptp: I will look into it a little more this evening

phlptp: I am changing BOOST_ROOT to mirror Boost_ROOT if it is not set

config/cmake/addBoost.cmake
phlptp: Never mind I checked 1.74 and it looks like it would most likely work.   I will add them.

nightlark (approved):

PR #2682, opened 2024-10-10
Update Azure Pipelines macOS CI runner VMs

nightlark:
The macOS-12 runner is going away at the start of December, and we'll start seeing random failures during brownouts in the next couple of weeks. Bump the oldest macOS version tested to macOS-13, and the newest version to macOS-14.

### DIFF
--- a/.ci/azure-pipelines.yml
+++ b/.ci/azure-pipelines.yml
@@ -64,11 +64,11 @@ jobs:
       matrix:
         XCode-latest:
           test_config: "ci"
-          vmImage: "macOS-13"
+          vmImage: "macOS-14"
         XCode-oldest:
           test_config: "ci"
-          vmImage: "macOS-12"
-          xcode_path: "/Applications/Xcode_13.2.1.app"
+          vmImage: "macOS-13"
+          xcode_path: "/Applications/Xcode_14.1.app"
     pool:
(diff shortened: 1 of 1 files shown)


.ci/azure-pipelines.yml
```diff
@@ -64,11 +64,11 @@ jobs:
...
         XCode-latest:
           test_config: "ci"
-          vmImage: "macOS-13"
+          vmImage: "macOS-14"
         XCode-oldest:
           test_config: "ci"
-          vmImage: "macOS-12"
-          xcode_path: "/Applications/Xcode_13.2.1.app"
+          vmImage: "macOS-13"
+          xcode_path: "/Applications/Xcode_14.1.app"
```
phlptp, line 71: Just checking Xcode14.1 on macOS-13?  Guessing it is correct, just wanted to check

nightlark: For macOS-13 there are a few versions available, the oldest is XCode 14: https://github.com/actions/runner-images/blob/main/images/macos/macos-13-Readme.md#xcode

macOS-14 and onward will only have one major version of XCode, with a version number that is +1 the macOS version: https://github.com/actions/runner-images/issues/10703

phlptp (approved):

PR #2712, opened 2024-12-06
Release update 3 6

phlptp:
### Summary

If merged this pull request will update code in preparation for 3.6 release

PR #2713, opened 2024-12-06
Take a pointer as an argument for the scope exit guard in the MPI core

nightlark:
GCC expects the lambda for the scope exit guard to take a pointer as an argument.


nightlark:
Already fixed in #2712

PR #2719, opened 2024-12-27
Add check for exported_symbols_list linker option on macOS

nightlark:
The linker included with the zig c/c++ compiler doesn't support the exported_symbols_list linker option on macOS. For c/c++ compilation, I think zig is just using llvm/clang, but I guess they don't use the linker from llvm.


phlptp (approved):

phlptp:
can we merge this or is there more work to be done?

PR #2720, opened 2024-12-28
update to work with boost 1.87

phlptp:
### Summary

If merged this pull request will remove some usages of deprecated boost features which were removed in boost 1.87 now that we only support back to 1.73

### DIFF
--- a/src/helics/apps/helicsWebServer.cpp
+++ b/src/helics/apps/helicsWebServer.cpp
@@ -587,8 +587,10 @@ class WebSocketsession: public std::enable_shared_from_this<WebSocketsession> {
             return fail(eCode, "helics web server read");
         }
 
-        const std::string_view result{boost::asio::buffer_cast<const char*>(buffer.data()),
-                                      buffer.size()};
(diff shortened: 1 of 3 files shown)


src/helics/apps/helicsWebServer.cpp
```diff
@@ -587,7 +587,9 @@ class WebSocketsession: public std::enable_shared_from_this<WebSocketsession> {
             return fail(eCode, "helics web server read");
         }
 
-        const std::string_view result{boost::asio::buffer_cast<const char*>(buffer.data()),
+        // const std::string_view result{boost::asio::buffer_cast<const char*>(buffer.data()),
+        //                              buffer.size()};
+        const std::string_view result{reinterpret_cast<const char*>(buffer.data().data()),
```
nightlark, line 592: data() has to be called twice?

phlptp: the first one extracts a mutable buffer, the second one gets the void * pointer to the raw data.  The second `.data()` is the replacement method for the buffer_cast.   I looked at that twice too, but that appears to be the way to do it in this case.

nightlark (approved):

PR #2732, opened 2025-02-17
Add the ability to read interface targets and alias 

phlptp:
from potential interfaces.

### Summary

If merged this pull request will add connector targets to potential interfaces

### DIFF
--- a/src/helics/apps/Connector.cpp
+++ b/src/helics/apps/Connector.cpp
@@ -369,8 +369,7 @@ static void fedPotentialInterfaceList(ConnectionsList& connections, const nlohma
                 addPotentialTargetVariations(connections, name, "source", "publications", input);
                 addPotentialTargetVariations(connections, name, "source", "targets", input);
                 if (input.contains("alias")) {
-                    std::string alias = input["alias"].get<std::string>();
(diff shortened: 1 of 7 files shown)


src/helics/apps/Connector.cpp
```diff
@@ -287,6 +292,63 @@ nlohmann::json TemplateMatcher::usedInterfaceGeneration()
...
+        if (interface.contains(targetFieldName)) {
+            connections.potentialInterfaceConnections.emplace_back(
+                interfaceName, interface[targetFieldName].get<std::string>());
+            found = true;
+        }
+    }
+    return found;
+}
+
+static bool addPotentialTargetVariations(ConnectionsList& connections,
```
nightlark, line 326: The "names" for targets that are equivalent feel like they should be documented, unless colliding potential target names isn't a concern.

nightlark (approved):

src/helics/apps/Connector.cpp
phlptp: The connector is mainly concerned about making connections, so they all end up equivalent even if there is slightly different meanings of the different variations in the actual constructor.  Even in that case duplication is fine and in some cases intentional.   But we probably should check on the documentation of that topic to make sure it is clear.

PR #2786, opened 2026-01-24
Migrate to manylinux for release builds

nightlark:
A C++20 compiler is now required, gcc >= 13. HBB is still on gcc 9, so we need a newer base image for release builds. This change will use the manylinux Docker images that are frequently used when building Python wheels with native extensions for a variety of platforms.

Some additional restructuring is done, to enable publishing precompiled aarch64 Linux binaries -- and possibly musl libc based ones later on.


nightlark:
It looks like there are some other files in the repo that pre-commit is giving errors for (MinGW PKGBUILD).

phlptp (approved):

PR #2787, opened 2026-01-24
Add Windows ARM64 release build

nightlark:
Adds release builds for Windows on ARM.


phlptp:
@nightlark  is this something that needs to be manually triggered to test?

PR #2798, opened 2026-06-15
add sanitizer builds to github actions

phlptp:
### Summary

If merged this pull request will add sanitizer builds to github actions where they actually seem to run properly


nightlark:
The latest scheduled run is showing syntax errors in the workflow files (after commit ef51983).

PR #2804, opened 2026-06-22
update copyright and date

phlptp:
### Summary

If merged this pull request will update the copyright and copyright dates.


benchmarks/helics/ActionMessageBenchmarks.cpp
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright (c) 2017-2025,
+Copyright (c) 2017-2026,
 Battelle Memorial Institute; Lawrence Livermore National Security, LLC; Alliance for Sustainable
```
josephmckinsey: Not every Alliance for Sustainable Energy is changed. Seems like they are missed every time it's broken by a new line.

phlptp: I think I got them now

Issue #2764, opened 2025-09-11
Race Condition in Broker Hierarchy Initialization due to Incorrect Federate Count


CanaliDavide:
### **Describe the bug**

When setting up a broker hierarchy as described in the documentation ([doc](https://docs.helics.org/en/latest/user-guide/advanced_topics/broker_hierarchies.html)), a race condition can occur. The simulation sometimes enters the initialization phase before all federates across all sub-brokers have successfully connected. This results in a crash for the federates that fail to connect in time.

### **System description**
I have a hierarchy of 4 brokers and 16 federates connected in this way: 

<img width="611" height="381" alt="Image" src="https://github.com/user-attachments/assets/b6df581e-551f-426a-b963-8bd38c78aaab" />

Each broker has 4 federates connected to it (represented by the blue diamonds).

### **What is the current behavior?**

The root broker does not wait for the total number of federates in the entire hierarchy to connect. It only waits for the number of federates specified by its `-f` flag, plus its direct sub-brokers to connect. In the described setup with 16 total federates, the root broker's log indicates it only registered 15 federates before proceeding, causing the 16th federate to be rejected.

The log from the federate that fails to connect shows the following error:
```
[2025-09-11 15:49:33.242] [console] [error] core12 (0)[connected_error]::broker responded with error: already in initialization mode
```

The log from the root broker shows it has proceeded with an incomplete set of federates:
```
[2025-09-11 15:49:32.451] [console] [info] root (1)[connected]::{
   "summary": {
      "brokers": 3,
      "cores": 15,
      "countable_federates": 15,
      "federates": 15
   }
}
[2025-09-11 15:49:32.451] [console] [warning] 1-hkS2R-O0DAE-OjEgV-CKRCb::Unable to connect to Publication target Generic12/0/out1
```

### **What is the expected behavior?**

The root broker (and any sub-broker) should wait for the total number of expected federates in its entire branch of the hierarchy before entering initialization mode. The simulation should only start once all 16 federates are confirmed to be connected.

### **What is the motivation / use case for changing the behavior?**

To ensure reliable startup of simulations that use a broker hierarchy, preventing race conditions and crashes. The current behavior makes hierarchical setups unstable and difficult to debug.

### **To Reproduce**

1.  Create a broker hierarchy with 4 brokers and 16 federates, as shown in the provided image.
    *   Broker 2-0 is the root broker.
    *   Brokers 1-0 and 1-1 are sub-brokers connected to 2-0.
[...]

PR #2826, opened 2026-07-24
add --local_federates, --local_brokers, and --required_federates argu…

phlptp:
arguments to control when a federation starts up.

### Summary

If merged this pull request will add additional options for managing when a complex co-simulation will startup.  these include local federates,  local brokers, and required_federates which allows named federates to be required.


phlptp:
Fixes #2764

Issue #2739, opened 2025-05-02
Iteration with both values and endpoints blocks the simulation


BuiMCanmet:
**Describe the bug**
There seems to be a bug with the iteration feature when there is a combination of publications and endpoints. The simulation does get stuck unable to proceed/break out from the iteration loop. This bug was discovered during the office hours (2025-05-01).

**What is the expected behavior?**
The expected behavior is that the simulation should still be able to iterate within the publication even if an existing endpoint federate is present.

**To Reproduce**
Using the existing Advanced iteration example while adding the controller will recreate the issue.
I've create a repo https://github.com/[BuiMCanmet/helics_iteration_endpoints_values_example](https://github.com/BuiMCanmet/helics_iteration_endpoints_values_example) containing the reproducible example using
```bash
$ helics run --path advanced_iteration_runner.json
```

**Environment (please complete the following information):**

- Operating System: Windows
- Language Extension: Python
- HELICS version: HELICS Library version 3.6.1 (2025-02-24) & Python HELICS version 3.6.1

```bash
$ helics_app --version
```
HELICS App version: 3.6.1 (2025-02-24)

**Additional context and information**
The temporary solution was to add the targeted argument to true for all endpoints connection. For the controller itself, it was to also add targeted +all destination_endpoints.

PR #2829, opened 2026-07-31
Iteration bug

phlptp:
### Summary

If merged this pull request will try to resolve some of the iteration bugs observed.


phlptp:
Fixes issue #2739

